PR #20, opened 2019-07-24
Ylaala/configure fix

ylaala-gpsw:
Fixing severals missing checks in weird situations due to incomplete ngl_config but possible

### DIFF
--- a/libnodegl/glcontext_eagl.m
+++ b/libnodegl/glcontext_eagl.m
@@ -40,6 +40,7 @@ struct eagl_priv {
     GLuint depthbuffer;
     GLuint fbo_ms;
     GLuint colorbuffer_ms;
+    int fb_initialized;
 };
 
 static int eagl_init_layer(struct glcontext *ctx)
@@ -201,6 +202,7 @@ static int eagl_init_framebuffer(struct glcontext *ctx)
 
     ngli_glViewport(ctx, 0, 0, ctx->width, ctx->height);
 
+    eagl->fb_initialized = 1;
     return 0;
 }
 
@@ -218,9 +220,11 @@ static void eagl_uninit(struct glcontext *ctx)
 {
     struct eagl_priv *eagl = ctx->priv_data;
 
-    eagl_reset_framebuffer(ctx);
-
-    ngli_glDeleteRenderbuffers(ctx, 1, &eagl->colorbuffer);
+    if (eagl->fb_initialized) {
+        eagl_reset_framebuffer(ctx);
+        ngli_glDeleteRenderbuffers(ctx, 1, &eagl->colorbuffer);
+        eagl->fb_initialized = 0;
+    }
 
     if (eagl->framework)
         CFRelease(eagl->framework);
--- a/libnodegl/api.c
+++ b/libnodegl/api.c
@@ -58,12 +58,14 @@ static int cmd_reconfigure(struct ngl_ctx *s, void *arg)
         current_config->handle    != config->handle    ||
         current_config->offscreen != config->offscreen ||
         current_config->samples   != config->samples) {
-        ngli_node_detach_ctx(s->scene);
+        if (s->scene)
+            ngli_node_detach_ctx(s->scene);
         s->backend->destroy(s);
         int ret = s->backend->configure(s, config);
         if (ret < 0)
             return ret;
-        ret = ngli_node_attach_ctx(s->scene, s);
+        if (s->scene)
+            ret = ngli_node_attach_ctx(s->scene, s);
         if (ret < 0)
             return ret;
         return 0;
--- a/libnodegl/api.c
+++ b/libnodegl/api.c
@@ -160,7 +160,9 @@ end:;
 
 static int cmd_stop(struct ngl_ctx *s, void *arg)
 {
-    s->backend->destroy(s);
+    if (s->backend)
+        s->backend->destroy(s);
+
     return 0;
 }
 


sw-team-release-gpsw:
[CLA assistant check](https://cla.gopro.com/sign) <br/>All committers have signed the CLA.

libnodegl/glcontext_eagl.m
```diff
@@ -208,10 +208,17 @@ static void eagl_reset_framebuffer(struct glcontext *ctx)
...
+        ngli_glDeleteFramebuffers(ctx, 1, &eagl->fbo);
+
+    if (eagl->fbo_ms)
+        ngli_glDeleteFramebuffers(ctx, 1, &eagl->fbo_ms);
+
+    if (eagl->depthbuffer)
+        ngli_glDeleteRenderbuffers(ctx, 1, &eagl->depthbuffer);
+
+    if (eagl->colorbuffer_ms)
+        ngli_glDeleteRenderbuffers(ctx, 1, &eagl->colorbuffer_ms);
```
mbouron-gpsw: You can drop this patch, glDeleteRenderbuffers() silently ignores 0 or names that do not correspond to a valid renderbuffer.

libnodegl/glcontext_eagl.m
```diff
@@ -220,7 +227,8 @@ static void eagl_uninit(struct glcontext *ctx)
 
     eagl_reset_framebuffer(ctx);
 
-    ngli_glDeleteRenderbuffers(ctx, 1, &eagl->colorbuffer);
+    if (eagl->colorbuffer)
+        ngli_glDeleteRenderbuffers(ctx, 1, &eagl->colorbuffer);
```
mbouron-gpsw: Same as above.

libnodegl/glcontext_eagl.m
ylaala-gpsw: There is a crash in eagl_reset_framebuffer if they didn't have been generated

mbouron-gpsw: Is there a way to reproduce it ?

ubitux: > An alternative to this patch would be to add a field fbo_initialized to the eagl_priv structure and call eagl_reset_framebuffer() and ngli_glDeleteRenderbuffers() only if fbo_initialized is set to 1. What do you think ?

I prefer the usual if null check, it's a common pattern.

ubitux: > An alternative to this patch would be to add a field fbo_initialized to the eagl_priv structure and call eagl_reset_framebuffer() and ngli_glDeleteRenderbuffers() only if fbo_initialized is set to 1. What do you think ?

I prefer this as well because it doesn't imply the wrong assumption that these functions do not work with a 0/null parameter, but because they lack a proper gl context.

ylaala-gpsw: Ok, `fbo_initialized` seems more explicit to me in this case, in `eagl_uninit()` we will just call `eagl_reset_framebuffer()` and `ngli_glDeleteRenderbuffers()` if frame buffers has been initialized like @mbouron suggested. I agree the crash is on the glfunctions not being loaded though. I will make the modifications, if you agree on an other variable naming by tomorrow let me know

ylaala-gpsw:
Branch have been updated, let me know if there is enough explanation in commits message

libnodegl/glcontext_eagl.m
```diff
@@ -201,6 +202,7 @@ static int eagl_init_framebuffer(struct glcontext *ctx)
 
     ngli_glViewport(ctx, 0, 0, ctx->width, ctx->height);
 
+    eagl->fb_initialized = 1;
```
mbouron-gpsw: eagl->fb_initialized = 1 should be set at the beginning of the function otherwise we could leak gl ressources if init_framebuffer() fails.

mbouron-gpsw:
The branch has been merged to master with the following modifications:

- `eagl->fb_initialized = 1` has been set at the beginning of the `eagl_init_framebuffer()` function to avoid potential gl leaks.
- The commit messages weren't precise enough regarding what they were fixing and how to reproduce the issues and have been reworded. Please try to follow this scheme in the future, it saves time during the review process to understand and reproduce the issue.

Following, the modifications of the commit descriptions:

- The commit description `If users called ngl_configure() with a context that has already been configured and no scene has been set a crash could happen in these functions when accessing the scene pointer` has been reworded to:
```
Fixes a crash in ngli_node_{attach,detach_ctx}() when called from
cmd_reconfigure() on a context with no scene set and one of the
following configuration fields changed: display, window, handle,
offscreen, samples.
    
Can be reproduced with:
    
    struct ngl_config config = {.offscreen=1, .width=16, .height=16};
    struct ngl_ctx *s = ngl_create();
    ngl_configure(s, &config);
    config.samples = 4;
    ngl_configure(s, &config);
    ngl_freep(s);
```
- The commit description `If users called ngl_freep() with a non configured context a crash may occurs in stop_thread() when accessing the backend pointer` has been reworded to:
```
Fixes a crash when cmd_stop() is called from ngl_freep() on a context
which failed to be configured due to an invalid backend.
    
Can be reproduced with:
    
    struct ngl_config config = {.backend=0x1234};
    struct ngl_ctx *ctx = ngl_create();
    ngl_configure(ctx, &config);
     ngl_freep(&ctx);
```
- The commit description `If users called ngl_configure() and the glcontext creation failed there was a crash when trying to access glfunction pointers` has been reworded to:
```
Fixes a crash in eagl_uninit() if the EAGL context creation previously
failed in eagl_init() or if one of the functions related to the
glcontext failed in ngli_glcontext_new().
    
The crash was due to ngli_glDeleteFramebuffers() and
ngli_glDeleteRenderbuffers() trying to use respectively
gl->funcs.DeleteFramebuffers and gl->funcs.DeleteRenderbuffers pointers
that weren't loaded due to the context not being initialized.
```

PR #22, opened 2020-01-13
utils: rename Exporter.finished() signal

lrouch-gpsw:
There is already a QThread.finished() signal that is emitted
automatically when the thread exits:
https://doc.qt.io/qt-5/qthread.html#finished

In PyQt5 you could overwrite parent classes signals. With PySide2 the
existing QThread.finished() is still triggered, even though we could
think it was overwritten by Exporter.finished(), leading to connected
slots being called twice.

In the case of Exporter.finished(), the signal must only be emitted if the
export actually succeeded. Hence the chosen solution here is to rename
the signal to Exporter.export_finished().

### DIFF
--- a/pynodegl-utils/pynodegl_utils/export.py
+++ b/pynodegl-utils/pynodegl_utils/export.py
@@ -35,7 +35,7 @@ class Exporter(QtCore.QThread):
 
     progressed = QtCore.Signal(int)
     failed = QtCore.Signal()
-    finished = QtCore.Signal()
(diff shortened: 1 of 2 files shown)


sw-team-release-gpsw:
[CLA assistant check](https://cla.gopro.com/sign) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.gopro.com/sign) before we can accept your contribution.<br/>

PR #51, opened 2020-07-27
tests: use AnimatedVec4 instead of AnimatedVec3 for tests

mbouron-gpsw:
The r32g32b32_sfloat format is not supported on AMD Polaris GPUs with
the Vulkan API.

### DIFF
--- a/libnodegl/doc/libnodegl.md
+++ b/libnodegl/doc/libnodegl.md
@@ -601,7 +601,7 @@ Parameter | Ctor. | Live-chg. | Type | Description | Default
 `wrap_s` |  |  | [`wrap`](#wrap-choices) | wrap parameter for the texture on the s dimension (horizontal) | `clamp_to_edge`
 `wrap_t` |  |  | [`wrap`](#wrap-choices) | wrap parameter for the texture on the t dimension (vertical) | `clamp_to_edge`
 `access` |  |  | [`access`](#access-choices) | texture access (only honored by the `Compute` node) | `read+write`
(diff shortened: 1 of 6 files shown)


ubitux:
Functionally OK but I'd split the function factor out from the API changes.

Also, removing the tests while keeping the feature means reducing the coverage and holding an half supported and non-tested feature, so it feels pretty incomplete as is to me.

ubitux:
> These formats are not supported by Metal and not widely supported by
devices running Vulkan.

OK, but then this one looks unnecessary as it is more specific:

> The r32g32b32_sfloat format is not supported on AMD Polaris GPUs with
the Vulkan API.

Note: we could also drop `NGLI_FORMAT_R32G32B32_*` but that can come later.

Aside from the description comment, LGTM

mbouron-gpsw:
> > These formats are not supported by Metal and not widely supported by
> > devices running Vulkan.
> 
> OK, but then this one looks unnecessary as it is more specific:
> 
> > The r32g32b32_sfloat format is not supported on AMD Polaris GPUs with
> > the Vulkan API.

Dropped.

> 
> Note: we could also drop `NGLI_FORMAT_R32G32B32_*` but that can come later.
> 

NGLI_FORMAT_R32G32B32_* can still be used for buffers.

> Aside from the description comment, LGTM

ubitux:
> > Note: we could also drop `NGLI_FORMAT_R32G32B32_*` but that can come later.
> 
> NGLI_FORMAT_R32G32B32_* can still be used for buffers.
> 

Looking at `pgcraft.c` they're not mapped, probably because no such format exists in glsl?

PR #52, opened 2020-07-28
glcontext_egl: rely on surfaceless contexts if available for all platforms

mbouron-gpsw:
Offscreen rendering relies on framebuffer objects and we should only create an unused pbuffer on context that do not support surfaceless context (to stay compatible with old drivers/devices). Moreover, Mesa 20.1.4 (on Arch Linux) broke pbuffer creation on Intel GPUs.

### DIFF
--- a/libnodegl/glcontext_egl.c
+++ b/libnodegl/glcontext_egl.c
@@ -449,7 +449,9 @@ static int egl_make_current(struct glcontext *ctx, int current)
 static void egl_swap_buffers(struct glcontext *ctx)
 {
     struct egl_priv *egl = ctx->priv_data;
-    eglSwapBuffers(egl->display, egl->surface);
+
+    if (!ctx->offscreen)
+        eglSwapBuffers(egl->display, egl->surface);
 }
 
 static int egl_set_swap_interval(struct glcontext *ctx, int interval)
--- a/libnodegl/glcontext_egl.c
+++ b/libnodegl/glcontext_egl.c
@@ -248,21 +248,11 @@ static int egl_init(struct glcontext *ctx, uintptr_t display, uintptr_t window, 
     }
 
     const EGLint type = ctx->backend == NGL_BACKEND_OPENGL ? EGL_OPENGL_BIT : EGL_OPENGL_ES2_BIT;
-    EGLint surface_type = EGL_NONE;
-    if (ctx->platform == NGL_PLATFORM_XLIB ||
-        ctx->platform == NGL_PLATFORM_ANDROID) {
-        surface_type = ctx->offscreen ? EGL_PBUFFER_BIT : EGL_WINDOW_BIT;
-    } else if (ctx->platform == NGL_PLATFORM_WAYLAND) {
-        if (!egl->has_surfaceless_context_ext) {
-            LOG(ERROR, "EGL_KHR_surfaceless_context is not supported");
-            return -1;
-        }
-        surface_type = EGL_WINDOW_BIT;
-    } else {
-        ngli_assert(0);
-    }
+    EGLint surface_type = ctx->offscreen ? EGL_PBUFFER_BIT : EGL_WINDOW_BIT;
 
-    const EGLint config_attribs[] = {
+try_again:;
+
+    EGLint config_attribs[] = {
         EGL_RENDERABLE_TYPE, type,
         EGL_SURFACE_TYPE, surface_type,
         EGL_RED_SIZE,   8,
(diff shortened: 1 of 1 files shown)


libnodegl/glcontext_egl.c
```diff
@@ -278,6 +268,15 @@ static int egl_init(struct glcontext *ctx, uintptr_t display, uintptr_t window,
...
     ret = eglChooseConfig(egl->display, config_attribs, &egl->config, 1, &nb_configs);
+    if (ret && nb_configs == 0) {
+        /* Fallback to EGL_WINDOW_BIT if the driver do not advertize any
+         * pbuffer configurations. This happens on Wayland with Mesa. */
+        if (surface_type == EGL_PBUFFER_BIT) {
+            surface_type = EGL_WINDOW_BIT;
+            goto try_again;
+        }
+    }
+
```
cboesch-gpsw, line 279: You can avoid the goto with the following straightforward pattern (untested):

```c
static const EGLint offscreen_surface_types[] = {EGL_PBUFFER_BIT, EGL_WINDOW_BIT, EGL_NONE};
static const EGLint onscreen_surface_types[]  = {EGL_WINDOW_BIT, EGL_NONE};

const EGLint *surface_types = ctx->offscreen ? offscreen_surface_types
                                             : onscreen_surface_types;

for (int i = 0; surface_types[i] != EGL_NONE; i++) {
    const EGLint config_attribs[] = {
        EGL_RENDERABLE_TYPE, type,
        EGL_SURFACE_TYPE, surface_types[i],
        ...
    };

    EGLint nb_configs;
    ret = eglChooseConfig(egl->display, config_attribs, &egl->config, 1, &nb_configs);
    if (ret == EGL_TRUE && nb_configs)
        break;
}
```

Two notes on your diff:
- the condition `ret && nb_configs == 0` is weird: `ret` will match `EGL_TRUE` (success) and all the errors except one (`EGL_FALSE`); is this on purpose?
- this is already in the existing code, but I would feel more comfortable with `nb_configs` being initialized to `0`; you never know if the implementation would initialize it properly depending on the different error paths (and you're reading it when it fails here)

mbouron-gpsw: The condition `ret && nb_configs == 0` is on purpose, it means that eglChooseConfig() did not fail but no matching configuration was found.
nb_configs is only set by eglChooseConfig() if it suceeds. nb_configs is not supposed to be read if eglChooseConfig() failed, see the condition below `!ret || !nb_configs`.

mbouron-gpsw: I also prefer the current code over its variant without goto.

cboesch-gpsw: You're reading `nb_configs` in case `ret` is `EGL_TRUE`, `EGL_BAD_DISPLAY`, `EGL_BAD_ATTRIBUTE`, `EGL_NOT_INITIALIZED` and `EGL_BAD_PARAMETER`:
1. do you really want to catch all these cases here? It looks like you only want to catch `EGL_TRUE`
2. the doc seems indeed to imply that reading `nb_configs` in these cases is safe and only `EGL_FALSE` leads to an uninitialized `nb_configs`; but I'm not sure about different interpretation (or implementation tidbits) where some/all of the `EGL_BAD_*` above leads to an unitialized one, in which case you would end up reading back the uninitialized `nb_configs`.

cboesch-gpsw: Also, since you have a fallback mechanism, why not always prefer `EGL_WINDOW_BIT` (best) and then fallback `EGL_PBUFFER_BIT` (worse) when it fails?

mbouron-gpsw: ret can only be EGL_TRUE or EGL_FALSE. If EGL_FALSE is returned EGL_BAD_* errors are generated and returned by eglGetError().

Regarding the order, EGL_WINDOW_BIT is not better than EGL_PBUFFER_BIT.
For offscreen rendering if a surfaceless context is used, the chosen configuration is irrelevant, we just pick one that is compatible. Since this code is compatible with older drivers/devices that might not support surfaceless contexts, I prefer to select EGL_PBUFFER_BIT first (as it is required by those drivers/devices).

cboesch-gpsw: > ret can only be EGL_TRUE or EGL_FALSE. If EGL_FALSE is returned EGL_BAD_* errors are generated and returned by eglGetError() (not eglChooseConfig()).

I missed that, OK; I'd still explicit EGL_TRUE, but that's just me.

> Regarding the order, EGL_WINDOW_BIT is not better than EGL_PBUFFER_BIT.
For offscreen rendering if a surfaceless context is used, the chosen configuration is irrelevant, we just pick one that is compatible. Since this code is compatible with older drivers/devices that might not support surfaceless contexts, I prefer to select EGL_PBUFFER_BIT first (as it is required by those drivers/devices). If we select EGL_WINDOW_BIT first, we might end up with a configuration that is not able to create a pbuffer later on.

I see.

PR #53, opened 2020-07-28
Tooling (3rd iteration)

cboesch-gpsw:


mbouron-gpsw:
Allowing the Time node in the block node is probably a good idea too (if someone does not want to rely on one-slot uniforms).

ngl-tools/player.c
```diff
@@ -215,6 +218,79 @@ static void mouse_pos_callback(SDL_Window *window, SDL_MouseMotionEvent *event)
...
+
+    struct ngl_node *quad       = ngl_node_create(NGL_NODE_QUAD);
+    struct ngl_node *program    = ngl_node_create(NGL_NODE_PROGRAM);
+    struct ngl_node *render     = ngl_node_create(NGL_NODE_RENDER, quad);
+    struct ngl_node *time       = ngl_node_create(NGL_NODE_TIME);
+    struct ngl_node *v_duration = ngl_node_create(NGL_NODE_UNIFORMFLOAT);
+    struct ngl_node *v_opacity  = ngl_node_create(NGL_NODE_UNIFORMFLOAT);
+    struct ngl_node *coord      = ngl_node_create(NGL_NODE_IOVEC2);
+    struct ngl_node *group      = ngl_node_create(NGL_NODE_GROUP);
+    struct ngl_node *gcfg       = ngl_node_create(NGL_NODE_GRAPHICCONFIG, group);
```
mbouron-gpsw, line 252: Missing node pointer checks.

ngl-tools/player.c
```diff
@@ -266,7 +343,9 @@ int player_init(struct player *p, const char *win_title, struct ngl_node *scene,
     if (ret < 0)
         return ret;
 
+    scene = add_progress_bar(scene);
```
mbouron-gpsw, line 341: Can we add an option to disable the seek bar (it could be useful to disable it for debugging/profiling) ?

ngl-tools/ngl-python.c
```diff
@@ -29,15 +29,19 @@
 #include "python_utils.h"
 
 struct ctx {
+    /* options */
```
mbouron-gpsw, line 33: Maybe also add this comment in your previous patch (tools/player) ? (or remove it from both).

ngl-tools/opts.c
```diff
@@ -0,0 +1,205 @@
...
+        [OPT_TYPE_STR]      = opt_str,
+        [OPT_TYPE_TIME]     = opt_time,
+        [OPT_TYPE_LOGLEVEL] = opt_loglevel,
+        [OPT_TYPE_BACKEND]  = opt_backend,
+        [OPT_TYPE_RATIONAL] = opt_rational,
+        [OPT_TYPE_COLOR]    = opt_color,
+    };
+
+    if (ac > 1 && (!strcmp(av[1], "-h") || !strcmp(av[1], "--help")))
+        return 101;
```
mbouron-gpsw, line 158: 101 ?

ngl-tools/ngl-player.c
```diff
@@ -99,21 +111,22 @@ static int probe(const char *filename, struct sxplayer_info *media_info)
...
         .cfg.clear_color[3] = 1.f,
     };
 
-    if (argc < 2) {
-        fprintf(stderr, "Usage: %s <media>\n", argv[0]);
-        return -1;
+    int ret = opts_parse(argc, argc - 1, argv, options, ARRAY_NB(options), &s);
+    if (ret < 0 || ret == 101 || argc < 2) {
+        opts_print_usage(argv[0], options, ARRAY_NB(options), " <media>");
+        return ret == 101 ? 0 : EXIT_FAILURE;
```
mbouron-gpsw, line 128: 101 is again used here ? you probably want to define a proper return code.

ngl-tools/ngl-render.c
```diff
@@ -165,6 +165,7 @@ int main(int argc, char *argv[])
 
     printf("%s -> %s %dx%d\n", s.input ? s.input : "<stdin>", s.output ? s.output : "-", s.cfg.width, s.cfg.height);
 
+    const int show_window = !s.cfg.offscreen;
     if (show_window) {
```
mbouron-gpsw, line 168: You can use directly s.cfg.offscreen instead of adding a new local variable or rename this local variable to offscreen in a later commit.

ngl-tools/common.h
```diff
@@ -27,6 +27,9 @@
 
 #define ARRAY_NB(x) ((int)(sizeof(x) / sizeof(*(x))))
 
+#define DEFAULT_WIDTH  640
+#define DEFAULT_HEIGHT 480
```
mbouron-gpsw, line 31: Maybe prefer a 1:1 resolution or 16:9 ?

ngl-tools/player.c
cboesch-gpsw: Added. Note that this was initially copied from `ngl-player` where they were not present.

ngl-tools/player.c
cboesch-gpsw: Added. I reworked the history so that `ngl-player` and `ngl-python` switch to this generic seekbar one after another using this option.

ngl-tools/ngl-python.c
cboesch-gpsw: Yeah this comment was not introduced at the appropriate commit; I moved it into the "tools/python: add many options" commit where the use of `opts_parse()` is introduced. Similarly I added the comment in the `ngl-render` commit.

This comment is here to split the options modified by `opts_parse()` from the local context variables; it should now be consistent across tools.

ngl-tools/opts.c
cboesch-gpsw: `101` as in "help". I added a `#define OPT_HELP -101` for that one. Note that I make `opts_parse()` return an enum with that value, because `opts_parse()` returns the number of consumed argument, but can also return errors (negative values). `OPT_HELP` now extends the range of available error that can be returned.

ngl-tools/ngl-player.c
cboesch-gpsw: Changed to use `OPT_HELP`

ngl-tools/ngl-render.c
cboesch-gpsw: Done that in a separate commit; I want to keep this one as is.

ngl-tools/common.h
cboesch-gpsw: Changed to 640x360 (16:9)

cboesch-gpsw:
> Allowing the Time node in the block node is probably a good idea too (if someone does not want to rely on one-slot uniforms).

Good point, forgot about it. Branch updated.

PR #55, opened 2020-07-31
pipeline: decouples draw/dispatch parameters from the pipeline creation.

mbouron-gpsw:
Also allows attribute buffers to be set after pipeline creation.

PR #56, opened 2020-07-31
Tooling v4

cboesch-gpsw:
To address before merging:
- [x] depends on https://github.com/gopro/gopro-lib-node.gl/pull/55
- [x] altering the stencil is probably not a great idea


libnodegl/gctx.h
```diff
@@ -105,6 +105,7 @@ struct gctx {
     int features;
     struct limits limits;
     struct pgcache pgcache;
+    struct texture *text_atlas;
```
mbouron-gpsw, line 108: Maybe rename text_atlast to font_atlas ?
Also I would place this field in ngl_ctx instead of gctx as pgcache will be moved from gctx to ngl_ctx. Its destruction should be placed in api.c (cmd_configure() and cmd_stop()) The gctx structure will only hold fields that are mandatory for the graphics initialization (context, device, swap chain, ...), see https://github.com/gopro/gopro-lib-node.gl/pull/61.

libnodegl/node_text.c
```diff
@@ -118,163 +129,398 @@ static const struct node_param text_params[] = {
...
+
+    struct graphicstate state = ctx->graphicstate;
+    state.stencil_test       = 1;
+    state.stencil_write_mask = 0xFF;
+    state.stencil_func       = NGLI_COMPARE_OP_ALWAYS;
+    state.stencil_ref        = 1;
+    state.stencil_read_mask  = 0xFF;
+    state.stencil_fail       = NGLI_STENCIL_OP_REPLACE;
+    state.stencil_depth_fail = NGLI_STENCIL_OP_REPLACE;
+    state.stencil_depth_pass = NGLI_STENCIL_OP_REPLACE;
```
mbouron-gpsw, line 532: I think we should check that the current rendertarget has a stencil attachment before using any stencil operation.

mbouron-gpsw:
Other than my two remarks, LGTM.

libnodegl/node_text.c
mbouron-gpsw: Note that this should also be done in pass.c

libnodegl/gctx.h
cboesch-gpsw: Done. This simplifies the diff a lot, thanks.

libnodegl/node_text.c
cboesch-gpsw: I removed the use of the stencil entirely for now. `font_scale > 1.0` will cause an overflow which I documented in the commit message, along with the stencil settings for when we will want to use it again.

libnodegl/api.c
```diff
@@ -187,6 +187,7 @@ static int cmd_draw(struct ngl_ctx *s, void *arg)
 
 static int cmd_stop(struct ngl_ctx *s, void *arg)
 {
+    ngli_texture_freep(&s->font_atlas); // allocated by the first node text
```
mbouron-gpsw, line 190: This should also be added in cmd_configure(). A texture object can't outlive the graphical context.

libnodegl/node_text.c
```diff
@@ -390,17 +391,18 @@ static int atlas_create(struct ngl_node *node)
     tex_params.min_filter = NGLI_FILTER_LINEAR;
     tex_params.mag_filter = NGLI_FILTER_NEAREST;
     tex_params.mipmap_filter = NGLI_MIPMAP_FILTER_LINEAR;
-    s->texture = ngli_texture_create(gctx);
-    if (!s->texture) {
+
+    ctx->font_atlas = ngli_texture_create(gctx); // freed at context destruction
```
mbouron-gpsw, line 432: // freed at context reconfiguration/destruction

libnodegl/api.c
cboesch-gpsw: Mmh, so somehow it actually made sense to have it within the graphical context? Anyway, I added the following commit `api: call cmd_stop() from cmd_configure()` to address your comment.

libnodegl/node_text.c
cboesch-gpsw: Done

mbouron-gpsw:
LGTM.

mbouron-gpsw (approved):

PR #58, opened 2020-08-06
pipeline: remove unnecessary pipeline_compute structure and its use

mbouron-gpsw:


libnodegl/node_render.c
```diff
@@ -146,6 +146,11 @@ static int render_init(struct ngl_node *node)
     struct ngl_ctx *ctx = node->ctx;
     struct render_priv *s = node->priv_data;
 
+    if (s->nb_instances < 1) {
+        LOG(ERROR, "nb_instances must be > 0");
+        return NGL_ERROR_INVALID_USAGE;
```
ubitux, line 151: `NGL_ERROR_INVALID_ARG` should be used here; `NGL_ERROR_INVALID_USAGE` refers to something else, stuff like wrong nested function calls, overall inconsistent parameters, etc. while `NGL_ERROR_INVALID_ARG` is simply a not allowed value, whatever the surrounding context.

We might have confused the two in other places, but that was the original intent.

mbouron-gpsw: Branch updated.

PR #61, opened 2020-08-26
gctx: remove ngl_ctx dependency

mbouron-gpsw:
Removes ngl_ctx dependency from gctx API.


libnodegl/node_graphicconfig.c
```diff
@@ -249,57 +249,51 @@ static int graphicconfig_update(struct ngl_node *node, double t)
...
-        COPY_PARAM(stencil_fail);
-        COPY_PARAM(stencil_depth_fail);
-        COPY_PARAM(stencil_depth_pass);
-
-        COPY_PARAM(cull_face);
-        COPY_PARAM(cull_face_mode);
-
-        COPY_PARAM(scissor_test);
-    }
+    s->graphicstate = *pending;
```
cboesch-gpsw, line 259: Unless I'm mistaken, this local pending graphic state is not needed anymore (it seems to be write-only)?

Also, can you describe briefly the purpose/benefit of this commit? Is it to save the restore and have only a lazy state honoring?

libnodegl/rnode.c
```diff
@@ -42,7 +42,7 @@ void ngli_rnode_reset(struct rnode *s)
     for (int i = 0; i < ngli_darray_count(&s->children); i++)
         ngli_rnode_reset(&children[i]);
     ngli_darray_reset(&s->children);
-    memset(s, 0, sizeof(*s));
+    ngli_rnode_init(s);
```
cboesch-gpsw, line 45: After this changes, `ngli_rnode_clear()` and `ngli_rnode_reset()` are pretty much exactly the same; can we do something about it?

libnodegl/gctx.h
```diff
@@ -44,7 +44,7 @@ struct gctx_class {
 
     void (*set_rendertarget)(struct gctx *s, struct rendertarget *rt);
     struct rendertarget *(*get_rendertarget)(struct gctx *s);
-    struct rendertarget_desc *(*get_default_rendertarget_desc)(struct gctx *s);
+    const struct rendertarget_desc *(*get_default_rendertarget_desc)(struct gctx *s);
```
cboesch-gpsw: Can't this change be fixed up in the commit introducing the callback, or is it not possible because it required other changes?

libnodegl/gctx.h
```diff
@@ -105,7 +105,6 @@ struct gctx {
     int version;
     int features;
     struct limits limits;
-    struct pgcache pgcache;
```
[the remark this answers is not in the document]
mbouron-gpsw: Updated, thanks.

libnodegl/pass.c
```diff
@@ -437,6 +437,17 @@ int ngli_pass_prepare(struct pass *s)
     struct ngl_ctx *ctx = s->ctx;
     struct gctx *gctx = ctx->gctx;
 
+    int format = ctx->rendertarget_desc->depth_stencil.format;
```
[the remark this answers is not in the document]
mbouron-gpsw: Updated.

libnodegl/hwconv.c
```diff
@@ -129,7 +129,7 @@ int ngli_hwconv_init(struct hwconv *hwconv, struct ngl_ctx *ctx,
         .type          = NGLI_PIPELINE_TYPE_GRAPHICS,
         .graphics      = {
             .topology    = NGLI_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN,
-            .state       = ctx->graphicstate,
+            .state       = NGLI_GRAPHICSTATE_DEFAULTS,
```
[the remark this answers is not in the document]
mbouron-gpsw: Done.

libnodegl/node_graphicconfig.c
mbouron-gpsw: I kept it because we might need it in order to support live changes in the graphicconfig node. I will update the commit description.

libnodegl/gctx.h
mbouron-gpsw: I haven't done it in the original commit because it would have required more changes, ie: constifying ngl_ctx.rendertarget_desc which is removed in a later commit in this PR.

libnodegl/rnode.c
mbouron-gpsw: I updated the clear function so it only calls ngli_rnode_reset() and do not perform an unnecessary call to ngli_rnode_clear(). For now, I would like to keep the distinction between clear() and reset(), like for darray. The fact that ngli_rnode_clear() and ngli_rnode_reset() makes the rnode structure ends up in the same state is not supposed to be known by the user.

libnodegl/node_graphicconfig.c
mbouron-gpsw: Commit description updated.

cboesch-gpsw: Great description, but:

> I kept it because we might need it in order to support live changes in the graphicconfig node.

Please mention that as well (or just drop the assign and its field, it can be re-introduced easily).

libnodegl/rnode.c
cboesch-gpsw: Why do you need `ngli_rnode_reset()` to be re-entrant if that's the purpose of `ngli_rnode_clear()`? Where it is required?

libnodegl/node_graphicconfig.c
mbouron-gpsw: Commit description updated.

libnodegl/rnode.c
mbouron-gpsw: ngli_rnode_clear() is a shortcut for ngli_rnode_reset() + ngli_rnode_init(). My intention was avoid calling ngli_darray_count()/ngli_darray_data() on a zeroed darray. Commit dropped.

cboesch-gpsw: > My intention was avoid calling ngli_darray_count()/ngli_darray_data()

Yup that's fine with me but then let's address the issue globally: we should assert in count() and data() and check everywhere where it could happen.

cboesch-gpsw (approved):
LGTM

(please just add a line break before "Moreover" in the description of `internal: use the rnode structure to store the rendering state`, because it's an unrelated paragraph)

PR #62, opened 2020-08-27
tests/api: adjust clear color value

cboesch-gpsw:
0.3 * 255 = 76.5, which can round to 0x4C or 0x4D depending on the
implementation. Use the unambigous 0.4 (0x66) instead.


mbouron-gpsw (approved):
Test passes successfully on my machine. LGTM.

PR #64, opened 2020-08-27
nodes: fix parameter overread in check_params_sanity()

cboesch-gpsw:
The parameter must be read as a pointer if and only it's a pointer
(which is implied by PARAM_FLAG_NON_NULL).


mbouron-gpsw (approved):
LGTM.

PR #65, opened 2020-08-28
Tooling v5

cboesch-gpsw:


ngl-tools/player.c
```diff
@@ -478,6 +552,10 @@ void player_main_loop(void)
             case SDL_MOUSEMOTION:
                 mouse_pos_callback(p->window, &event.motion);
                 break;
+            case SDL_USEREVENT:
+                run = handle_map[event.user.code](event.user.data1) == 0;
+                free(event.user.data1);
+                break;
```
mbouron-gpsw, line 558: I think a leak of event.user.data1 can happen when player_main_loop() has exited but the server hasn't stopped yet (meaning it can still accept packets and potentially still push events (with its user data allocated). I think we should call a cleanup routine of user events after the server thread has stopped.

ngl-tools/ngl-desktop.c
```diff
@@ -0,0 +1,608 @@
...
+        free(p);
+    return 0;
+}
+
+static int handle_tag_scene(const uint8_t *data, int size)
+{
+    if (size < 1 || data[size - 1] != 0) // check if string is nul-terminated
+        return NGL_ERROR_INVALID_DATA;
+    const char *scene = (const char *)data;
+    return send_player_signal(PLAYER_SIGNAL_SCENE, scene, strlen(scene) + 1);
```
mbouron-gpsw, line 144: Is there a particular reason to not pass size directly here ?

ngl-tools/ngl-desktop.c
```diff
@@ -0,0 +1,608 @@
...
+    static const float fg_color[]  = {1.0, 2/3., 0.0, 1.0};
+    static const float subtext_h[] = {0.0, 0.5, 0.0};
+
+    struct ngl_node *group = ngl_node_create(NGL_NODE_GROUP);
+    struct ngl_node *texts[] = {
+        ngl_node_create(NGL_NODE_TEXT),
+        ngl_node_create(NGL_NODE_TEXT),
+    };
+
+    if (!group || !texts[0] || !texts[1])
```
mbouron-gpsw, line 412: Potential leak of group and texts.

ngl-tools/ngl-desktop.c
```diff
@@ -0,0 +1,608 @@
...
+    struct addrinfo hints = {
+        .ai_family   = AF_UNSPEC,
+        .ai_socktype = SOCK_STREAM,
+        .ai_flags    = AI_PASSIVE,
+    };
+
+    int ret = getaddrinfo(s->host, s->port, &hints, &s->addr_info);
+    if (ret) {
+        fprintf(stderr, "getaddrinfo: %s\n", gai_strerror(ret));
+        return -1;
```
mbouron-gpsw, line 433: Maybe use NGL_ERROR_? instead of -1 in this function ?

ngl-tools/player.c
cboesch-gpsw: OK. I added the following to `player_uninit()`:
```c
SDL_Event event;
while (SDL_PollEvent(&event))
    if (event.type == SDL_USEREVENT)
        free(event.user.data1);
```

ngl-tools/ngl-desktop.c
cboesch-gpsw: No it was an overlook due to using this string pattern in other packet related places. Replaced `strlen(scene) + 1` with `size`.

ngl-tools/ngl-desktop.c
cboesch-gpsw: Yeah good point. It's not even potential since I forgot to unref unconditionally `texts[0..1]` after they're added to the `group`. Fixed now.

ngl-tools/ngl-desktop.c
cboesch-gpsw: OK. Probably not ideal but I picked `NGL_ERROR_IO` for now. Since it's only in the scope of the tool (in opposition to within the public API) it should be good enough.

ngl-tools/ngl-ipc.c
```diff
@@ -0,0 +1,349 @@
...
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include <nodegl.h>
+
+#include "common.h"
+#include "ipc.h"
+#include "opts.h"
+
+#define UPLOAD_CHUNK_SIZE (1024 * 1024)
```
mbouron-gpsw, line 40: Maybe we should keep the chunk size under a standard/common MTU value ?

ngl-tools/ngl-ipc.c
```diff
@@ -0,0 +1,349 @@
...
+    s->upload_fd = 0;
+    free(s->upload_buffer);
+    s->upload_buffer = NULL;
+}
+
+static int handle_info(const uint8_t *data, int size)
+{
+    if (size < 1)
+        return NGL_ERROR_INVALID_DATA;
+    if (data[size - 1] == 0) {
```
mbouron-gpsw, line 186: Shouldn't this test be merge if the one above like for other handling function ?

mbouron-gpsw:
Other than my remarks, LGTM overall.

ngl-tools/ngl-ipc.c
cboesch-gpsw: But if we do that, it will cause a lot more write syscalls.

ngl-tools/ngl-ipc.c
cboesch-gpsw: Yes, done.

ngl-tools/ngl-ipc.c
```diff
@@ -0,0 +1,347 @@
...
+    return 0;
+}
+
+static int handle_filepart(struct ctx *s, const uint8_t *data, int size)
+{
+    if (size != 4)
+        return NGL_ERROR_INVALID_DATA;
+    const int written = IPC_U32_READ(data);
+    s->uploaded_size += written;
+    fprintf(stderr, "\ruploading %s... %d%%", s->uploadfile, (int)(s->uploaded_size * 100 / s->upload_stat.st_size));
```
mbouron-gpsw, line 197: * 100LL fixes an overflow on armv7h.

ngl-tools/ngl-ipc.c
mbouron-gpsw: Indeed. I just tested between machines on different networks if some fragmentation was happening, it was not the case. Let's keep 1024*1024 for now.

ngl-tools/ngl-ipc.c
cboesch-gpsw: I played with the idea of using `-D_FILE_OFFSET_BITS=64` instead to make sure `sizeof(off_t) == 8`, but for now let's go for this simple fix. Branch updated, thanks.

mbouron-gpsw (approved):

jmoguillansky-gpsw (commented):
"#include unistd.h": this code is platform specific.  
Can we update the code to work cross platform?

ubitux:
> "#include unistd.h": this code is platform specific.
> Can we update the code to work cross platform?

`unistd.h` is used in several places in the codebase already, we rely on it for various stuff. What portable solution do you propose? For Linux, MacOS, iOS and Android this is not an issue, and for Windows, MinGW has a compat layer that handles that for us.

PR #67, opened 2020-09-03
Remove triangle_fan topology

cboesch-gpsw:


jmoguillansky-gpsw (approved):
Looks good.  Good idea to remove triangle-fan: since metal/dx12 don't directly support.  Can consider using triangle strips in the future.

cboesch-gpsw:
> Can consider using triangle strips in the future.

That's a good point, not sure why I didn't do that in the first place. I'll update the PR.

cboesch-gpsw:
Branch updated. I switched to triangle strip for all the quads, which admittedly simplifies the diff and the resulting code. OTOH I kept using triangles + indices for the node circle and the square2circle demo (since all the triangles recenter on the same vertex), and for the cropboard test with the indices (for variety/complexity).

jmoguillansky-gpsw (approved):

mbouron-gpsw (approved):
LGTM.

PR #68, opened 2020-09-03
Remove back+front culling

mbouron-gpsw:
Back+Front face culling is not supported by Metal and DirectX 12.


libnodegl/node_circle.c
```diff
@@ -64,9 +64,9 @@ static int circle_init(struct ngl_node *node)
     uvcoords[0] = 0.5;
     uvcoords[1] = 0.5;
     for (i = 1; i < (nb_vertices - 1); i++) {
-        const double angle = (i - 1) * step;
-        const double x = sin(angle) * s->radius;
+        const double angle = (i - 1) * -step;
         const double y = cos(angle) * s->radius;
+        const double x = sin(angle) * s->radius;
```
cboesch-gpsw, line 69: Can you elaborate on the fix itself?

Also, you're moving the `x` assign line but it does seem identical.

tests/refs/shape_circle_cull_back.ref
```diff
@@ -0,0 +1 @@
+00000000000000000000000000000000 00000000000000000000000000000000 00000000000000000000000000000000 00000000000000000000000000000000
```
cboesch-gpsw, line 1: So half of the tests give a black frame: can't we do something more interesting with the culling?

tests/shape.py
```diff
@@ -367,3 +367,33 @@ def shape_triangles_mat4_attribute(cfg):
...
+
+    geometry = None
+    if shape == 'triangle':
+        geometry = ngl.Triangle()
+    elif shape == 'quad':
+        geometry = ngl.Quad()
+    elif shape == 'circle':
+        geometry = ngl.Circle()
+    else:
+        assert False
```
cboesch-gpsw, line 383:
```python
geometry_cls = dict(
    triangle=ngl.Triangle,
    quad=ngl.Quad,
    circle=ngl.Circle,
)
geometry = geometry_cls[shape]()
```

tests/shape.py
```diff
@@ -367,3 +367,33 @@ def shape_triangles_mat4_attribute(cfg):
...
+def _get_shape_function(shape, cull_face_mode):
+    @test_fingerprint()
+    @scene()
+    def shape_function(cfg):
+        return _get_shape_scene(cfg, shape, cull_face_mode)
+    return shape_function
+
+
+for shape in ['triangle', 'quad', 'circle']:
+    for cull_face_mode in ['front', 'back']:
```
cboesch-gpsw, line 398: please use `tuple` or `set` in these iterations

libnodegl/node_circle.c
mbouron-gpsw: Done.

tests/shape.py
mbouron-gpsw: Done.

tests/shape.py
mbouron-gpsw: Done.

tests/refs/shape_circle_cull_back.ref
mbouron-gpsw: Front-face culling tests are supposed to discard the triangles provided by the circle,quad,circle nodes hence the black frames. If the resulting black frame is due to a bug in node.gl it will be caught by the back-culling/non-culling versions of the tests.

cboesch-gpsw: OK

cboesch-gpsw (approved):

PR #71, opened 2020-09-08
added swapchain module.  Added buffer map/unmap/download functions

jmoguillansky-gpsw:


mbouron-gpsw:
Can you split your commit in 4:
* buffer: add map/unmap/download (that applies on top of master or before WIP:vulkan, and it only adds the new functions in gctx.c/h / buffer.c/h / buffer_gl.c/h
* buffer_vk: implement map/unmap/download (that will be squashed later on in WIP: vulkan)
* internal: add swapchain API (same as the first buffer commit, ie: it only adds the new functions in gctx.c/h, swapchain.c/h, swapchain_gl.c/h (which must be implemented even if it's probably an empty stub, due to OpenGL having no control over the swapchain))
* swapchain_vk: implement swapchain API (also squashed later on in WIP: vulkan)

This way, the commits that can be applied on top of master (and that are not Vulkan related) can be merged into master when they are ready.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,21 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
     return 0;
 }
 
+int ngli_buffer_gl_download(struct buffer* s, void* data, uint32_t size, uint32_t offset) {
+    LOG(ERROR, "not supported");
```
mbouron-gpsw, line 76: This needs to be implemented if the gl context supports it. In order to do so, what you can do in a previous commit (named glfeatures: add buffer map/unmap support) is to create a wrapper around glMapBufferRange/glUnmapBuffer() and probe its support:
* add 'glMapBufferRange', glUnmapBuffer' to cmds_optional list in libnodegl/gen-gl-wrappers.py, then run make nodegl-gen_gl_wrappers from the root directory of the project.
* add a new feature in libnodegl/features.h (maybe NGLI_FEATURE_MAP_BUFFER or whatever makes sense in this context)
* add the relevant declarations in libnodegl/glfeatures_data.h (version and es_version set + relevant gl function offsets)

You can look for various glfeatures: * commits for reference (for example: 1a36371afd58360ed2fbd7bdd69a9094f638066c glfeatures: add sync object feature)

PR #75, opened 2020-09-09
NGLI Optimizations (GL): add buffer map/unmap/download, add swapchain module

jmoguillansky-gpsw:


libnodegl/glcontext_egl.c
```diff
@@ -83,11 +83,13 @@ EGLBoolean ngli_eglDestroyImageKHR(struct glcontext *gl, EGLImageKHR image)
     return egl->DestroyImageKHR(egl->display, image);
 }
 
+#if defined(TARGET_ANDROID)
```
mbouron-gpsw, line 86: Please reword commit message to glcontext_egl: guard ngli_eglGetNativeClientBufferANDROID() declaration

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+    return 0;
+}
+
+void ngli_buffer_gl_unmap(struct buffer* s) {
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+	
+	if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
+        LOG(WARNING, "context does support glUnmapBuffer");
```
mbouron-gpsw, line 106: Indentation.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+    return NGL_ERROR_UNSUPPORTED;
+}
+
+int ngli_buffer_gl_map(struct buffer *s, int size, uint32_t offset, void** data)
+{
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+	
+	if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
```
mbouron-gpsw, line 86: Identation.

libnodegl/buffer.h
```diff
@@ -21,6 +21,7 @@
 
 #ifndef BUFFER_H
 #define BUFFER_H
+#include <stdint.h>
```
mbouron-gpsw, line 24: Please add empty line before this include.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
     return 0;
 }
 
+int ngli_buffer_gl_download(struct buffer* s, void* data, uint32_t size, uint32_t offset) {
+    LOG(ERROR, "not supported");
```
mbouron-gpsw, line 76: This method should be implemented.

libnodegl/swapchain.h
```diff
@@ -0,0 +1,34 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef SWAPCHAIN_H
+#define SWAPCHAIN_H
+#include <stdint.h>
```
mbouron-gpsw, line 24: New line before this include.

libnodegl/swapchain.c
```diff
@@ -0,0 +1,39 @@
...
+    return gctx->class->swapchain_create(gctx);
+}
+
+void ngli_swapchain_destroy(struct gctx *gctx) {
+	if (!gctx->class->swapchain_destroy) return NGL_ERROR_UNSUPPORTED;
+    gctx->class->swapchain_destroy(gctx);
+}
+
+int ngli_swapchain_acquire_image(struct gctx *gctx, uint32_t *image_index) {
+	if (!gctx->class->swapchain_acquire_image) return NGL_ERROR_UNSUPPORTED;
```
mbouron-gpsw, line 35: Indentation + the return statement should placed on a new line (same for the others occurrences).

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+    return NGL_ERROR_UNSUPPORTED;
+}
+
+int ngli_buffer_gl_map(struct buffer *s, int size, uint32_t offset, void** data)
+{
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+
```
mbouron-gpsw, line 85: Spurious empty characters.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+{
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+	
+	if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
+        LOG(WARNING, "context does support glMapBufferRange");
+        return NGL_ERROR_GENERIC;
+    }
+
```
mbouron-gpsw, line 90: Spurious empty characters.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+    *data = ngli_glMapBufferRange(gl, GL_ARRAY_BUFFER, offset, size, GL_MAP_READ_BIT | GL_MAP_WRITE_BIT); //TODO: pass access flags
+
+    return 0;
+}
+
+void ngli_buffer_gl_unmap(struct buffer* s) {
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+
```
mbouron-gpsw, line 104: Spurious empty characters.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+void ngli_buffer_gl_unmap(struct buffer* s) {
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+	
+	if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
+        LOG(WARNING, "context does support glUnmapBuffer");
+    }
+
```
mbouron-gpsw, line 108: Spurious empty characters.

libnodegl/gen-gl-wrappers.py
```diff
@@ -109,6 +109,10 @@
     'glUniform2uiv',
     'glUniform3uiv',
     'glUniform4uiv',
+
```
mbouron-gpsw, line 112: Spurious empty characters.

ngl-tools/Makefile
```diff
@@ -27,7 +27,7 @@ SDL_LDLIBS := $(shell $(PKG_CONFIG) --libs   sdl2)
 SXPLAYER_CFLAGS := $(shell $(PKG_CONFIG) --cflags libsxplayer)
 SXPLAYER_LDLIBS := $(shell $(PKG_CONFIG) --libs   libsxplayer)
 
-TOOLS_CFLAGS := $(shell $(PKG_CONFIG) --cflags libnodegl)
+TOOLS_CFLAGS := $(shell $(PKG_CONFIG) --cflags libnodegl) -fPIC
```
mbouron-gpsw, line 30: Why is this needed on Ubuntu 18.10 and not on more recent distributions (Ubuntu 19.x, 20.04) ? Is there another benefit of using this flag ?

mbouron-gpsw: Does this issue also happen when the project is not build with SHARED=yes ?

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+	if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
+        LOG(WARNING, "context does support glMapBufferRange");
+        return NGL_ERROR_GENERIC;
+    }
+    
+    if (!data) {
+        LOG(ERROR, "data is NULL");
+        return NGL_ERROR_GENERIC;
+    }
+    *data = ngli_glMapBufferRange(gl, GL_ARRAY_BUFFER, offset, size, GL_MAP_READ_BIT | GL_MAP_WRITE_BIT); //TODO: pass access flags
```
mbouron-gpsw, line 95: The gl buffer must be bound before calling ngli_glMapBufferRange().

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,43 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+
+    return 0;
+}
+
+void ngli_buffer_gl_unmap(struct buffer* s) {
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
```
mbouron-gpsw, line 103: Unused buffer_gl variable.

ngl-tools/Makefile
jmoguillansky-gpsw: We need the flag -fPIC to resolve a compile issue.  It enabled position independent code.  
Perhaps this flag was already set in your environment indirectly via pkg-config and a third-party lib, but regardless of third-party libs: we need to ensure that the right compile flags are enabled in our code.

libnodegl/buffer_gl.c
jmoguillansky-gpsw: what is the issue with indentation?

mbouron-gpsw:
The commits "style: fix indentation issues", "buffer: add download" and "buffer: bind before map/unmap", should be fixed up in the relevant previous commits.

ngl-tools/Makefile
mbouron-gpsw: Can you rename this commit to "tools/build: enable position independant code

Fixes the following build issue on Ubuntu 18.10:
/usr/bin/ld: ngl-serialize.o: relocation R_X86_64_32 against `.rodata.str1.8' can not be used when making a PIE object; recompile with -fPIC
/usr/bin/ld: python_utils.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC"

cboesch-gpsw, line 30: > Perhaps this flag was already set in your environment indirectly via pkg-config and a third-party lib

No, the reason you need to add `-fPIC` is because of a bug in Debian/Ubuntu packaging and Python. You can see it because the only 2 binaries it's unable to link are the 2 that depend on Python. You can find an extensive explanation of this issue in `.github/workflows/ci_linux.yml`:

```yaml
#
# This -fPIE is a workaround for the following link issue:
#     relocation R_X86_64_32 against `.rodata.str1.8' can not be used
#     when making a PIE object; recompile with -fPIE
#
# To summarize the issue:
#
# - On Python side, `python3-config` is a script generated from a
#   template using a global CFLAGS. But at the same time, the script
#   uses LIBS and SYSLIBS to construct the LDFLAGS (and not LDFLAGS for
#   some moronic reason).
# - On Debian/Ubuntu side, they decided to have the following injected:
#   CFLAGS="-specs=/usr/share/dpkg/no-pie-compile.specs"
#   LDFLAGS="-specs=/usr/share/dpkg/no-pie-link.specs"
#   And unfortunately, the former can not go without the latter
#   (otherwise there is a link error).
#
# As a result, `python3-config --cflags` will return the `-specs=...`
# flags, but `python3-config --ldflags` will not return the
# complementary `-specs=...`.
#
export CFLAGS=-fPIE
```

I opened an issue a few months ago on Debian bug tracker: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959945. Needless to say, they don't give a damn so far, but feel free to insist.

> but regardless of third-party libs: we need to ensure that the right compile flags are enabled in our code.

This is incorrect as well: while `-fPIC` makes sense for shared libraries, is not very meaningful for end binaries. It could be used for some security reasons, but this is merely a distribution choice, not something the project has to decide for the user.

So what I suggest instead if you really want to workaround the bugs of Debian/Ubuntu/Python, is to make an exception for this configuration only. For example, like this:

```diff
diff --git a/.github/workflows/ci_linux.yml b/.github/workflows/ci_linux.yml
index 3e03e7c4..bfca20fc 100644
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -30,28 +30,6 @@ jobs:

     - name: Build
       run: |
-        #
-        # This -fPIE is a workaround for the following link issue:
-        #     relocation R_X86_64_32 against `.rodata.str1.8' can not be used
-        #     when making a PIE object; recompile with -fPIE
-        #
-        # To summarize the issue:
-        #
-        # - On Python side, `python3-config` is a script generated from a
-        #   template using a global CFLAGS. But at the same time, the script
-        #   uses LIBS and SYSLIBS to construct the LDFLAGS (and not LDFLAGS for
-        #   some moronic reason).
-        # - On Debian/Ubuntu side, they decided to have the following injected:
-        #   CFLAGS="-specs=/usr/share/dpkg/no-pie-compile.specs"
-        #   LDFLAGS="-specs=/usr/share/dpkg/no-pie-link.specs"
-        #   And unfortunately, the former can not go without the latter
-        #   (otherwise there is a link error).
-        #
-        # As a result, `python3-config --cflags` will return the `-specs=...`
-        # flags, but `python3-config --ldflags` will not return the
-        # complementary `-specs=...`.
-        #
-        export CFLAGS=-fPIE
         make -j$(($(nproc)+1))
     - name: Run tests with GL backend
       run: |
diff --git a/ngl-tools/Makefile b/ngl-tools/Makefile
index 44784858..b7ad0650 100644
--- a/ngl-tools/Makefile
+++ b/ngl-tools/Makefile
@@ -74,6 +74,32 @@ PYTHON_LDLIBS_EMBED-yes = $(shell python$(PYTHON_MAJOR)-config --ldflags --embed
 PYTHON_HAS_EMBED := $(shell python$(PYTHON_MAJOR)-config --embed >/dev/null && echo yes || echo no)
 PYTHON_LDLIBS := $(PYTHON_LDLIBS_EMBED-$(PYTHON_HAS_EMBED))
 TOOLS += python serialize
+
+#
+# This -fPIE is a workaround for the following link issue:
+#     relocation R_X86_64_32 against `.rodata.str1.8' can not be used
+#     when making a PIE object; recompile with -fPIE
+#
+# To summarize the issue:
+#
+# - On Python side, `python3-config` is a script generated from a
+#   template using a global CFLAGS. But at the same time, the script
+#   uses LIBS and SYSLIBS to construct the LDFLAGS (and not LDFLAGS for
+#   some moronic reason).
+# - On Debian/Ubuntu side, they decided to have the following injected:
+#   CFLAGS="-specs=/usr/share/dpkg/no-pie-compile.specs"
+#   LDFLAGS="-specs=/usr/share/dpkg/no-pie-link.specs"
+#   And unfortunately, the former can not go without the latter
+#   (otherwise there is a link error).
+#
+# As a result, `python3-config --cflags` will return the `-specs=...`
+# flags, but `python3-config --ldflags` will not return the
+# complementary `-specs=...`.
+#
+DISTRIB_ID := $(shell lsb_release -si 2>/dev/null)
+ifeq ($(DISTRIB_ID),$(filter $(DISTRIB_ID),Ubuntu Debian))
+PYTHON_CFLAGS += -fPIE
+endif
 endif

 TOOLS_BINS = $(addprefix ngl-, $(addsuffix $(EXESUF), $(TOOLS)))
```

This is untested, but you get the idea. Also, It's also important to limit this to Python objects, it's not required for the other executables.

libnodegl/egl.h
```diff
@@ -36,7 +36,9 @@ EGLImageKHR ngli_eglCreateImageKHR(struct glcontext *gl,
 EGLBoolean ngli_eglDestroyImageKHR(struct glcontext *gl,
                                    EGLImageKHR image);
 
+#if defined(TARGET_ANDROID)
 EGLClientBuffer ngli_eglGetNativeClientBufferANDROID(struct glcontext *gl,
                                                      const struct AHardwareBuffer *buffer);
+#endif
```
cboesch-gpsw, line 42: Can you provide the configuration/scenario in which this fails to compile and the complete message? I'm curious about the 2 conflicting definitions.

libnodegl/features.h
```diff
@@ -53,6 +53,7 @@
 #define NGLI_FEATURE_SOFTWARE                     (1 << 28)
 #define NGLI_FEATURE_UINT_UNIFORMS                (1 << 29)
 #define NGLI_FEATURE_EGL_ANDROID_GET_IMAGE_NATIVE_CLIENT_BUFFER (1 << 30)
+#define NGLI_FEATURE_MAP_BUFFER                   (1 << 31)
```
cboesch-gpsw, line 56: This is undefined because it's a signed shift in which you reach the MSB. You need to do something like `1U << 31`.

libnodegl/buffer.c
```diff
@@ -39,6 +39,21 @@ int ngli_buffer_upload(struct buffer *s, const void *data, int size)
     return s->gctx->class->buffer_upload(s, data, size);
 }
 
+int ngli_buffer_download(struct buffer* s, void* data, uint32_t size, uint32_t offset)
+{
+    return s->gctx->class->buffer_download(s, data, size, offset);
+}
+
+int ngli_buffer_map(struct buffer *s, int size, uint32_t offset, void** data)
```
cboesch-gpsw, line 47: `void **` doesn't make much sense and may lead to aliasing issue; you want `void *`.

Also style point: beware in C the convention is to stick the `*` to the variable name (you have this issue in various places). While semantically weird, the reason is because declaring multiple pointers looks like `type *foo, *bar`, where sticking the `*` to the type would be confusing and prone to error.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,51 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
     return 0;
 }
 
+int ngli_buffer_gl_download(struct buffer* s, void* data, uint32_t size, uint32_t offset) {
```
cboesch-gpsw, line 75: Another style point: we have a line break before the `{` for function name. It's just a habit we have from other projects, sorry about that inconsistency.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,51 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+
+int ngli_buffer_gl_map(struct buffer *s, int size, uint32_t offset, void** data)
+{
+    struct gctx_gl *gctx_gl = (struct gctx_gl *)s->gctx;
+    struct glcontext *gl = gctx_gl->glcontext;
+    struct buffer_gl *s_priv = (struct buffer_gl *)s;
+
+    if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
+        LOG(WARNING, "context does support glMapBufferRange");
+        return NGL_ERROR_GENERIC;
```
cboesch-gpsw, line 93: `NGL_ERROR_UNSUPPORTED`?

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,51 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+    }
+
+    if (!data) {
+        LOG(ERROR, "data is NULL");
```
cboesch-gpsw, line 97: Can this be triggered by a end user (from someone using the public API) or can it be triggered only because of a bug in our code? If the latter, please use `ngli_assert(data)` instead.

libnodegl/swapchain.c
```diff
@@ -0,0 +1,43 @@
...
+    gctx->class->swapchain_destroy(gctx);
+}
+
+int ngli_swapchain_acquire_image(struct gctx *gctx, uint32_t *image_index) {
+    if (!gctx->class->swapchain_acquire_image)
+        return NGL_ERROR_UNSUPPORTED;
+    return gctx->class->swapchain_acquire_image(gctx, image_index);
+}
+
+
```
cboesch-gpsw, line 43: Trailing line breaks

libnodegl/swapchain.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <stdint.h>
+
+struct gctx;
+
+int ngli_swapchain_create(struct gctx *s);
+
+void ngli_swapchain_destroy(struct gctx *s);
+
+int ngli_swapchain_acquire_image(struct gctx *s, uint32_t *image_index);
```
cboesch-gpsw, line 33: style nitpick: please pack the prototypes together

jmoguillansky-gpsw:
Thanks, and I should probably update the warning message too …

jmoguillansky-gpsw:
Thanks, I’ll follow this approach and test it

jmoguillansky-gpsw:
Thanks, good catch

jmoguillansky-gpsw:
We’re returning a void*.  So we can either use void* as a return value, or as a parameter: but in this case we need to pass by address (e.g. void**)?

jmoguillansky-gpsw:
Thanks, agreed

libnodegl/features.h
mbouron-gpsw: Maybe it is also time to add a commit to change glcontext and gctx features to int64_t ?

ngl-tools/Makefile
mbouron-gpsw: You will probably want to scope the execution of lsb_release to Linux systems.

libnodegl/egl.h
jmoguillansky-gpsw: lcontext_egl.c:86:17: error: conflicting types for ‘ngli_eglGetNativeClientBufferANDROID’
 EGLClientBuffer ngli_eglGetNativeClientBufferANDROID(struct glcontext *gl, const struct AHardwareBuffer *buffer)
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from glcontext_egl.c:40:
egl.h:39:17: note: previous declaration of ‘ngli_eglGetNativeClientBufferANDROID’ was here
 EGLClientBuffer ngli_eglGetNativeClientBufferANDROID(struct glcontext *gl,
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
glcontext_egl.c: In function ‘ngli_eglGetNativeClientBufferANDROID’:
glcontext_egl.c:89:46: warning: passing argument 1 of ‘egl->GetNativeClientBufferANDROID’ from incompatible pointer type [-Wincompatible-pointer-types]
     return egl->GetNativeClientBufferANDROID(buffer);
                                              ^~~~~~
glcontext_egl.c:89:46: note: expected ‘const struct AHardwareBuffer *’ but argument is of type ‘const struct AHardwareBuffer *’

ngl-tools/Makefile
cboesch-gpsw: > You will probably want to scope the execution of lsb_release to Linux systems.

You're not even sure to have `lsb_release` available on Linux; the call is silenced by `2>/dev/null`, so it should be fine everywhere.

libnodegl/glcontext.h
```diff
@@ -48,7 +48,7 @@ struct glcontext {
     int version;
 
     /* GL features */
-    int features;
+    int64_t features;
```
mbouron-gpsw, line 51: This should be part of a dedicated commit (with the other chunk regarding gctx).

libnodegl/egl.h
cboesch-gpsw: > expected ‘const struct AHardwareBuffer *’ but argument is of type ‘const struct AHardwareBuffer *’

??

libnodegl/swapchain.c
```diff
@@ -0,0 +1,41 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+ 
+#include "swapchain.h"
+#include "gctx.h"
+
+int ngli_swapchain_create(struct gctx *gctx) {
```
cboesch-gpsw, line 25: style

libnodegl/swapchain.c
```diff
@@ -0,0 +1,41 @@
...
+    if (!gctx->class->swapchain_create)
+        return NGL_ERROR_UNSUPPORTED;
+    return gctx->class->swapchain_create(gctx);
+}
+
+void ngli_swapchain_destroy(struct gctx *gctx) {
```
cboesch-gpsw, line 31: style

libnodegl/swapchain.c
```diff
@@ -0,0 +1,41 @@
...
+    if (!gctx->class->swapchain_destroy)
+        return NGL_ERROR_UNSUPPORTED;
+    gctx->class->swapchain_destroy(gctx);
+}
+
+int ngli_swapchain_acquire_image(struct gctx *gctx, uint32_t *image_index) {
```
cboesch-gpsw, line 37: style

libnodegl/buffer.c
```diff
@@ -39,6 +39,21 @@ int ngli_buffer_upload(struct buffer *s, const void *data, int size)
     return s->gctx->class->buffer_upload(s, data, size);
 }
 
+int ngli_buffer_download(struct buffer* s, void* data, uint32_t size, uint32_t offset)
```
cboesch-gpsw, line 42: style (`buffer*` and `void*`)

libnodegl/buffer.c
```diff
@@ -39,6 +39,21 @@ int ngli_buffer_upload(struct buffer *s, const void *data, int size)
...
+{
+    return s->gctx->class->buffer_download(s, data, size, offset);
+}
+
+int ngli_buffer_map(struct buffer *s, int size, uint32_t offset, void **data)
+{
+    return s->gctx->class->buffer_map(s, size, offset, data);
+}
+
+void ngli_buffer_unmap(struct buffer* s)
```
cboesch-gpsw, line 52: style

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,49 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
     return 0;
 }
 
+int ngli_buffer_gl_download(struct buffer* s, void* data, uint32_t size, uint32_t offset)
```
cboesch-gpsw, line 75: style

libnodegl/buffer.h
```diff
@@ -39,6 +41,9 @@ struct buffer {
 struct buffer *ngli_buffer_create(struct gctx *gctx);
 int ngli_buffer_init(struct buffer *s, int size, int usage);
 int ngli_buffer_upload(struct buffer *s, const void *data, int size);
+int ngli_buffer_download(struct buffer* s, void* data, uint32_t size, uint32_t offset);
+int ngli_buffer_map(struct buffer *s, int size, uint32_t offset, void** data);
+void ngli_buffer_unmap(struct buffer* s);
```
cboesch-gpsw, line 46: style on all 3 prototypes

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,49 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+{
+    void* src;
```
cboesch-gpsw, line 77: style

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,49 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+    int ret = ngli_buffer_gl_map(s, size, offset, &src);
+    if (ret < 0) 
+        return ret;
+    memcpy(data, src, size);
+    ngli_buffer_gl_unmap(s);
+    return 0;
+}
+
+int ngli_buffer_gl_map(struct buffer *s, int size, uint32_t offset, void** data)
```
cboesch-gpsw, line 86: style

cboesch-gpsw: OK for keeping `void**`; but please name it `datap`, that's usually the naming convention we use for double dimension pointers.

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,49 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+
+    if (!(gl->features & NGLI_FEATURE_MAP_BUFFER)) {
+        LOG(WARNING, "context does not support glMapBufferRange");
+        return NGL_ERROR_UNSUPPORTED;
+    }
+
+    ngli_assert(data);
+    ngli_glBindBuffer(gl, GL_ARRAY_BUFFER, s_priv->id);
+    //TODO: use persistent mapped buffers, asynchronous
+    *data = ngli_glMapBufferRange(gl, GL_ARRAY_BUFFER, offset, size, GL_MAP_READ_BIT | GL_MAP_WRITE_BIT); //TODO: pass access flags
```
cboesch-gpsw, line 100: style: please add a space between `//` and `TODO`

libnodegl/buffer_gl.c
```diff
@@ -71,6 +72,49 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
...
+
+    return 0;
+}
+
+void ngli_buffer_gl_unmap(struct buffer* s) {
```
cboesch-gpsw, line 105: style (`buffer*` + `{`)

libnodegl/buffer_gl.h
```diff
@@ -35,6 +35,9 @@ struct gctx;
 struct buffer *ngli_buffer_gl_create(struct gctx *gctx);
 int ngli_buffer_gl_init(struct buffer *s, int size, int usage);
 int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size);
+int ngli_buffer_gl_download(struct buffer* s, void* data, uint32_t size, uint32_t offset);
+int ngli_buffer_gl_map(struct buffer *s, int size, uint32_t offset, void** data);
+void ngli_buffer_gl_unmap(struct buffer* s);
```
cboesch-gpsw, line 40: style on all 3 prototypes

libnodegl/gctx.h
```diff
@@ -63,6 +63,9 @@ struct gctx_class {
     struct buffer *(*buffer_create)(struct gctx *ctx);
     int (*buffer_init)(struct buffer *s, int size, int usage);
     int (*buffer_upload)(struct buffer *s, const void *data, int size);
+    int (*buffer_download)(struct buffer* s, void* data, uint32_t size, uint32_t offset);
+    int (*buffer_map)(struct buffer *s, int size, uint32_t offset, void** data);
+    void (*buffer_unmap)(struct buffer* s);
```
cboesch-gpsw, line 68: style on all 3 prototypes

libnodegl/gctx_gl.c
```diff
@@ -521,6 +521,9 @@ const struct gctx_class ngli_gctx_gl = {
     .buffer_create = ngli_buffer_gl_create,
     .buffer_init   = ngli_buffer_gl_init,
     .buffer_upload = ngli_buffer_gl_upload,
+    .buffer_download = ngli_buffer_gl_download,
+    .buffer_map = ngli_buffer_gl_map,
+    .buffer_unmap = ngli_buffer_gl_unmap,
```
cboesch-gpsw, line 526: style: vertical align

libnodegl/features.h
```diff
@@ -53,6 +53,7 @@
 #define NGLI_FEATURE_SOFTWARE                     (1 << 28)
 #define NGLI_FEATURE_UINT_UNIFORMS                (1 << 29)
 #define NGLI_FEATURE_EGL_ANDROID_GET_IMAGE_NATIVE_CLIENT_BUFFER (1 << 30)
+#define NGLI_FEATURE_MAP_BUFFER                   (1ull << 31)
```
cboesch-gpsw, line 56: we use capitalized notation in other places

libnodegl/glcontext.h
cboesch-gpsw: Also we probably want `uint64_t`

ngl-tools/Makefile
```diff
@@ -67,6 +67,34 @@ PYTHON_LDLIBS_EMBED-yes = $(shell python$(PYTHON_MAJOR)-config --ldflags --embed
...
+# As a result, `python3-config --cflags` will return the `-specs=...`
+# flags, but `python3-config --ldflags` will not return the
+# complementary `-specs=...`.
+#
+    ifeq ($(TARGET_OS), Linux)
+        DISTRIB_ID := $(shell lsb_release -si 2>/dev/null)
+        ifeq ($(DISTRIB_ID),$(filter $(DISTRIB_ID),Ubuntu Debian))
+            PYTHON_CFLAGS += -fPIE
+        endif
+    endif
```
cboesch-gpsw, line 97: We don't indent in other makefiles, so better keep it consistent.

Also, you forgot to remove the chunk from the CI.

ubitux: Oh and the commit message and description need adjustment; probably something like "tools/build: workaround a Python linking bug specific to Ubuntu/Debian"

jmoguillansky-gpsw:
Yes strange compile error.  Maybe it has something to do with forward declarations.  Regardless, we don't need to compile in this code for other platforms

cboesch-gpsw:
> Yes strange compile error. Maybe it has something to do with forward declarations. Regardless, we don't need to compile in this code for other platforms

Yes but maybe we can fix it without so much ifdefery. Sure it fixes your issue, but it's just like with the `-fPIC`, it's important to understand the issue before trying to fix the symptoms. You didn't say on which environment and configuration you have this issue. Is it when compiling on Ubuntu? MacOS? Cross-compiling?

If I understand the issue correctly, to me it looks like you could address the issue by moving:

```c
#if defined(TARGET_ANDROID)
#include <android/native_window.h>
#include <android/hardware_buffer.h>
#endif
```

from `libnodegl/glcontext_egl.c` to `libnodegl/egl.h`, and making sure the following:

```c
#include <EGL/egl.h>
#include <EGL/eglext.h>
```

Is only included once (in the header).

Can you test that?

ngl-tools/Makefile
jmoguillansky-gpsw: Different platforms have different flags enabled in their toolchain.  I don't want to dive deep into details regarding this issue.  From my perspective: 1) we should try to avoid platform-specific workarounds / compile-flags.  We should choose a reference linux platform and focus on that build.  Any platform-specific workarounds can be a separate patch file.  This way we keep the main makefiles as clean as possible

cboesch-gpsw: > Different platforms have different flags enabled in their toolchain.

Yes but this is unrelated to the issue here. It's not because of a different configuration, it's because of a bug in the packaging causing an inconsistency between the compiler flags and linker flags.

> I don't want to dive deep into details regarding this issue.

The issue is already analyzed, please at least take the time to read the comment or the bug report.

> From my perspective: 1) we should try to avoid platform-specific workarounds / compile-flags.

Yes, as long as it's not a bug.

> We should choose a reference linux platform and focus on that build. Any platform-specific workarounds can be a separate patch file.

That's what we did until now: users with broken systems had to workaround by themselves.

> This way we keep the main makefiles as clean as possible

That's what we had. Now ideally users with broken systems can communicate with their upstream to push a fix for the issue. I don't see any other sane solution.

Anyway, see https://github.com/gopro/gopro-lib-node.gl/pull/88 for the patch.

mbouron-gpsw:
Closing this PR:
* glcontext_egl: guard ngli_eglGetNativeClientBufferANDROID() declaration, has been merged into master
* glfeatures: change flag to 64 bit, has been merged into master (with small changes, gctx->features has also been changed to uint64_t)
* tools/build: enable position independent, a corresponding commit has been merged into master by Clément
* glfeatures: add glMapBufferRange / glUnmapBuffer, has been merged into origin/vulkan
* buffer: add map/unmap/download functions, has been merged into origin/vulkan
* internal: add swapchain, has been merged into origin/vulkan

PR #76, opened 2020-09-09
NGLI Optimizations (GL + vulkan): add buffer map/unmap/download., add swapchain module

jmoguillansky-gpsw:


libnodegl/buffer_gl.c
```diff
@@ -73,8 +73,10 @@ int ngli_buffer_gl_upload(struct buffer *s, const void *data, int size)
 }
 
 int ngli_buffer_gl_download(struct buffer* s, void* data, uint32_t size, uint32_t offset) {
-    LOG(ERROR, "not supported");
-    return NGL_ERROR_UNSUPPORTED;
+    void* src;
+    ngli_buffer_gl_map(s, size, offset, &src);
```
mbouron-gpsw, line 77: return of ngli_buffer_gl_map should be checked here.

PR #78, opened 2020-09-17
tools: add framerate support

mbouron-gpsw:


ngl-tools/player.c
```diff
@@ -225,6 +242,14 @@ static int key_callback(SDL_Window *window, SDL_KeyboardEvent *event)
...
         update_time(clipi64(p->frame_ts + 10 * 1000000, 0, p->duration));
         break;
+    case SDLK_o:
+        p->paused = 1;
+        set_frame_index(p->frame_index - 1);
+        break;
+    case SDLK_p:
+        p->paused = 1;
+        set_frame_index(p->frame_index + 1);
+        break;
```
cboesch-gpsw, line 261: 1. Can you update `lasthover` so the progress bar pops up on seeking with `o`/`p` as well?
2. This doesn't handle overflow and underflow: we can go in negative and longer than the duration; not sure yet if we want clamping or looping though (I'd go for clamping just like we do with normal seeking, and eventually later on add first/last frame bindings, but that's just me)

ngl-tools/common.c
```diff
@@ -36,6 +37,14 @@ int64_t gettime(void)
     return 1000000 * (int64_t)tv.tv_sec + tv.tv_usec;
 }
 
+int64_t gettime_relative(void)
+{
+    struct timespec ts;
+    int ret = clock_gettime(CLOCK_MONOTONIC, &ts);
+    return ret != 0 ? 0 : ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+
```
cboesch-gpsw, line 45: extra linebreak

ngl-tools/common.h
```diff
@@ -30,6 +30,7 @@
 #define DEFAULT_HEIGHT 360
 
 int64_t gettime(void);
+int64_t gettime_relative(void);
```
cboesch-gpsw, line 33: Do you mind adding in the commit message that `gettime_relative()` is appropriate typically for clocks and benchmarks because it is guaranteed monotonic and not prone to random clock jittering?

ngl-tools/player.c
```diff
@@ -118,7 +118,7 @@ static int screenshot(void)
     ret = ngl_configure(p->ngl, config);
     if (ret < 0)
         fprintf(stderr, "Could not configure node.gl for onscreen rendering\n");
-    p->clock_off = gettime() - p->frame_ts;
+    p->clock_off = gettime_relative() - p->frame_ts;
```
cboesch-gpsw, line 122: Maybe in another commit, but `gettime()` could/should be replaced in `ngl-render` as well.

ngl-tools/player.c
mbouron-gpsw: I would go with clamping too.

ngl-tools/common.c
mbouron-gpsw: Done.

ngl-tools/common.h
mbouron-gpsw: Done.

ngl-tools/player.c
mbouron-gpsw: Done.

ngl-tools/player.c
mbouron-gpsw: Done.

ngl-tools/player.c
```diff
@@ -415,6 +442,15 @@ int player_init(struct player *p, const char *win_title, struct ngl_node *scene,
...
     p->enable_ui = enable_ui;
+    p->framerate[0] = 60;
+    p->framerate[1] = 1;
+
+    if (!framerate[0] || !framerate[1]) {
+        fprintf(stderr, "Invalid framerate %d/%d\n", framerate[0], framerate[1]);
+        return -1;
+    }
+    memcpy(p->framerate, framerate, sizeof(p->framerate));
+    p->duration_i = llrint(p->duration_f * framerate[0] / (double)framerate[1]);
```
cboesch-gpsw, line 462: For some reason this is initialized to `0` in my test, so it breaks when using `o`/`p` (goes back to `0` all the time).

Also, don't you need to update it when the framerate is changed?

mbouron-gpsw: Update the handle_framerate function to update duration_i.

mbouron-gpsw: Updated handle_framerate() and handle_duration() to set duration_i according to the new framerate. This fixes the issue regarding duration_i being set to 0 on my side.

PR #80, opened 2020-09-18
tools/player: update progress bar UI

mbouron-gpsw:

PR #81, opened 2020-09-18
Small player fixes

cboesch-gpsw:


cboesch-gpsw:
Oups, this depends on https://github.com/gopro/gopro-lib-node.gl/pull/80

ngl-tools/player.c
```diff
@@ -626,6 +626,7 @@ void player_main_loop(void)
             case SDL_USEREVENT:
                 run = handle_map[event.user.code](event.user.data1) == 0;
                 free(event.user.data1);
+                p->text_last_frame_index = -1;
```
mbouron-gpsw, line 629: Shouldn't p->text_last_frame_index also be initialized to -1 in player_init() ?

cboesch-gpsw: Yep that's correct, and it's already true on `master`, so I added that in a separated commit.

mbouron-gpsw: Thanks.

mbouron-gpsw (approved):

PR #83, opened 2020-09-19
utils: remove PySide2 dependency from scene controllers

cboesch-gpsw:
We don't want the scene() decorator to pull PySide2 in the dependency
tree, because this prevents using Python scenes on system where
Qt/PySide2 are not well supported (this is typically the case on Windows
currently).

Note that the scene() decorator and associated objects probably belongs
into pynodegl as it is core to the Python API and relied on in the
ngl-tools.


jmoguillansky-gpsw (approved):
Thanks, this is very useful

mbouron-gpsw (approved):

PR #84, opened 2020-09-21
Restore Windows support

cboesch-gpsw:


libnodegl/log.c
```diff
@@ -58,6 +58,7 @@ static void default_callback(void *arg, int level, const char *filename, int ln,
     printf("%s[%s] %s:%d %s: %s%s\n", color_start,
            log_strs[level], filename, ln, fn, logline,
            color_end);
+    fflush(stdout);
```
mbouron-gpsw, line 61: Won't this degrade performance when log level is set to debug on others systems ? Wouldn't scoping it to Windows only be better ?

mbouron-gpsw:
Other than my remark regarding the use of fflush(), LGTM.

libnodegl/log.c
cboesch-gpsw: Well, the logging itself is in a pipe, other system are already flushing per line. Don't we want to force that specific behavior anyway? I mean, do we really want log to allow buffering many/infinite lines before showing them to the user?

mbouron-gpsw: Ok if it does not change anything without a pipe.

mbouron-gpsw (approved):

PR #86, opened 2020-09-21
gctx_gl: check capture_buffer before reading pixels

mbouron-gpsw:
ngli_rendertarget_read_pixels() must not be called with data set to
NULL. This commit fixes a potential crash when node.gl is configured to
render offscreen and no capture_buffer is set.

Fixes a regression introduced by
0be6c0b929dfeade2bca163f36eabe963726b1f4.


cboesch-gpsw (approved):
Should be fine, but I'm not a fan of calling a "capture" function when there is no capture.

PR #87, opened 2020-09-22
glcontext_wgl: do not create a puffer for offscreen rendering

mbouron-gpsw:
node.gl offscreen rendering relies on FBOs, creating a dedicated pixel
buffer context is not required and superfluous.


jmoguillansky-gpsw (approved):
Agreed: can just render to texture and download the contents.

cboesch-gpsw (approved):

PR #88, opened 2020-09-22
tools/build: automatically workaround a Python linking bug specific t…

cboesch-gpsw:
…o Ubuntu/Debian

The issue has been reported¹ months ago, and ignored.

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959945

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -30,28 +30,6 @@ jobs:
 
     - name: Build
       run: |
-        #
-        # This -fPIE is a workaround for the following link issue:
-        #     relocation R_X86_64_32 against `.rodata.str1.8' can not be used
-        #     when making a PIE object; recompile with -fPIE
-        #
-        # To summarize the issue:
-        #
-        # - On Python side, `python3-config` is a script generated from a
-        #   template using a global CFLAGS. But at the same time, the script
-        #   uses LIBS and SYSLIBS to construct the LDFLAGS (and not LDFLAGS for
-        #   some moronic reason).
-        # - On Debian/Ubuntu side, they decided to have the following injected:
-        #   CFLAGS="-specs=/usr/share/dpkg/no-pie-compile.specs"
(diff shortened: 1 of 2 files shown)


jmoguillansky-gpsw (approved):
Thanks!  Much appreciated

mbouron-gpsw (approved):

cboesch-gpsw:
I added a WIP/fixup commit: what workaround do you guys prefer? The vanilla patch or with the fixup?

jmoguillansky-gpsw:
are you talking about .github/workflows/ci_linux.yml ?  How do we use that?

cboesch-gpsw:
> are you talking about .github/workflows/ci_linux.yml ?

No, I'm talking about the "WIP this is a test" commit I added to the PR. It's an alternative way, probably more correct. Basically it emulates what would happen if Ubuntu/Debian fixed their packaging.

>  How do we use that?

The CI is the automation you can access in https://github.com/gopro/gopro-lib-node.gl/actions, or in the PR under "All checks have passed". But it's unrelated to what I was asking here.

jmoguillansky-gpsw:
I'm ok with either approach, no strong preference.

I think its best to hide these changes from the makefile, and put it in a patch file that patches the makefile on certain distribution.  Before we run make we can run a script to apply patches.  This way the makefile is cleaner.

cboesch-gpsw:
> I'm ok with either approach, no strong preference. I think its best to hide these changes from the makefile, and put it in a patch file that patches the makefile on certain distribution. Before we run make we can run a script to apply patches. This way the makefile is cleaner.

But what exactly would you be patching here?

jmoguillansky-gpsw:
Something like 

python-bug-workaround.patch:
ngl-tools/Makefile:
TOOLS += python serialize
+PYTHON_CFLAGS += ...
+PYTHON_LDFLAGS += ...

jmoguillansky-gpsw:
Yes but the problem with this approach is that we need to keep updating the patch when we update the makefile.  
Another option is to use some script command like "sed" to patch the makefile as a preprocessing step

PR #93, opened 2020-10-01
iovar precision

cboesch-gpsw:


jmoguillansky-gpsw (approved):

libnodegl/precision.h
```diff
@@ -0,0 +1,37 @@
...
+
+enum {
+    NGLI_PRECISION_AUTO,
+    NGLI_PRECISION_HIGH,
+    NGLI_PRECISION_MEDIUM,
+    NGLI_PRECISION_LOW,
+    NGLI_PRECISION_NB
+};
+
+extern const struct param_choices ngli_precision_choices;
```
mbouron-gpsw, line 35: Is it to make sure we don't end up with multiple definitions/declarations of ngli_precision_choices in the library ?

cboesch-gpsw: Yes, they would end up being duplicated in the doc as well.

mbouron-gpsw (approved):

tests/shape.py
```diff
@@ -27,11 +27,31 @@
 from pynodegl_utils.misc import scene
 from pynodegl_utils.toolbox.colors import COLORS
 from pynodegl_utils.toolbox.colors import get_random_color_buffer
+from pynodegl_utils.tests.cmp_cuepoints import test_cuepoints
 from pynodegl_utils.tests.cmp_fingerprint import test_fingerprint
 from pynodegl_utils.toolbox.shapes import equilateral_triangle_coords
 from pynodegl_utils.toolbox.grid import autogrid_simple
 
 
+@test_cuepoints(points=dict(bl=(-1, -1), br=(1, -1), tr=(1, 1), tl=(-1, 1), c=(0, 0)), nb_keyframes=1, tolerance=5)
```
mbouron-gpsw, line 36: Isn't nb_keyframes=1 the default ? is yes, can we remove it ?

tests/shape.py
```diff
@@ -27,11 +27,31 @@
...
+def shape_precision_iovar(cfg):
+    cfg.aspect_ratio = (1, 1)
+    vert = '''
+void main()
+{
+    ngl_out_pos = ngl_projection_matrix * ngl_modelview_matrix * ngl_position;
+    color = vec4((ngl_out_pos.xy + 1.0) * .5, 1.0 - (ngl_out_pos.x + 1.0) * .5 - (ngl_out_pos.y + 1.0) * .5, 1.0);
+}
+'''
+    frag = 'void main() { ngl_out_color = color; }'
```
mbouron-gpsw, line 47: nit: i would use a docstring declaration like for the vertex shader (it's easier to read (at least for me))

cboesch-gpsw: done

tests/shape.py
cboesch-gpsw: correct, removed

PR #98, opened 2020-10-06
tests/rtt: add rtt_load_attachment test

mbouron-gpsw:
Ensures that when a node rtt draw ends, we restore the previous
rendertarget without discarding its attachments.

### DIFF
--- a/tests/rtt.py
+++ b/tests/rtt.py
@@ -23,7 +23,9 @@
 import array
 import pynodegl as ngl
 from pynodegl_utils.misc import scene
+from pynodegl_utils.tests.cmp_cuepoints import test_cuepoints
 from pynodegl_utils.tests.cmp_fingerprint import test_fingerprint
+from pynodegl_utils.toolbox.colors import COLORS
 
 
 def _get_cube():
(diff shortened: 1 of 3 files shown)


tests/rtt.py
```diff
@@ -171,3 +173,30 @@ def rtt_function(cfg):
...
+    program = ngl.Program(vertex=cfg.get_vert('color'), fragment=cfg.get_frag('color'))
+    program.update_vert_out_vars(var_tex0_coord=ngl.IOVec2(), var_uvcoord=ngl.IOVec2())
+    background = ngl.Render(quad, program)
+    background.update_frag_resources(color=ngl.UniformVec4(value=COLORS['white']))
+
+    quad = ngl.Quad((-1, -1, 0), (2, 0, 0), (0, 2, 0))
+    program = ngl.Program(vertex=cfg.get_vert('color'), fragment=cfg.get_frag('color'))
+    program.update_vert_out_vars(var_tex0_coord=ngl.IOVec2(), var_uvcoord=ngl.IOVec2())
+    render = ngl.Render(quad, program)
+    render.update_frag_resources(color=ngl.UniformVec4(value=COLORS['orange']))
```
ubitux, line 191: Can you use a different name than `render` to prevent the confusion with the following `render`?

mbouron-gpsw: Done. I renamed the other render to foreground.

cboesch-gpsw (approved):

PR #99, opened 2020-10-06
Simplify graph view

cboesch-gpsw:


pynodegl-utils/pynodegl_utils/clock.py
```diff
@@ -20,55 +20,25 @@
...
-        self.reset_running_time()
-        self._running = True
-
-    def stop(self):
-        self._running = False
-
-    def is_running(self):
-        return self._running
-
     def reset_running_time(self):
```
mbouron-gpsw, line 37: Since this class is not a clock anymore, resetting the running time does not make sense. This function and its use should be removed. IMHO this class should be made private to graph_view.py and probably renamed to something other than Clock.

cboesch-gpsw: > Since this class is not a clock anymore, resetting the running time does not make sense. This function and its use should be removed.

Indeed, simplified.

> IMHO this class should be made private to graph_view.py

Done

> and probably renamed to something other than Clock.

I don't really know how to name it, but we may want to inline it in `GraphView` instead anyway. I added a note in the commit description.

mbouron-gpsw (approved):

PR #100, opened 2020-10-07
rtt: detect non optimal render passes

mbouron-gpsw:


libnodegl/node_rtt.c
```diff
@@ -119,13 +119,40 @@ static int rtt_init(struct ngl_node *node)
...
+        const struct ngl_node *child = children[i];
+        if (child->class->id == NGL_NODE_RENDERTOTEXTURE ||
+            child->class->id == NGL_NODE_COMPUTE) {
+            info->has_rtt_or_compute = 1;
+            continue;
+        } else if (child->class->id == NGL_NODE_RENDER) {
+            info->render_counts[info->has_rtt_or_compute ? 0 : 1] += 1;
+            continue;
+        }
+        get_renderpass_children_info(child, info);
```
cboesch-gpsw, line 140: nit: you can remove the 2 `continue` if you use a `else` here

libnodegl/node_rtt.c
```diff
@@ -119,13 +119,40 @@ static int rtt_init(struct ngl_node *node)
...
+    for (int i = 0; i < ngli_darray_count(&node->children); i++) {
+        const struct ngl_node *child = children[i];
+        if (child->class->id == NGL_NODE_RENDERTOTEXTURE ||
+            child->class->id == NGL_NODE_COMPUTE) {
+            info->has_rtt_or_compute = 1;
+            continue;
+        } else if (child->class->id == NGL_NODE_RENDER) {
+            info->render_counts[info->has_rtt_or_compute ? 0 : 1] += 1;
+            continue;
+        }
```
cboesch-gpsw, line 139: The logic is not obvious to me. Unless I'm mistaken, it will behave like this:

- `render`-`render`-`compute/rtt` → `render_count=[0,2]` (no warning)
- `compute/rtt`-`render`-`render` → `render_count=[2,0]` (no warning)
- `render`-`compute/rtt`-`render` → `render_count=[1,1]` (warning)
-`compute/rtt`-`render`-`compute/rtt` → `render_count=[1,0]` (no warning)

1. is it OK warning wise (I'm thinking in particular about the last one)?
2. should the render "counts" be swapped somehow for readability?
3. do you need the "counter"? (I'm not sure exactly how I would name what it actually counts)
4. don't you want to just warn when there is > 1 (or maybe > 0) directions flips?

libnodegl/node_rtt.c
```diff
@@ -119,13 +119,40 @@ static int rtt_init(struct ngl_node *node)
...
 {
     struct ngl_ctx *ctx = node->ctx;
     struct gctx *gctx = ctx->gctx;
     struct rnode *rnode = ctx->rnode_pos;
     struct rtt_priv *s = node->priv_data;
 
+    struct renderpass_children_info info = {0};
+    get_renderpass_children_info(s->child, &info);
+    if (info.render_counts[0] && info.render_counts[1])
+        LOG(WARNING, "the underlying render pass might not be optimal as it contains a rtt or compute node in the middle of it");
```
cboesch-gpsw, line 154: Note: you're already aware of it but we need to warn only in a debug mode.

libnodegl/pass.c
```diff
@@ -693,13 +693,25 @@ int ngli_pass_exec(struct pass *s)
...
+        };
+
         if (s->indices_buffer)
             ngli_pipeline_draw_indexed(pipeline, s->indices_buffer, s->indices_format, s->nb_indices, s->nb_instances);
         else
             ngli_pipeline_draw(pipeline, s->nb_vertices, s->nb_instances);
-    else
+    } else {
         ngli_pipeline_dispatch(pipeline, params->nb_group_x, params->nb_group_y, params->nb_group_z);
+    }
```
cboesch-gpsw, line 714: this could go in a dedicated commit, it's typically a potential fix for a compiler warning

libnodegl/pass.c
```diff
@@ -693,13 +693,25 @@ int ngli_pass_exec(struct pass *s)
...
+        if (ctx->bind_current_rendertarget) {
+            struct gctx *gctx = ctx->gctx;
+            ngli_gctx_set_rendertarget(gctx, ctx->current_rendertarget);
+            if (ctx->clear_current_rendertarget) {
+                ngli_gctx_clear_color(gctx);
+                ngli_gctx_clear_depth_stencil(gctx);
+            }
+            ctx->bind_current_rendertarget = 0;
+            ctx->clear_current_rendertarget = 0;
+        };
```
cboesch-gpsw, line 706: trailing `;`

libnodegl/node_rtt.c
```diff
@@ -376,9 +376,12 @@ static void rtt_draw(struct ngl_node *node)
...
 
-    struct rendertarget *rt = s->rt;
-    struct rendertarget *prev_rt = ngli_gctx_get_rendertarget(gctx);
-    ngli_gctx_set_rendertarget(gctx, rt);
+    struct rendertarget *prev_rendertarget = ctx->current_rendertarget;
+    int prev_clear = ctx->clear_current_rendertarget;
+
+    ctx->current_rendertarget = s->rt;
+    ctx->bind_current_rendertarget = 1;
+    ctx->clear_current_rendertarget = !(s->features & FEATURE_NO_CLEAR);
```
cboesch-gpsw, line 384: Would you mind moving  this closer to `ngli_node_draw(s->child)`? It's not obvious at first sight which following code may change them. Having them close to the draw calls immediately shows that one of the children might do that.

libnodegl/node_rtt.c
```diff
@@ -392,20 +395,28 @@ static void rtt_draw(struct ngl_node *node)
...
     ngli_node_draw(s->child);
 
+    if (ctx->bind_current_rendertarget) {
+        ngli_gctx_set_rendertarget(gctx, ctx->current_rendertarget);
+        if (ctx->clear_current_rendertarget) {
+            ngli_gctx_clear_color(gctx);
+            ngli_gctx_clear_depth_stencil(gctx);
+        }
+        ctx->bind_current_rendertarget = 0;
+        ctx->clear_current_rendertarget = 0;
```
cboesch-gpsw, line 407: Is this needed? For the `rendertarget_resolve()` below maybe?

I'm asking because it's set to `1` just a little after.

libnodegl/pass.c
```diff
@@ -693,13 +693,25 @@ int ngli_pass_exec(struct pass *s)
...
+    if (s->pipeline_type == NGLI_PIPELINE_TYPE_GRAPHICS) {
+        if (ctx->bind_current_rendertarget) {
+            struct gctx *gctx = ctx->gctx;
+            ngli_gctx_set_rendertarget(gctx, ctx->current_rendertarget);
+            if (ctx->clear_current_rendertarget) {
+                ngli_gctx_clear_color(gctx);
+                ngli_gctx_clear_depth_stencil(gctx);
+            }
+            ctx->bind_current_rendertarget = 0;
+            ctx->clear_current_rendertarget = 0;
```
cboesch-gpsw, line 705: Any reason not to factor that out with the other identical chunk in `node_rtt`?

libnodegl/node_rtt.c
mbouron-gpsw: It is not needed, it is here for consistency with pass.c (and we properly set the state after the bind/clear operation) even though it is set to 1 later on.

libnodegl/pass.c
mbouron-gpsw: Done.

libnodegl/pass.c
mbouron-gpsw: Done.

libnodegl/pass.c
mbouron-gpsw: I find it easier to read this way because we know exactly what is happening. Also, the clear calls and its state will disappear in an upcoming commit that changes the rendertarget API making this block something like that `
if (ctx->bind_current_rendertarget) {
            struct gctx *gctx = ctx->gctx;
            ngli_gctx_set_rendertarget(gctx, ctx->current_rendertarget);
            ctx->bind_current_rendertarget = 0;
}
`

libnodegl/node_rtt.c
mbouron-gpsw: Done.

libnodegl/node_rtt.c
mbouron-gpsw: Done.

libnodegl/node_rtt.c
mbouron-gpsw: 1. it is OK warning wise
2. I don't think it needs to be inversed. [0] is the counter of renders before a potential interruption, [1] is the counter of renders after a potential interruption
3. I don't need the counter, I need to know if there is at least one render before or after a potential interruption of the underlying renderpass.
4. I really need to warn if there at least one render before after an interruption of the underlying render pass (hence the

mbouron-gpsw: Regarding 2. the ternary expression was inversed in the original commit ``` c
info->render_counts[info->has_rtt_or_compute ? 0 : 1] += 1;
```
This has been fixed.

cboesch-gpsw: The ternary is not needed but I guess you want to keep it explicit.

cboesch-gpsw (approved):
LGTM

PR #102, opened 2020-10-12
rendertarget: add load and store operations

mbouron-gpsw:
Also moves from ngli_gctx_set_rendertarget() to ngli_gctx_{begin,end}_render_pass().

### DIFF
--- a/libnodegl/gctx.c
+++ b/libnodegl/gctx.c
@@ -115,14 +115,19 @@ void ngli_gctx_transform_projection_matrix(struct gctx *s, float *dst)
     s->class->transform_projection_matrix(s, dst);
 }
 
+void ngli_gctx_begin_render_pass(struct gctx *s, struct rendertarget *rt)
+{
+    s->class->begin_render_pass(s, rt);
+}
+
+void ngli_gctx_end_render_pass(struct gctx *s)
+{
+    s->class->end_render_pass(s);
+}
+
 void ngli_gctx_get_rendertarget_uvcoord_matrix(struct gctx *s, float *dst)
 {
     s->class->get_rendertarget_uvcoord_matrix(s, dst);
-}
-
-void ngli_gctx_set_rendertarget(struct gctx *s, struct rendertarget *rt)
-{
-    s->class->set_rendertarget(s, rt);
 }
 
 struct rendertarget *ngli_gctx_get_rendertarget(struct gctx *s)
--- a/libnodegl/gctx.c
+++ b/libnodegl/gctx.c
@@ -155,26 +155,6 @@ void ngli_gctx_get_scissor(struct gctx *s, int *scissor)
     s->class->get_scissor(s, scissor);
 }
 
-void ngli_gctx_set_clear_color(struct gctx *s, const float *color)
-{
-    s->class->set_clear_color(s, color);
-}
-
-void ngli_gctx_get_clear_color(struct gctx *s, float *color)
-{
-    s->class->get_clear_color(s, color);
-}
-
-void ngli_gctx_clear_color(struct gctx *s)
-{
-    s->class->clear_color(s);
-}
-
-void ngli_gctx_clear_depth_stencil(struct gctx *s)
-{
-    s->class->clear_depth_stencil(s);
-}
-
 int ngli_gctx_get_preferred_depth_format(struct gctx *s)
 {
     return s->class->get_preferred_depth_format(s);
--- a/libnodegl/gctx.c
+++ b/libnodegl/gctx.c
@@ -175,11 +175,6 @@ void ngli_gctx_clear_depth_stencil(struct gctx *s)
     s->class->clear_depth_stencil(s);
 }
 
-void ngli_gctx_invalidate_depth_stencil(struct gctx *s)
-{
-    s->class->invalidate_depth_stencil(s);
-}
-
 int ngli_gctx_get_preferred_depth_format(struct gctx *s)
 {
     return s->class->get_preferred_depth_format(s);
--- a/libnodegl/gctx.c
+++ b/libnodegl/gctx.c
@@ -75,7 +75,7 @@ int ngli_gctx_draw(struct gctx *s, struct ngl_node *scene, double t)
 
     if (scene) {
         struct ngl_ctx *ctx = scene->ctx;
-        struct rendertarget *rt = ngli_gctx_get_rendertarget(s);
+        struct rendertarget *rt = ngli_gctx_get_default_rendertarget(s);
         ctx->available_rendertargets[0] = rt;
         ctx->available_rendertargets[1] = rt;
         ctx->current_rendertarget = rt;
@@ -130,9 +130,9 @@ void ngli_gctx_get_rendertarget_uvcoord_matrix(struct gctx *s, float *dst)
     s->class->get_rendertarget_uvcoord_matrix(s, dst);
 }
 
-struct rendertarget *ngli_gctx_get_rendertarget(struct gctx *s)
+struct rendertarget *ngli_gctx_get_default_rendertarget(struct gctx *s)
 {
-    return s->class->get_rendertarget(s);
+    return s->class->get_default_rendertarget(s);
 }
 
 const struct rendertarget_desc *ngli_gctx_get_default_rendertarget_desc(struct gctx *s)
--- a/libnodegl/gctx.c
+++ b/libnodegl/gctx.c
@@ -75,9 +75,11 @@ int ngli_gctx_draw(struct gctx *s, struct ngl_node *scene, double t)
 
     if (scene) {
         struct ngl_ctx *ctx = scene->ctx;
-        ctx->current_rendertarget = ngli_gctx_get_rendertarget(s);
+        struct rendertarget *rt = ngli_gctx_get_rendertarget(s);
+        ctx->available_rendertargets[0] = rt;
+        ctx->available_rendertargets[1] = rt;
+        ctx->current_rendertarget = rt;
         ctx->bind_current_rendertarget = 0;
-        ctx->clear_current_rendertarget = 0;
         LOG(DEBUG, "draw scene %s @ t=%f", scene->label, t);
         ngli_node_draw(scene);
     }
--- a/libnodegl/gctx.c
+++ b/libnodegl/gctx.c
@@ -79,7 +79,7 @@ int ngli_gctx_draw(struct gctx *s, struct ngl_node *scene, double t)
         ctx->available_rendertargets[0] = rt;
         ctx->available_rendertargets[1] = rt;
         ctx->current_rendertarget = rt;
-        ctx->bind_current_rendertarget = 0;
+        ctx->begin_render_pass = 0;
         LOG(DEBUG, "draw scene %s @ t=%f", scene->label, t);
         ngli_node_draw(scene);
     }
--- a/libnodegl/gctx_gl.c
+++ b/libnodegl/gctx_gl.c
@@ -296,8 +296,6 @@ static int gl_init(struct gctx *s)
     const GLint scissor[] = {0, 0, gl->width, gl->height};
     ngli_gctx_set_scissor(s, scissor);
 
-    ngli_gctx_set_clear_color(s, config->clear_color);
-
     return 0;
 }
 
@@ -472,50 +470,6 @@ static void gl_get_scissor(struct gctx *s, int *scissor)
     memcpy(scissor, &s_priv->scissor, sizeof(s_priv->scissor));
 }
 
-static void gl_set_clear_color(struct gctx *s, const float *color)
-{
-    struct gctx_gl *s_priv = (struct gctx_gl *)s;
-    struct glcontext *gl = s_priv->glcontext;
-    memcpy(s_priv->clear_color, color, sizeof(s_priv->clear_color));
-    ngli_glClearColor(gl, color[0], color[1], color[2], color[3]);
-}
-
-static void gl_get_clear_color(struct gctx *s, float *color)
-{
-    struct gctx_gl *s_priv = (struct gctx_gl *)s;
-    memcpy(color, &s_priv->clear_color, sizeof(s_priv->clear_color));
-}
-
-static void gl_clear_color(struct gctx *s)
-{
-    struct gctx_gl *s_priv = (struct gctx_gl *)s;
-    struct glcontext *gl = s_priv->glcontext;
-    struct glstate *glstate = &s_priv->glstate;
-
-    const int scissor_test = glstate->scissor_test;
-    ngli_glDisable(gl, GL_SCISSOR_TEST);
-
-    ngli_glClear(gl, GL_COLOR_BUFFER_BIT);
-
-    if (scissor_test)
-        ngli_glEnable(gl, GL_SCISSOR_TEST);
-}
-
-static void gl_clear_depth_stencil(struct gctx *s)
-{
-    struct gctx_gl *s_priv = (struct gctx_gl *)s;
-    struct glcontext *gl = s_priv->glcontext;
-    struct glstate *glstate = &s_priv->glstate;
-
-    const int scissor_test = glstate->scissor_test;
-    ngli_glDisable(gl, GL_SCISSOR_TEST);
-
-    ngli_glClear(gl, GL_DEPTH_BUFFER_BIT | GL_STENCIL_BUFFER_BIT);
-
-    if (scissor_test)
-        ngli_glEnable(gl, GL_SCISSOR_TEST);
-}
-
 static int gl_get_preferred_depth_format(struct gctx *s)
 {
     return NGLI_FORMAT_D16_UNORM;
@@ -546,10 +500,6 @@ const struct gctx_class ngli_gctx_gl = {
     .get_viewport             = gl_get_viewport,
     .set_scissor              = gl_set_scissor,
     .get_scissor              = gl_get_scissor,
-    .set_clear_color          = gl_set_clear_color,
-    .get_clear_color          = gl_get_clear_color,
-    .clear_color              = gl_clear_color,
-    .clear_depth_stencil      = gl_clear_depth_stencil,
     .get_preferred_depth_format = gl_get_preferred_depth_format,
     .get_preferred_depth_stencil_format = gl_get_preferred_depth_stencil_format,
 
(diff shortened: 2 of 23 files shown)


libnodegl/gctx_gl.c
```diff
@@ -170,9 +167,17 @@ static int offscreen_rendertarget_init(struct gctx *s)
         .colors[0] = {
             .attachment     = config->samples ? s_priv->ms_color : s_priv->color,
             .resolve_target = config->samples ? s_priv->color    : NULL,
+            .load_op        = NGLI_LOAD_OP_LOAD,
+            .load_value[0]  = config->clear_color[0],
```
jmoguillansky-gpsw, line 171: should this be called clear_value?

libnodegl/rendertarget_gl.c
```diff
@@ -244,10 +299,34 @@ int ngli_rendertarget_gl_init(struct rendertarget *s, const struct rendertarget_
                 draw_buffers[-1] = GL_COLOR_ATTACHMENT0 + i;
             }
 
+            s_priv->clear = clear_buffers;
```
cboesch-gpsw, line 302: `NGLI_FEATURE_DRAW_BUFFERS` implies the presence of `glClearBuffer*`?

libnodegl/rendertarget_gl.h
```diff
@@ -32,12 +32,19 @@ struct rendertarget_gl {
     GLuint prev_id;
     GLenum draw_buffers[NGLI_MAX_COLOR_ATTACHMENTS];
     GLenum blit_draw_buffers[NGLI_MAX_COLOR_ATTACHMENTS*(NGLI_MAX_COLOR_ATTACHMENTS+1)/2];
+    GLenum clear_flags;
+    GLenum invalidate_attachments[NGLI_MAX_COLOR_ATTACHMENTS + 2];
```
cboesch-gpsw, line 36: a short comment on the `+2` would be welcome ("// +2 for depth+stencil" or something like that I'm assuming)

libnodegl/rendertarget.h
```diff
@@ -27,6 +27,17 @@
 
 #define NGLI_MAX_COLOR_ATTACHMENTS 8
 
+enum {
+    NGLI_LOAD_OP_DONTCARE,
```
cboesch-gpsw, line 31: Is it OK to have `DONTCARE` on `0`? No need for an unset/unknown value?

libnodegl/node_rtt.c
```diff
@@ -314,9 +328,8 @@ static int rtt_prefetch(struct ngl_node *node)
             if (ret < 0)
                 return ret;
             rt_params.depth_stencil.attachment = depth;
-
-            if (!(s->features & FEATURE_NO_CLEAR))
-                s->invalidate_depth_stencil = 1;
```
cboesch-gpsw, line 324:
```
libnodegl/node_rtt.c:#define FEATURE_NO_CLEAR    (1 << 2)
libnodegl/node_rtt.c:        {"no_clear",FEATURE_NO_CLEAR,.desc=NGLI_DOCSTRING("not cleared between draws (non-deterministic)")},
```

What do we do about these?

libnodegl/node_rtt.c
```diff
@@ -260,9 +262,17 @@ static int rtt_prefetch(struct ngl_node *node)
...
+                float *load_value = rt_params.colors[rt_params.nb_colors].load_value;
+                memcpy(load_value, s->clear_color, sizeof(s->clear_color));
+                rt_params.colors[rt_params.nb_colors].store_op = NGLI_STORE_OP_STORE;
             } else {
                 rt_params.colors[rt_params.nb_colors].attachment = texture;
                 rt_params.colors[rt_params.nb_colors].attachment_layer = j;
+                rt_params.colors[rt_params.nb_colors].load_op = NGLI_LOAD_OP_CLEAR;
+                float *load_value = rt_params.colors[rt_params.nb_colors].load_value;
+                memcpy(load_value, s->clear_color, sizeof(s->clear_color));
+                rt_params.colors[rt_params.nb_colors].store_op = NGLI_STORE_OP_STORE;
```
cboesch-gpsw, line 275: can't you move this chunk out of the `if`/`else`? I don't see any difference

libnodegl/rendertarget_gl.c
```diff
@@ -244,10 +299,34 @@ int ngli_rendertarget_gl_init(struct rendertarget *s, const struct rendertarget_
...
             s_priv->resolve = resolve_draw_buffers;
         }
     }
 
+    for (int i = 0; i < params->nb_colors; i++) {
+        const struct attachment *color = &params->colors[i];
+        if (color->load_op == NGLI_LOAD_OP_DONTCARE ||
+            color->load_op == NGLI_LOAD_OP_CLEAR) {
+            s_priv->clear_flags |= GL_COLOR_BUFFER_BIT;
+        }
```
cboesch-gpsw, line 312: Probably a naive question, but why would you set the clear flag if you don't care?

mbouron-gpsw: In OpenGL clearing the attachments before any operation on the framebuffer tells the driver it does not need to load the attachment (saving bandwidth). This is our only way to tell we don't want to load the attachment unlike in Vulkan where you can specify explicitely DONTCARE or LOAD.

libnodegl/node_rtt.c
mbouron-gpsw: There is no difference between the two, I made it this way for readability since I am setting the attachment_desc fields in the order they are declared in the structure.

libnodegl/node_rtt.c
mbouron-gpsw: I forgot to take it into account, two solutions:
* We can remove this feature if we don't use it
* Take this flag into account

libnodegl/rendertarget.h
mbouron-gpsw: It is OK to have it set to DONTCARE IMHO, users need to explicitly specify what they do with their attachment (and the default is DONTCARE).

mbouron-gpsw: Vulkan have LOAD/STORE on 0, will reorder the enum.

libnodegl/node_rtt.c
cboesch-gpsw: I'm fine with both; I think the `no_clear` purpose was for some slit scan and other non-deterministic experiments?

libnodegl/node_rtt.c
cboesch-gpsw: I usually tend to look for differences in similar blocks, and it took me some time because I didn't find any, so I'm not sure I agree with the readability argument. Not a blocking comment though, so feel free to ignore and resolve this conversation.

libnodegl/rendertarget_gl.c
cboesch-gpsw: Ah yeah I remember that indeed. Don't you think a comment explaining this optimization would be welcome here, or is it really common knowledge?

libnodegl/rendertarget_gl.c
mbouron-gpsw: ARB_draw_buffers does introduce glClearBuffer*, glClearBuffer* functions have been introduced later on.
It is available at the same version of OpenGL/OpenGLES (3.0+), so i took this shortcut, I should probably have a proper dedicated feature for it.

tests/rtt.py
```diff
@@ -139,7 +139,8 @@ def _get_rtt_scene(cfg, features='depth', texture_ds_format=None, samples=0, mip
         [texture],
         features=features,
         depth_texture=texture_depth,
-        samples=samples
+        samples=samples,
+        clear_color=(0, 0, 0, 1),
```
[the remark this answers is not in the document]
mbouron-gpsw: It shouldn't be needed in other places.

libnodegl/gctx.c
```diff
@@ -173,11 +175,6 @@ void ngli_gctx_clear_depth_stencil(struct gctx *s)
     s->class->clear_depth_stencil(s);
 }
 
-void ngli_gctx_invalidate_depth_stencil(struct gctx *s)
-{
-    s->class->invalidate_depth_stencil(s);
-}
-
```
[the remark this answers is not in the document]
mbouron-gpsw: I'll see if I can split it, i don't want to leave a function if it does not work properly with the new API (even if it's removed in the next commit).

libnodegl/node_rtt.c
mbouron-gpsw: Yes its purpose was to do experiments like the slit scan effect, I don't mind keeping it or removing it.

tests/rtt.py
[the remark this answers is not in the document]
jmoguillansky-gpsw: can we have a default param for clear_color (for rtt node)?

libnodegl/rendertarget_gl.c
jmoguillansky-gpsw: Do we need glClearBuffer or can we just use glClear?

libnodegl/gctx_gl.c
```diff
@@ -170,9 +167,17 @@ static int offscreen_rendertarget_init(struct gctx *s)
         .colors[0] = {
             .attachment     = config->samples ? s_priv->ms_color : s_priv->color,
             .resolve_target = config->samples ? s_priv->color    : NULL,
+            .load_op        = NGLI_LOAD_OP_LOAD,
```
jmoguillansky-gpsw, line 170: maybe we should put a comment for now? (TODO: optimize)

libnodegl/gctx_gl.c
```diff
@@ -319,9 +321,15 @@ static int gl_resize(struct gctx *s, int width, int height, const int *viewport)
 
 static int gl_pre_draw(struct gctx *s, double t)
 {
-    ngli_gctx_clear_color(s);
-    ngli_gctx_clear_depth_stencil(s);
+    struct gctx_gl *s_priv = (struct gctx_gl *)s;
+    struct glcontext *gl = s_priv->glcontext;
+    const struct ngl_config *config = &s->config;
 
+    ngli_gctx_begin_render_pass(s, config->offscreen ? s_priv->rt : NULL);
```
jmoguillansky-gpsw, line 328: what if the first node we draw is an RTT node?  Do we know at beginning of frame draw which renderpass to use?

libnodegl/gctx_gl.c
```diff
@@ -396,31 +406,48 @@ static void gl_get_rendertarget_uvcoord_matrix(struct gctx *s, float *dst)
...
     struct gctx_gl *s_priv = (struct gctx_gl *)s;
-    return s_priv->rendertarget;
-}
 
-static const struct rendertarget_desc *gl_get_default_rendertarget_desc(struct gctx *s)
-{
-    struct gctx_gl *s_priv = (struct gctx_gl *)s;
-    return &s_priv->default_rendertarget_desc;
+    if (s_priv->rendertarget) {
+        ngli_rendertarget_gl_resolve(s_priv->rendertarget);
```
jmoguillansky-gpsw, line 446: maybe we don't have to do explicit resolve?  

https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_multisampled_render_to_texture.txt

I think GL might be able to do an implicit resolve

libnodegl/gctx_gl.c
```diff
@@ -396,31 +406,48 @@ static void gl_get_rendertarget_uvcoord_matrix(struct gctx *s, float *dst)
...
+        ngli_rendertarget_gl_invalidate(s_priv->rendertarget);
```
jmoguillansky-gpsw, line 447: does this indicate to discard the contents?  Maybe we should have a check "if store-op == dont-care"?

libnodegl/glwrappers.h
```diff
@@ -679,9 +691,9 @@ static inline void ngli_glScissor(const struct glcontext *gl, GLint x, GLint y,
     check_error_code(gl, "glScissor");
 }
 
-static inline void ngli_glShaderBinary(const struct glcontext *gl, GLsizei count, const GLuint * shaders, GLenum binaryformat, const void * binary, GLsizei length)
+static inline void ngli_glShaderBinary(const struct glcontext *gl, GLsizei count, const GLuint * shaders, GLenum binaryFormat, const void * binary, GLsizei length)
 {
-    gl->funcs.ShaderBinary(count, shaders, binaryformat, binary, length);
+    gl->funcs.ShaderBinary(count, shaders, binaryFormat, binary, length);
```
jmoguillansky-gpsw, line 696: what binary format are we using ?  SPIRV?

libnodegl/rendertarget_gl.c
mbouron-gpsw: We need glClearBuffer* as the new API can specify a different clear color for each color attachment like in Vulkan.

tests/rtt.py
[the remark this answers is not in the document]
mbouron-gpsw: Yes, the commit makes it default to (0, 0, 0, 0) (as stated in the commit description).

libnodegl/gctx.c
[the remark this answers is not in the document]
mbouron-gpsw: We don't need this function anymore, store_op is set to DONT_CARE where we can, and gl invalidate framebuffer is called internally in gl_set_rendertarget(), then in ngli_gctx_end_render_pass() in the next commit.

libnodegl/rendertarget_gl.h
mbouron-gpsw: invalidate_attachments is an array containing all the attachments to invalidate, it is prepared at init time so we don't have to do it later on in the invalidate function. ngli_rendertarget_gl_invalidate() is called unconditionally in set_rendertarget(), if the array is empty the invalidate operation is a no-op.

libnodegl/gctx_gl.c
mbouron-gpsw: ngli_rendertarget_gl_invalidate() only invalidates the attachements with store-op set to dont-care.

libnodegl/glwrappers.h
mbouron-gpsw: We don't use glShaderBinary(). I guess we can remove it from the wrapper. Thanks for pointing that out.

mbouron-gpsw: Maybe we should also remove the other function that are not used. It should be part of another PR. Regarding this change, it has been produced after updating the gl wrapper to introduce the new glClearBuffer* functions.

libnodegl/gctx_gl.c
mbouron-gpsw: We don't know that yet, the proper solution is to crawl the tree (like we do in node_rtt) to know what we will need to do: one render pass (clear+store) or two render passes (clear+store, load+store) and then tell gctx to initialize the render pass(es). IMHO, we also need to move the render pass initialization away from the gctx initialization to do that properly.

What we can do now to improve the current situation on OpenGL is to have a flag to indicate that we need to do an explicit clear when binding the default framebuffer for the first time (in pass.c). It won't be optimal for the Vulkan backend (as we won't be able to specify load-op to DONT_CARE/CLEAR). IMHO, it is better to spend time on the proper solution to improve both OpenGL and Vulkan (even though it's a lot more work).

libnodegl/gctx_gl.c
mbouron-gpsw: Thanks, I did not know this extension existed. I will try to enable its use in a later PR.

libnodegl/gctx_gl.c
mbouron-gpsw: Commit updated.

libnodegl/rendertarget.h
mbouron-gpsw: Commit updated.

libnodegl/node_rtt.c
mbouron-gpsw: I removed the no_clear feature in a separate commit.

libnodegl/rendertarget_gl.h
mbouron-gpsw: Added a comment: // max color attachment + depth and stencil attachments

libnodegl/node_rtt.c
mbouron-gpsw: I'd like to keep it this way, if you prefer i can unroll it:
``` c
rt_params.colors[rt_params.nb_colors].clear_color[0] = s->clear_color[0];
rt_params.colors[rt_params.nb_colors].clear_color[1] = s->clear_color[1];
[...]
```

libnodegl/rendertarget_gl.h
cboesch-gpsw: max color attachment is pretty obvious (I mean, it's literally written in the declaration) but fine with me

libnodegl/rendertarget_gl.c
cboesch-gpsw: ok...

libnodegl/rendertarget_gl.c
mbouron-gpsw: I added a commit adding the glClearBuffer* function under the NGLI_FEATURE_CLEAR_BUFFER.

libnodegl/rendertarget.h
mbouron-gpsw: NGLI_LOAD_OP_LOAD and NGLI_STORE_OP_STORE are now set to 0.

libnodegl/rendertarget_gl.c
```diff
@@ -225,6 +273,18 @@ int ngli_rendertarget_gl_init(struct rendertarget *s, const struct rendertarget_
...
+    } else {
+        s_priv->invalidate = invalidate_noop;
+    }
+
+    if (gl->features & NGLI_FEATURE_CLEAR_BUFFER) {
+        s_priv->clear = clear_buffers;
+    } else {
+        s_priv->clear = clear_buffer;
+    }
+
```
cboesch-gpsw, line 287: do we really need to be overly verbose like this vs:

```c
s_priv->invalidate = (gl->features & NGLI_FEATURE_INVALIDATE_SUBDATA) ? invalidate : invalidate_noop;
s_priv->clear      = (gl->features & NGLI_FEATURE_CLEAR_BUFFER) ? clear_buffers : clear_buffer;
```

mbouron-gpsw: It is to keep the code consistent with the check below regarding the DRAW_BUFFERS feature.

cboesch-gpsw (approved):

libnodegl/gctx_gl.c
jmoguillansky-gpsw, line 328: another consideration is that compute (e.g. dispatch) is not supposed to be called inside a renderpass

mbouron-gpsw: Added the following code to pass.c:
``` c
     } else {
+        if (!ctx->start_render_pass) {
+            struct gctx *gctx = ctx->gctx;
+            ngli_gctx_end_render_pass(gctx);
+            ctx->current_rendertarget = ctx->available_rendertargets[1];
+            ctx->start_render_pass = 1;
+        }
+
         ngli_pipeline_dispatch(pipeline, params->nb_group_x, params->nb_group_y, params->nb_group_z);
     }
```

PR #103, opened 2020-10-12
media: prevent Media nodes from being shared

cboesch-gpsw:


libnodegl/node_media.c
```diff
@@ -213,6 +213,20 @@ static int media_init(struct ngl_node *node)
...
 }
 
+static int media_prepare(struct ngl_node *node)
+{
+    struct media_priv *s = node->priv_data;
+    if (s->nb_parents++) {
+        /*
+         * On Android, the frame can only be uploaded once, which discards
+         * them, so we can't have 2 Textures sharing the same Media.
+         */
```
mbouron-gpsw, line 223: I would say something like this insteaf: On Android, the frame can only be uploaded once and each subsequent upload will be a noop which results in an empty texture. This limitation prevents us from sharing the Media node.

This comment can also be put in the commit description.

libnodegl/node_media.c
```diff
@@ -213,6 +213,20 @@ static int media_init(struct ngl_node *node)
...
+        LOG(ERROR, "Media nodes can not be shared, share the Texture instead");
```
mbouron-gpsw, line 224: Media nodes can not be shared, the Texture should be shared instead.

mbouron-gpsw (approved):
Other than my remarks, LGTM.

libnodegl/node_media.c
cboesch-gpsw: Message adjusted

PR #112, opened 2020-10-15
block: do not switch from SSBO type back to UBO

cboesch-gpsw:


jmoguillansky-gpsw:
I think eventually we should expose the concepts of SSBO and UBO at the public Node.GL API level instead of trying to infer.

tests/compute.py
```diff
@@ -84,7 +84,7 @@ def compute_particules(cfg):
         ],
         layout='std430',
     )
-    opositions = ngl.Block(fields=[ngl.BufferVec3(count=nb_particules, label='positions')], layout='std430')
+    opositions = ngl.Block(fields=[ngl.BufferVec3(count=nb_particules, label='positions')], layout='std140')
```
jmoguillansky-gpsw, line 87: Why are we changing the layout to std140?

ubitux: It's explained in the description of the commit: we want to have the internals to switch to SSBO because it's writable, not because it's std430.

tests/compute.py
```diff
@@ -198,7 +199,7 @@ def compute_histogram(cfg, show_dbg_points=False):
     texture = ngl.Texture2D(width=size, height=size, data_src=texture_buffer)
     texture.set_format('r32g32b32a32_sfloat')
 
-    histogram_block = ngl.Block(layout='std430', label='histogram')
+    histogram_block = ngl.Block(layout='std140', label='histogram')
```
jmoguillansky-gpsw, line 202: maybe in the future we should hide glsl-specific syntax?

ubitux:
> I think eventually we should expose the concepts of SSBO and UBO at the public Node.GL API level instead of trying to infer.

I believe the exact opposite: this is implementation specific, so we choose what's best for the user.

tests/compute.py
ubitux: Yes I agree: I have this in my TODO list (dropping the `layout` option entirely), but it requires a few sensible changes internally.

tests/compute.py
jmoguillansky-gpsw: ok so this is just a way to test that the updated pgcraft logic works?  It's not a requirement in compute.py to use one layout over the other?

cboesch-gpsw: > ok so this is just a way to test that the updated pgcraft logic works?

The `pass.c` logic to be more specific, but yes.

> It's not a requirement in compute.py to use one layout over the other?

Yes: if we drop the commit fix, the compute test wouldn't work with `std140` (because the render would switch back from SSBO to UBO), while having it default to `std430` forces SSBO everywhere.

jmoguillansky-gpsw: I see

jmoguillansky-gpsw:
Seems that Metal API also doesn't expose concepts like "SSBO, UBO", it must have some internal logic to choose where to store the buffer, etc.  I can see the benefit in handling this logic automatically, in terms of simplifying the public API.

mbouron-gpsw:
Please reword the commit messages according to my offline remarks, otherwise LGTM.

mbouron-gpsw (approved):

PR #115, opened 2020-10-20
rtt: remove vflip parameter

mbouron-gpsw:
This parameter was useful for users relying on the OpenGL backend and
prepared the geometry and uv coordinates to deal with the rendertarget
output being vertically flipped.

Since df18098f01ff9795018b23dfc844225f325db527, we alter the
rendertarget output coordinate matrix for OpenGL offscreen rendering.

Moreover, this parameter won't work as expected with other graphics API.

### DIFF
--- a/libnodegl/node_rtt.c
+++ b/libnodegl/node_rtt.c
@@ -39,7 +39,6 @@ struct rtt_priv {
     int samples;
     float clear_color[4];
     int features;
-    int vflip;
 
     int use_rt_resume;
     int width;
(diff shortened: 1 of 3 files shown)


libnodegl/node_rtt.c
```diff
@@ -356,20 +353,18 @@ static int rtt_prefetch(struct ngl_node *node)
...
-    if (s->vflip) {
-        /* flip vertically the color and depth textures so the coordinates
-         * match how the uv coordinates system works */
-        for (int i = 0; i < s->nb_color_textures; i++) {
-            struct texture_priv *texture_priv = s->color_textures[i]->priv_data;
-            struct image *image = &texture_priv->image;
-            ngli_gctx_get_rendertarget_uvcoord_matrix(gctx, image->coordinates_matrix);
-        }
+    /* flip vertically the color and depth textures so the coordinates
+     * match how the uv coordinates system works */
```
ubitux, line 357: This "flip vertically" is not accurate anymore, isn't it? Maybe it could be changed to simply "transform" or something like this in another commit?

ubitux (approved):

libnodegl/node_rtt.c
mbouron-gpsw: Done.

cboesch-gpsw (approved):

PR #118, opened 2020-10-22
Misc small build stuff

cboesch-gpsw:


mbouron-gpsw (approved):

ngl-tools/common.c
```diff
@@ -19,6 +19,8 @@
  * under the License.
  */
 
+#define _POSIX_C_SOURCE 199309L // clock_gettime()
```
jmoguillansky-gpsw, line 22: In the future maybe we should put compile guards around this, so that it's only defined on specific platforms (same with unistd.h, etc)?

jmoguillansky-gpsw (approved):

ngl-tools/common.c
cboesch-gpsw: Why would you guard that define? It's mandatory if you want to use `clock_gettime()`.

jmoguillansky-gpsw: if we port to visual studio toolchain will there be porting issues?

cboesch-gpsw: Probably yes, but that wouldn't be related to the POSIX define, but the use of the function itself.

PR #120, opened 2020-10-23
rendertarget_gl: do not try to delete the default framebuffer

mbouron-gpsw:
While this won't be an issue with the OpenGL default framebuffer as the
framebuffer value is 0 (glDeleteFramebuffers() silently ignores 0), it
is an issue on the EAGL platform as the default framebuffer is a
framebuffer object managed by glcontext_eagl.m.

Forgotten in 0f799469b069a5d6b90724ae6cb96c02bdb1dc41.

### DIFF
--- a/libnodegl/rendertarget_gl.c
+++ b/libnodegl/rendertarget_gl.c
@@ -259,6 +259,8 @@ int ngli_rendertarget_gl_init(struct rendertarget *s, const struct rendertarget_
     s->width = params->width;
     s->height = params->height;
 
+    s_priv->wrapped = 0;
+
     int ret;
     if (require_resolve_fbo(s)) {
         ret = create_fbo(s, 1);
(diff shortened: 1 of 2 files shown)


libnodegl/rendertarget_gl.c
```diff
@@ -410,6 +413,7 @@ int ngli_default_rendertarget_gl_init(struct rendertarget *s, const struct rende
     s->width = params->width;
     s->height = params->height;
 
+    s_priv->wrapped = 1;
```
cboesch-gpsw, line 416: While it's not technically needed, I'd explicit `s_priv->wrapped = 0` in `ngli_default_rendertarget_gl_init()` so we can distinguish clearly the 2 codepaths:
- `ngli_rendertarget_gl_init()` owning the framebuffers
- `ngli_default_rendertarget_gl_init()` not owning the framebuffers

cboesch-gpsw (approved):
Should be fine aside from my comment.

It won't come as a surprise if I say this code is kind of confusing in many ways (probably due to the large set of intertwined codepath scenarios). I'm not comfortable enough with the code to say how it could be improved though; maybe some detailed comments in strategic places, but it's hard to tell.

libnodegl/rendertarget_gl.c
mbouron-gpsw: Done.

PR #121, opened 2020-10-23
texture: move access field to pipeline_texture_desc

mbouron-gpsw:


libnodegl/pipeline.h
```diff
@@ -31,6 +31,13 @@
 
 struct gctx;
 
+enum {
+    NGLI_ACCESS_READ_WRITE,
+    NGLI_ACCESS_READ_ONLY,
+    NGLI_ACCESS_WRITE_ONLY,
```
cboesch-gpsw, line 37: can you do `enum {undefined, ro, wo, rw}`, so that `rw=ro|wo` (`3=1|2`)?

libnodegl/texture.h
```diff
@@ -46,15 +46,6 @@ enum {
...
 
-enum {
-    NGLI_ACCESS_UNDEFINED,
-    NGLI_ACCESS_READ_BIT,
-    NGLI_ACCESS_WRITE_BIT,
-    NGLI_ACCESS_READ_WRITE,
-    NGLI_ACCESS_NB
-};
-
-NGLI_STATIC_ASSERT(texture_access, (NGLI_ACCESS_READ_BIT | NGLI_ACCESS_WRITE_BIT) == NGLI_ACCESS_READ_WRITE);
```
cboesch-gpsw, line 57: Ah yeah, and please keep the assert... :)

libnodegl/pipeline_gl.c
```diff
@@ -257,12 +268,10 @@ static void set_textures(struct pipeline *s, struct glcontext *gl)
 
         if (texture_binding->desc.type == NGLI_TYPE_IMAGE_2D) {
             GLuint texture_id = 0;
-            GLenum access = GL_READ_WRITE;
+            GLenum access = get_gl_access(texture_binding->desc.access);
```
cboesch-gpsw, line 271: `access` can be made `const` now

libnodegl/texture_gl.c
```diff
@@ -65,17 +65,6 @@ GLint ngli_texture_get_gl_wrap(int wrap)
...
-    [NGLI_ACCESS_READ_BIT]   = GL_READ_ONLY,
-    [NGLI_ACCESS_WRITE_BIT]  = GL_WRITE_ONLY,
-    [NGLI_ACCESS_READ_WRITE] = GL_READ_WRITE,
-};
-
-GLenum ngli_texture_get_gl_access(int access)
-{
-    return gl_access_map[access];
-}
-
```
cboesch-gpsw, line 78: This move could have been kept in a separate commit to reduce the diff.

libnodegl/pipeline_gl.c
mbouron-gpsw: Done.

libnodegl/pipeline.h
mbouron-gpsw: There is no need for an undefined value anymore. The pipeline must either set RW, RO or WO. Although the enum does not describe a bitset anymore as we just need a straight mapping between this value and the backend value. This is why this commit changes the API and perform the move at the same time.

mbouron-gpsw: Also I am not opposed to make the default (0) something else than NGLI_ACCESS_READ_WRITE, NGLI_ACCESS_READ_ONLY can be a good candidate.

cboesch-gpsw: Even if it doesn't make technical sense, it is semantically more correct to have `rw=ro|wo`, and it also will make a technical evolution to this model easier in the future; I believe that the `writable` flag today leading to `ro` XOR `wo` is not meant to stay anyway.

mbouron-gpsw: I decided to kept the original enum defintions.

libnodegl/texture.h
mbouron-gpsw: I kept the assert as well as the original definition of the enum.

libnodegl/texture_gl.c
mbouron-gpsw: Done.

cboesch-gpsw (approved):
LGTM

Issue #105, opened 2020-10-14
set memory qualifiers in pgcraft


jmoguillansky-gpsw:
Can we set memory qualifiers in inject_block (e.g. readonly, writeonly, etc)?

https://github.com/gopro/gopro-lib-node.gl/blob/4e1456401fe96fb12c058e30bcc0d1f398854ec9/libnodegl/pgcraft.c#L455

for data unit tests: if we don't set "readonly" for glsl buffer we get compile error:
Shader requires fragmentStoresAndAtomics but is not enabled on the device

cboesch-gpsw:
Sure, can you send a patch? There is already something similar done for the images. So you could add a `writable` field to `pgcraft_block` (like it's done in `pgcraft_texture`), set it appropriately in `pass.c`, and then adjust `pgcraft.c` to add the qualifier (maybe only when it's supported? not exactly sure of the support).

jmoguillansky-gpsw:
Here's more info:

some devices don't support writing to a buffer (SSBO) in vertex or fragment stage.  It's an optional feature.
See vkcontext.c: "vertexPipelineStoresAndAtomics" and "fragmentStoresAndAtomics".

In this case we have to specify "readonly" on the buffer in glsl.  

Here's the steps to reproduce the compile error.  

Step 1: in vkcontext.c: disable features "vertexPipelineStoresAndAtomics" and "fragmentStoresAndAtomics".  

Step 2: the following unit tests should reproduce: compute_particules, and various data unit tests.

The solution: we can put more precise memory qualifiers (e.g. mark buffer as readonly if we only read from it).

jmoguillansky-gpsw:
Here's a patch, but not sure if it's the best solution:

in pgcraft.c, in inject_block:

    const char* memory_qualifiers = (stage == NGLI_PROGRAM_SHADER_COMP) ? "" : "readonly";
    ngli_bstr_printf(b, " %s %s %s_block {\n", keyword, memory_qualifiers, named_block->name);

ubitux:
I explained in details what you can do to fix, it's not much changes. Please send a patch/make a PR.

jmoguillansky-gpsw:
Sure I'll make a PR

Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #124, opened 2020-10-27
pgcraft: specify buffer memory qualifiers

mbouron-gpsw:


libnodegl/pgcraft.c
```diff
@@ -405,6 +406,10 @@ static int inject_block(struct pgcraft *s, struct bstr *b,
         ngli_bstr_printf(b, "layout(%s)", layout);
     }
 
+    if (block->type == NGLI_TYPE_STORAGE_BUFFER) {
```
jmoguillansky-gpsw, line 409: can you confirm this doesn't break the GL backend?

mbouron-gpsw: I confirm it does break the GL/GLES backends, all the tests pass.

libnodegl/pgcraft.c
```diff
@@ -405,6 +405,10 @@ static int inject_block(struct pgcraft *s, struct bstr *b,
...
+        ngli_bstr_printf(b, " %s", named_block->writable ? "" : "readonly");
+    }
+
```
cboesch-gpsw, line 412: I'd prefer this form:

```c
if (block->type == NGLI_TYPE_STORAGE_BUFFER && !named_block->writable)
    ngli_bstr_print(b, " readonly");
```

- no extra { }
- printf → print
- no useless space added

libnodegl/pgcraft.h
```diff
@@ -65,6 +65,7 @@ struct pgcraft_block {
     const char *instance_name;
     int stage;
     int variadic;
+    int writable;
```
cboesch-gpsw, line 68: Can you add "Closes #105" somewhere in the commit?

libnodegl/pipeline_gl.c
```diff
@@ -310,10 +315,24 @@ static void set_buffers(struct pipeline *s, struct glcontext *gl)
...
+            LOG(ERROR, "context does not support uniform buffer objects");
+            return NGL_ERROR_UNSUPPORTED;
+        }
+
+        if (pipeline_buffer_desc->type == NGLI_TYPE_STORAGE_BUFFER &&
+            !(gl->features & NGLI_FEATURE_SHADER_STORAGE_BUFFER_OBJECT)) {
+            LOG(ERROR, "context does not support shader storage buffer objects");
+            return NGL_ERROR_UNSUPPORTED;
+        }
+
```
cboesch-gpsw, line 338: 1. How can this be triggered?
2. Should we really have these check at GL levels? don't they belong at the pipeline level for all backends instead?
3. Is this redundant with the check in node_block? (→ should we drop them from there?)

libnodegl/pgcraft.c
mbouron-gpsw: I also prefer this form, thanks. Branch updated.

libnodegl/pgcraft.h
mbouron-gpsw: Done.

libnodegl/pipeline_gl.c
mbouron-gpsw: 1. This can be triggered if someone directly use the pipeline API (with or without pgcraft).

2. We already have checks for other feature (uint uniforms, instanced arrays, compute shaders, draw instanced). So IMHO it makes sense to have such checks in the pipeline API (GL backend). I would prefer moving all the relevant check in a dedicated PR to the upper level (pipeline) only once we have reworked the features "API". We should also consider doing that for other structure (texture, rendertarget, ...) at the same time.

3. IMHO, it is not redundant with the check in `node_block.c` since we are allowed to use the pipeline API in other part of the code (hud, text, ...).

cboesch-gpsw: > 1. This can be triggered if someone directly use the pipeline API (with or without pgcraft).

OK

> 2. We already have checks for other feature (uint uniforms, instanced arrays, compute shaders, draw instanced). So IMHO it makes sense to have such checks in the pipeline API (GL backend). I would prefer moving all the relevant check in a dedicated PR to the upper level (pipeline) only once we have reworked the features "API". We should also consider doing that for other structure (texture, rendertarget, ...) at the same time.

OK

> 3. IMHO, it is not redundant with the check in `node_block.c` since we are allowed to use the pipeline API in other part of the code (hud, text, ...).

But can we use the block node without a pipeline?

mbouron-gpsw: I don't think so. The checks in node_block.c seems redundant after this commit. I will try to remove them in a separate commit.

cboesch-gpsw: OK so apparently if we remove the check in node block, we end up trying to compile the shader before it reaches the pipeline, which causes a hard fail on the backend side. We want to fail before that, and having the check in node block addresses this. I'm not sure how we can improve this redundant checks situation, but for now it's probably better to keep all of them.

cboesch-gpsw (approved):

PR #131, opened 2020-11-06
pgcraft: support OpenGL extensions

mbouron-gpsw:


libnodegl/glcontext.c
```diff
@@ -169,6 +169,33 @@ static int glcontext_probe_version(struct glcontext *glcontext)
     return 0;
 }
 
+static int glcontext_probe_glsl_version(struct glcontext *glcontext)
+{
+
```
cboesch-gpsw, line 174: extra linebreak

libnodegl/glcontext.h
```diff
@@ -47,6 +47,9 @@ struct glcontext {
     /* GL api */
     int version;
 
+    /* GLSL version */
+    int glsl_version;
+
```
cboesch-gpsw, line 52: I know this is to be consistent with the other comments, but it's really not helpful. (I'm not asking for a change)

libnodegl/pgcraft.c
```diff
@@ -489,15 +490,75 @@ const char *ublock_names[] = {
     [NGLI_PROGRAM_SHADER_COMP] = "comp",
 };
 
-static void set_glsl_header(struct pgcraft *s, struct bstr *b)
+static int has_ssbos(struct pgcraft *s, const struct pgcraft_params *params, int stage)
```
cboesch-gpsw, line 493: do you mind using something like `contain_X()` instead of `has_` to avoid the confusion with having the capability?

libnodegl/pgcraft.c
```diff
@@ -489,15 +490,75 @@ const char *ublock_names[] = {
...
+        if (pgcraft_block->stage == stage && block->type == NGLI_TYPE_STORAGE_BUFFER)
+            return 1;
+    }
+    return 0;
+}
+
+static int has_images(struct pgcraft *s, const struct pgcraft_params *params, int stage)
+{
+    struct darray *texture_infos_array = &s->texture_infos;
+    struct pgcraft_texture_info *texture_infos = ngli_darray_data(texture_infos_array);
```
cboesch-gpsw, line 507: can be made const

libnodegl/pgcraft.c
```diff
@@ -489,15 +490,75 @@ const char *ublock_names[] = {
...
+    const struct ngl_config *config = &gctx->config;
+
     ngli_bstr_printf(b, "#version %d%s\n", s->glsl_version, s->glsl_version_suffix);
 
-    if (ngli_darray_data(&s->texture_infos)) {
-        if (s->required_tex_exts)
-            for (int i = 0; s->required_tex_exts[i]; i++)
-                ngli_bstr_printf(b, "#extension %s : require\n", s->required_tex_exts[i]);
+    const int require_ssbo_extension = has_ssbos(s, params, stage);
+    const int require_image_extensions = has_images(s, params, stage);
```
cboesch-gpsw, line 528: can you use "feature" instead of "extension" here? We need the feature, but we may not need the extension; the decision to add the extension being if we're older than the minimum required glsl.

libnodegl/pgcraft.c
```diff
@@ -489,15 +490,75 @@ const char *ublock_names[] = {
...
+    };
+
+    for (int i = 0; i < NGLI_ARRAY_NB(extensions); i++) {
+        if (extensions[i].backend == config->backend &&
+            extensions[i].glsl_version > s->glsl_version &&
+            extensions[i].required)
+            ngli_bstr_printf(b, "#extension %s : require\n", extensions[i].extension);
+    }
+
+    if (ngli_darray_data(&s->texture_infos)) {
```
cboesch-gpsw, line 561: OK this was a mistake on my part but I think we need to use darray_count instead here (because we could imagine an implementation were we have a pre-allocated buffer even with 0 entry)

cboesch-gpsw (approved):
I think I only have cosmetics comments, your commits LGTM

libnodegl/pgcraft.c
mbouron-gpsw: Done.

libnodegl/glcontext.c
mbouron-gpsw: Fixed.

libnodegl/glcontext.h
mbouron-gpsw: Indeed, a cleanup is planned though.

libnodegl/pgcraft.c
mbouron-gpsw: From my point of view, we need the extension but it is or it isn't part of core language available.

libnodegl/pgcraft.c
mbouron-gpsw: I'll add a dedicated commit.

libnodegl/pgcraft.c
mbouron-gpsw: Ok.

libnodegl/pgcraft.c
mbouron-gpsw: Done.

libnodegl/pgcraft.c
mbouron-gpsw: Done.

libnodegl/pgcraft.c
mbouron-gpsw: Renamed _feature as well as renaming the extensions table to features, so it is more correct and consistent.

PR #132, opened 2020-11-09
Meson support for libnodegl and ngl-tools

cboesch-gpsw:


libnodegl/meson.build
```diff
@@ -0,0 +1,508 @@
...
+# We are also checking for IPHONE and not IOS but since the latter is actually
+# more specific (doesn't include tvOS and watchOS) we may want to consider
+# switching to it at some point.
+# See https://github.com/mesonbuild/meson/issues/7944 for more information
+iphone_check = '''#include <TargetConditionals.h>
+#if !TARGET_OS_IPHONE
+#error not iphone
+#endif'''
+if host_system == 'darwin' and cc.compiles(iphone_check, name: 'iPhone target')
+  host_system = 'iphone'
```
jmoguillansky-gpsw, line 52: can't the user pass in this information?

libnodegl/meson.build
```diff
@@ -0,0 +1,508 @@
...
+    lib_src += host_backend_cfg.get('src')
+    lib_deps += gbackend_deps
+    conf_data_dict += {gbackend_cfg.get('cfg'): true}
+    conf_data_dict += {host_backend_cfg.get('cfg') : true}
+  endif
+endforeach
+
+
+#
+# VAAPI dependency
```
jmoguillansky-gpsw, line 300: Shouldn't this be inside a "if target == linux" clause?

libnodegl/meson.build
```diff
@@ -0,0 +1,508 @@
...
+  'updatedoc',
+  command: [cp, doc_file, meson.current_source_dir() / 'doc'],
+)
+
+
+#
+# OpenGL specifications
+#
+
+gl_xml = custom_target(
```
jmoguillansky-gpsw, line 434: should this be inside "if BACKEND_GL" clause (only enable if gl backend enabled)?

cboesch-gpsw: It doesn't matter: the target has to be explicitly called by the developer to be triggered. And as a developer, it could make sense to update the binding even if you're not on a system that supports it.

libnodegl/meson.build
cboesch-gpsw: What if BSD users want that as well? We could have various other target that use vaapi so I'd rather just detect if it's available on the system, it doesn't cost much.

libnodegl/meson.build
cboesch-gpsw: It doesn't look like a user can alter the host-system from within a cross-file directly

mbouron-gpsw:
vaapi support is missing from build.

cboesch-gpsw:
> vaapi support is missing from build.

Oups. Should be fixed now, thanks.

mbouron-gpsw:
> > vaapi support is missing from build.
> 
> Oups. Should be fixed now, thanks.

I confirm it is fixed, thanks.

mbouron-gpsw:
The build on Windows (MingW64) does not detect the dependencies of the opengl32, gdi32. This results in a build that does not have any backend.

cboesch-gpsw:
> The build on Windows (MingW64) does not detect the following dependencies: opengl32, gdi32. This results in a build that does not have any backend available.

Ah yeah I treated them as dependencies (causing detection through `pkg-config`) instead of libs (classic `-l...`); should be fixed now:

```diff
diff --git a/libnodegl/meson.build b/libnodegl/meson.build
index 9d52c97a..4b861fae 100644
--- a/libnodegl/meson.build
+++ b/libnodegl/meson.build
@@ -298,7 +298,7 @@ gbackends_cfg = {
       'windows': {
         'src': files('glcontext_wgl.c'),
         'cfg': 'HAVE_GLPLATFORM_WGL',
-        'deps': ['opengl32', 'gdi32'],
+        'libs': ['opengl32', 'gdi32'],
       },
     },
   },
```

mbouron-gpsw:
> > The build on Windows (MingW64) does not detect the following dependencies: opengl32, gdi32. This results in a build that does not have any backend available.
> 
> Ah yeah I treated them as dependencies (causing detection through `pkg-config`) instead of libs (classic `-l...`); should be fixed now:
> 
> ```diff
> diff --git a/libnodegl/meson.build b/libnodegl/meson.build
> index 9d52c97a..4b861fae 100644
> --- a/libnodegl/meson.build
> +++ b/libnodegl/meson.build
> @@ -298,7 +298,7 @@ gbackends_cfg = {
>        'windows': {
>          'src': files('glcontext_wgl.c'),
>          'cfg': 'HAVE_GLPLATFORM_WGL',
> -        'deps': ['opengl32', 'gdi32'],
> +        'libs': ['opengl32', 'gdi32'],
>        },
>      },
>    },
> ```

I confirm it fixes the issue, thanks.

mbouron-gpsw:
I've tested on Linux, Android, MinGW and macOS without any issue.

mbouron-gpsw:
I also did a test on iOS, no issue so far. It looks like everything is working. LGTM.

mbouron-gpsw (approved):

PR #134, opened 2020-11-10
Misc commits from the ngfx integration branch

mbouron-gpsw:


libnodegl/darray.c
```diff
@@ -103,8 +103,7 @@ void *ngli_darray_pop(struct darray *darray)
 
 void *ngli_darray_get(const struct darray *darray, int index)
 {
-    if (index < 0 || index >= darray->count)
-        return NULL;
+    ngli_assert(index >= 0 && index < darray->count);
```
cboesch-gpsw, line 106: If we do that, we need to:
1. in pipeline, remove the redundant asserts after each `ngli_darray_get()`
2. in deserialize, add the check to avoid the assert which could be triggered with a crafted input scene

jmoguillansky-gpsw: In darray we only do an assertion on the index, not if item is NULL.  
In pipeline, etc we assert that item is not NULL.  Should I remove those?

jmoguillansky-gpsw: "in deserialize, add the check to avoid the assert which could be triggered with a crafted input scene"
> Can you clarify?

mbouron-gpsw: The commit has been updated.
In deserialize.c we could end up parsing a malformed or malicious input where the id passed to get_abs_node() is >= to ngli_darray_count(nodes_array), in the previous form of the patch that would have ended up in a assert (which we want to avoid).
Regarding the check in pipeline_gl.c, they were done on the address of the item (meaning we can remove thoses).

libnodegl/deserialize.c
```diff
@@ -223,7 +223,10 @@ static int parse_kvs(const char *s, int *nb_kvsp, char ***keysp, int **valsp)
 
 static struct ngl_node **get_abs_node(struct darray *nodes_array, int id)
 {
-    return ngli_darray_get(nodes_array, ngli_darray_count(nodes_array) - id - 1);
+    int index = ngli_darray_count(nodes_array) - id - 1;
```
ubitux, line 226: nit: `const int index = ...`

jmoguillansky-gpsw: updated

cboesch-gpsw (approved):

PR #142, opened 2020-11-13
Debug: Enable Programmatic GPU Capture

jmoguillansky-gpsw:
Enable auto generation of GPU capture traces, via renderdoc when using GL / Vulkan / DirectX 12 , and via Metal APIs when using Metal.  
Usage:
set environment variable "DEBUG_CAPTURE=on" when running unit test
May need to also set LD_PRELOAD=librenderdoc.so.  This allows renderdoc to intercept API calls.
Captures are written to /tmp/Renderdoc
For Metal GPU capture: may need to also enable Metal capture in XCode->Edit Scheme-> GPU Frame Capture...

Benefits:
1) Simplifies process of capturing GPU traces
2) Supports capture of GPU traces with headless rendering

Troubleshooting:
Building renderdoc from source and debugging through the renderdoc source code is very helpful for troubleshooting.
Renderdoc also generates intermediate log files (see /tmp/Renderdoc), however the log files may be deleted when the program exists.  So, may need to set breakpoints to review the log file.  

TODO:
1) renderdoc is currently not packaged via a package manager.  
Need to package renderdoc or build from source automatically.
2) install build artifacts (e.g. renderdoc library) to nodegl-env/lib automatically


libnodegl/renderdoc_capture.c
```diff
@@ -0,0 +1,40 @@
...
+
+void init_capture() {
+#ifdef _WIN32
+    // At init, on windows
+    HMODULE mod = LoadLibraryA("renderdoc.dll");
+    assert(mod);
+    pRENDERDOC_GetAPI RENDERDOC_GetAPI =
+        (pRENDERDOC_GetAPI)GetProcAddress(mod, "RENDERDOC_GetAPI");
+    int ret = RENDERDOC_GetAPI(eRENDERDOC_API_Version_1_4_1, (void **)&rdoc_api);
+    assert(ret == 1);
```
cboesch-gpsw, line 23: we have our own assert, but i think you want to error out gracefully here

libnodegl/renderdoc_capture.c
```diff
@@ -0,0 +1,40 @@
...
+#else
+    void *mod = dlopen("librenderdoc.so", RTLD_LAZY); //RTLD_NOW | RTLD_NOLOAD);
+    if (!mod) fprintf(stderr, "could not load librenderdoc.so: %s", dlerror());
```
cboesch-gpsw, line 26: we have a `LOG()` helper

libnodegl/renderdoc_capture.c
```diff
@@ -0,0 +1,40 @@
...
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <dlfcn.h>
+#endif
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+static RENDERDOC_API_1_4_1 *rdoc_api = NULL;
+void* _capture_device = NULL; //TODO: remove
```
cboesch-gpsw, line 13: why not now?

libnodegl/meson.build
```diff
@@ -389,6 +390,41 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
...
+        message ('renderdoc include: @0@'.format(renderdoc_include_dirs))
+    endif
+    
+    renderdoc = declare_dependency(
+        include_directories: renderdoc_include_dirs,
+        link_args: [ '-L' + renderdoc_library_dirs, '-ldl' ],
+    )
+    lib_src += 'renderdoc_capture.c'
+    lib_deps += renderdoc
+    lib_c_args += '-DENABLE_CAPTURE'
```
cboesch-gpsw, line 423: we have a configuration mechanism for that sort of things

libnodegl/meson.build
```diff
@@ -389,6 +390,41 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
...
+#
+# GPU capture dependencies
+#
+
+if host_system == 'windows' or host_system == 'linux'
+    enable_renderdoc_capture = true
+elif host_system == 'darwin'
+    enable_metal_capture = true
+endif
+
```
cboesch-gpsw, line 402: mmh you want to use `meson_options.txt` here, and have it autodetected somehow; you also probably want to do that the same way we do it with debug mem/scene/etc.

libnodegl/meson.build
```diff
@@ -389,6 +390,41 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
...
+        #TODO: fix
+        renderdoc_dir= '../renderdoc_1.10'
+        renderdoc_include_dirs = include_directories(renderdoc_dir + '/renderdoc/api/app')
+        renderdoc_library_dirs = renderdoc_dir+'/cmake-build-debug/lib'
+        message ('renderdoc include: @0@'.format(renderdoc_include_dirs))
+    endif
+    
+    renderdoc = declare_dependency(
+        include_directories: renderdoc_include_dirs,
+        link_args: [ '-L' + renderdoc_library_dirs, '-ldl' ],
```
cboesch-gpsw, line 419: that looks wrong, you should never need to specify `-L` or `-l` with meson

libnodegl/meson.build
```diff
@@ -389,6 +390,41 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
...
+    )
+    lib_src += 'renderdoc_capture.c'
```
cboesch-gpsw, line 421: you want to use `files()` here

libnodegl/meson.build
```diff
@@ -389,6 +390,41 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
...
+    lib_deps += renderdoc
+    lib_c_args += '-DENABLE_CAPTURE'
+elif enable_metal_capture
+    lib_src += 'mtl_capture.mm'
+    lib_c_args += '-DENABLE_CAPTURE'
+endif
```
cboesch-gpsw, line 427: we will probably want the support for this on master before vulkan, so you probably want to split out that code and submit a PR for master

libnodegl/meson.build
```diff
@@ -389,6 +390,41 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
...
+
+if enable_renderdoc_capture
+    message('Enabling RenderDoc Capture')
+    if host_system == 'windows'
+        #set (RENDERDOC_DIR C:/Program\ Files/RenderDoc)
+        #set (RENDERDOC_INCLUDE_DIRS ${RENDERDOC_DIR})
+        #set (RENDERDOC_LIBRARY_DIRS ${RENDERDOC_DIR})
+    elif host_system == 'linux'
+        #TODO: fix
+        renderdoc_dir= '../renderdoc_1.10'
```
cboesch-gpsw, line 411: that doesn't look good, especially since it's a backward reference

libnodegl/capture.h
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2018 GoPro Inc.
```
cboesch-gpsw, line 2: 2020

libnodegl/capture.h
```diff
@@ -0,0 +1,33 @@
...
+extern "C" {
+#endif
+
+void init_capture(void);
+void begin_capture(void);
+void end_capture(void);
+
+#ifdef __cplusplus
+}
+#endif
```
cboesch-gpsw, line 33: please use the pattern you see in other files

```c
#ifndef CAPTURE_H
#defined CAPTURE_H

void init_capture(void);
void begin_capture(void);
void end_capture(void);

#endif
```

libnodegl/gctx_gl.c
```diff
@@ -38,6 +38,10 @@
 #include "program_gl.h"
 #include "rendertarget_gl.h"
 #include "texture_gl.h"
+#ifdef ENABLE_CAPTURE
+#include "capture.h"
+static int DEBUG_CAPTURE;
```
cboesch-gpsw, line 43: no mutable globals please, put that somewhere into the context

libnodegl/renderdoc_capture.c
jmoguillansky-gpsw: updated

libnodegl/renderdoc_capture.c
jmoguillansky-gpsw: updated

libnodegl/renderdoc_capture.c
jmoguillansky-gpsw: updated (removed)

libnodegl/capture.h
jmoguillansky-gpsw: updated

libnodegl/capture.h
jmoguillansky-gpsw: updated

libnodegl/gctx_gl.c
jmoguillansky-gpsw: updated

libnodegl/renderdoc_capture.c
ubitux: Asserts that can be triggered are never ok; multiple reasons for that:
- even developers want clear errors
- asserts are not a cheap error checking
- asserts hint about impossible codepaths: someone (or a compiler) may assume by reading the code that `RENDERDOC_GetAPI()` could actually never fail and only be `1` all the time (even if it might be true today, it may not tomorrow)

Anyway, just add a `if`, that's how it's done everywhere.

jmoguillansky-gpsw: FYI: the renderdoc documentation also uses assert in their sample code: https://renderdoc.org/docs/in_application_api.html. 

A few comments:
1) this code path is only executed when user enables GPU capture (disabled by default), so you don't have to worry about the main code path being less stable
2) assert: a certain value is expected (e.g.  ("this condition should never happen")) and if not true, then an error is triggered and execution stops.  If statement: either path is valid.  
3) "developers want clear errors": we can modify the ngli_assert macro to dump a log message if assertion fails if you like, but it should be clear to the developer what the error is because they can look at the line of code.  
4) "asserts are not cheap error checking": only enabled in debug build.  
5) "RENDERDOC_GetAPI() could actually never fail and only be 1 all the time": yes this is the expected behavior.  If our code is implemented correctly it should never fail.

jmoguillansky-gpsw:
moved to #196

libnodegl/renderdoc_capture.c
cboesch-gpsw: > FYI: the renderdoc documentation also uses assert in their sample code: https://renderdoc.org/docs/in_application_api.html.
> 
> A few comments:
> 
> 1. this code path is only executed when user enables GPU capture (disabled by default), so you don't have to worry about the main code path being less stable

Just because it's a developer code path doesn't mean it should have an different and violent error management

> 2. assert: a certain value is expected (e.g.  ("this condition should always be true")) and if not true, then an error is triggered and execution stops.  If statement: either path is valid.

Except you don't know if it's always going to succeed, especially since it's an external library you don't control.

> 3. "developers want clear errors": we can modify the ngli_assert macro to dump a log message if assertion fails if you like, but it should be clear to the developer what the error is because they can look at the line of code.

Or you could just use the same mechanism we everywhere: a simple `if`/`return`

> 4. "asserts are not cheap error checking": only enabled in debug build.

By "cheap" I meant "I don't want to bother adding a `if` and forwarding my code return so let's just add an assert".

> 5. "RENDERDOC_GetAPI() could actually never fail and only be 1 all the time": yes this is the expected behavior.  If our code is implemented correctly  (and more importantly, if the environment is setup correctly) it should never fail.

This is not true. `RENDERDOC_GetAPI` can currently fail if the parameters are invalid. You also don't know if a future version is going to do an allocation or initialization of some sort that can fail.

This is an init function, just add proper error checking. You'll need it for other errors that happen in that function anyway.

libnodegl/renderdoc_capture.c
cboesch-gpsw: It appears you didn't. Also please fix the many style issues you have. This one is typically missing a line break.

jmoguillansky-gpsw: this PR is outdated.  Moved to #196 .  I will close this one

libnodegl/renderdoc_capture.c
jmoguillansky-gpsw: I will update (in PR #196)

PR #144, opened 2020-11-16
Temp dir cleanup

cboesch-gpsw:


mbouron-gpsw:
A separate commit should made get_temp_dir (cmp.py) private since it's only used locally. Also, shouldn't get_temp_dir (cmp.py) rely on get_nodegl_tempdir() ?

ubitux:
> A separate commit should made get_temp_dir (cmp.py) private since it's only used locally. Also, shouldn't get_temp_dir (cmp.py) rely on get_nodegl_tempdir() ?

`get_temp_dir()` is removed/inlined using `get_nodegl_tempdir()` in "utils/tests: use get_nodegl_tempdir()". Do you want me to split that commit?

mbouron-gpsw:
> > A separate commit should made get_temp_dir (cmp.py) private since it's only used locally. Also, shouldn't get_temp_dir (cmp.py) rely on get_nodegl_tempdir() ?
> 
> `get_temp_dir()` is removed/inlined using `get_nodegl_tempdir()` in "utils/tests: use get_nodegl_tempdir()". Do you want me to split that commit?

No, that's OK like this. LGTM then.

mbouron-gpsw (approved):

pynodegl-utils/pynodegl_utils/ui/graph_view.py
```diff
@@ -189,7 +189,7 @@ def _init_ctx(self, rendering_backend):
         )
 
     def _update_graph(self, dot_scene):
-        basename = op.join(tempfile.gettempdir(), 'ngl_scene.')
+        basename = op.join(misc.get_nodegl_tempdir(), 'ngl_scene.')
```
mbouron-gpsw, line 192: You can remove the tempfile import.

cboesch-gpsw: Done

PR #151, opened 2020-11-25
Compute workgroups node parameters

cboesch-gpsw:


jmoguillansky-gpsw (commented):
Does compute node have workgroup count and compute program has workgroup size?

cboesch-gpsw:
> Does compute node have workgroup count and compute program has workgroup size?

yes

jmoguillansky-gpsw:
In Metal, when we call dispatch, we pass threadsPerGrid (workgroup count) and threadsPerThreadGroup (workgroup size).  
If we want to align with Metal, perhaps both params should be part of Compute node.  
Also personally I prefer "threadsPerThreadGroup" to "workgroup size": I think it's easier to understand, but either is okay.

cboesch-gpsw:
> In Metal, when we call dispatch, we pass threadsPerGrid (workgroup count) and threadsPerThreadGroup (workgroup size).
> If we want to align with Metal, perhaps both params should be part of Compute node.
> Also personally I prefer "threadsPerThreadGroup" to "workgroup size": I think it's easier to understand, but either is okay.

How is it in D3D?

jmoguillansky-gpsw:
Also, we should be able to call the same compute program with different params for work group size, so in that regard it might be better to put work group size in compute node

jmoguillansky-gpsw:
In DirectX 12 (and Vulkan) you only pass the number of workgroups in dispatch call.  The workgroup size is defined in the shader.

cboesch-gpsw:
> In DirectX 12 (and Vulkan) you only pass the number of workgroups in dispatch call. The workgroup size is defined in the shader.

And what are the name used to refer to the size/count, typically in the limit/capabilities?

jmoguillansky-gpsw:
In DX12: "numthreads" = workgroup size.  https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-attributes-numthreads.  Max threads is defined in the spec.  
In Metal: threadsPerThreadGroup / maxTotalThreadsPerThreadGroup

jmoguillansky-gpsw:
It looks good to me

cboesch-gpsw:
> In DX12: "numthreads" = workgroup size. https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-attributes-numthreads. Max threads is defined in the spec.
> In Metal: threadsPerThreadGroup / maxTotalThreadsPerThreadGroup

OK so, we have 2 things:
- the name: we can change that to `workgroup_count` and `workground_size` to `thread_count` and `thread_size` to match Metal+D3D vs Khronos; usually we go along the Khronos naming for consistency, but we can maybe make an exception (@mbouron-gpsw an opinion?)
- the positioning: I still believe we want the group size at the shader level for the reasons mentioned in the commit description (fragility implied by duplication and having it away from the shader); note that both size and count are getting unified internally very quickly: typically the `pass` merge them.

mbouron-gpsw:
> > In DX12: "numthreads" = workgroup size. https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-attributes-numthreads. Max threads is defined in the spec.
> > In Metal: threadsPerThreadGroup / maxTotalThreadsPerThreadGroup
> 
> OK so, we have 2 things:
> 
>     * the name: we can change that from `workgroup_count`/`workground_size` to `thread_count`/`thread_size` to match Metal+D3D vs Khronos; usually we go along the Khronos naming for consistency, but we can maybe make an exception (@mbouron-gpsw an opinion?)

I would keep the Khronos naming for consistency unless we also rename NGL_CAP_MAX_COMPUTE_* (and probably our glimits struct as well).

> 
>     * the positioning: I still believe we want the group size at the shader level for the reasons mentioned in the commit description (fragility implied by duplication and having it away from the shader); note that both size and count are getting unified internally very quickly: typically the `pass` merge them.

No strong opinion here, what is the most convenient for the users. I haven't used extensively compute shaders yet and I can be wrong but it looks like some compute shaders are written against specific workgroup sizes and won't necessarily adapt to "random" workgroup sizes.

jmoguillansky-gpsw:
"* the positioning: I still believe we want the group size at the shader level for the reasons mentioned in the commit description (fragility implied by duplication and having it away from the shader); note that both size and count are getting unified internally very quickly: typically the `pass` merge them.
"
>> We can't unify the size and count, there's a limit on max workgroup size (numThreadsPerWorkGroup), and it can be device specific, at least on Vulkan backend.  The two parameters are different.  Workgroup count defines how many workgroups, and workgroup size defines how many threads per workgroup.  

>> I think the high level API should specify both workgroup size and count.  The 3 backends can have different internal logic. 
On Vulkan / Dx12 we can specify the workgroup size in the generated shader, and the workgroup count in the Vulkan / Dx12 API call.  It might be possible to pass the workgroup size using "specialization constants".  
On Metal backend we can specify the workgroup size and count in the API call, if I understand correctly.

cboesch-gpsw:
> > * the positioning: I still believe we want the group size at the shader level for the reasons mentioned in the commit description (fragility implied by duplication and having it away from the shader); note that both size and count are getting unified internally very quickly: typically the `pass` merge them.
> 
> No strong opinion here, what is the most convenient for the users. I haven't used extensively compute shaders yet and I can be wrong but it looks like some compute shaders are written against specific workgroup sizes and won't necessarily adapt to "random" workgroup sizes.

Yeah AFAICT that's the case, because there usually is inter communication between "threads" in the workgroup. And at the above level, workgroups are independent, so their count can be variable. This matches the current vision: size at `ComputeProgram` level, count at `Compute` level.

Note that the current proposal doesn't change the existing state: workgroup count is already at `Compute` level, and the workgroup size is in the shader string of the `ComputeProgram`: this PR doesn't change this decision.

mbouron-gpsw (approved):

PR #152, opened 2020-11-25
Port to Windows / Visual Studio Toolchain on Master Branch

jmoguillansky-gpsw:
Port node.gl to Windows using Visual Studio Toolchain.  

Status:
GL backend working on windows.

Ok:                 210
Expected Fail:      0
Fail:               1

This PR can be broken down into the following features:
build: add windows / msvc support
nodegl: port code to windows
ngl-tools: port code to windows


external/scripts/sync.sh
```diff
@@ -0,0 +1,23 @@
+OS=$1
```
cboesch-gpsw, line 1: - script is missing a shebang
- if posix shell, you want `set -xeu`
- this script is getting changed several times in the history, you need to squash the commits: I had to review that file several times for nothing

external/scripts/sync.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "OS: [$OS]"
+TMP_DIR=/tmp/nodegl/external
+
+if [ $OS == "win64" ]; then
+    mkdir -p win64
+
+    mkdir -p $TMP_DIR/pkg-config && wget -nc https://github.com/jmoguillansky-gpsw/gopro-nodegl-thirdparty-binaries/blob/master/external/win64/pkg-config.tgz?raw=true -O $TMP_DIR/pkg-config/pkg-config.tgz -P $TMP_DIR/pkg-config
+    cd win64 && tar xzf $TMP_DIR/pkg-config/pkg-config.tgz; cd -
```
cboesch-gpsw, line 10: please use a sub-shell instead of `cd ; cd -`

external/scripts/sync.sh
```diff
@@ -1,18 +1,19 @@
...
 
 if [ $OS == "win64" ]; then
     mkdir -p win64
 
     mkdir -p $TMP_DIR/pkg-config && wget -nc https://github.com/jmoguillansky-gpsw/gopro-nodegl-thirdparty-binaries/blob/master/external/win64/pkg-config.tgz?raw=true -O $TMP_DIR/pkg-config/pkg-config.tgz -P $TMP_DIR/pkg-config
     cd win64 && tar xzf $TMP_DIR/pkg-config/pkg-config.tgz; cd -
 
-    mkdir -p $TMP_DIR/ffmpeg && wget -nc https://github.com/jmoguillansky-gpsw/gopro-nodegl-thirdparty-binaries/blob/master/external/win64/ffmpeg_x64-windows.tgz?raw=true -O $TMP_DIR/ffmpeg/ffmpeg_x64-windows.tgz -P $TMP_DIR/ffmpeg
-    cd win64 && tar xzf $TMP_DIR/ffmpeg/ffmpeg_x64-windows.tgz; cd -
+    mkdir -p $TMP_DIR/ffmpeg && wget --header "Authorization: token $TOKEN" -nc https://api.github.com/repos/jmoguillansky-gpsw/gopro-pkg/actions/artifacts/28753508/zip -O $TMP_DIR/ffmpeg/ffmpeg_x64-windows.zip -P $TMP_DIR/ffmpeg
```
cboesch-gpsw, line 12: how long will these artifacts survive?

Makefile
```diff
@@ -154,6 +154,8 @@ ifneq ($(TARGET_OS),MinGW-w64)
 else
 	cp -r $< $@
 endif
+#TODO: submit PR for sxplayer changes
```
cboesch-gpsw, line 224: can you do that instead? the current branch can't be reviewed until the associated PR is merged in sxplayer

Makefile
```diff
@@ -29,10 +27,22 @@ PYTHON_MAJOR = 3
...
 
+ifeq ($(TARGET_OS),Windows)
+PYTHON     ?= python.exe
+PREFIX     ?= nodegl-env
+W_PWD = $(shell wslpath -w .)
+W_PREFIX ?= $(W_PWD)\$(PREFIX)
+$(info PYTHON: $(PYTHON))
+$(info PREFIX: $(PREFIX))
+$(info W_PREFIX: $(W_PREFIX))
+else
```
cboesch-gpsw, line 41: So you are able to run this Makefile? why the external sync script then? you can download from here

Makefile
```diff
@@ -45,14 +55,28 @@ endif
 
 SXPLAYER_VERSION ?= 9.6.0
 
+ifeq ($(TARGET_OS), Windows)
+#TODO: identify correct path
```
cboesch-gpsw, line 59: what isn't correct?

libnodegl/meson.build
```diff
@@ -314,7 +330,7 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
     gbackend_deps += dependency(dep, required: opt_gbackend)
   endforeach
   foreach dep : host_backend_cfg.get('libs', [])
-    gbackend_deps += cc.find_library(dep, required: opt_gbackend)
+    gbackend_deps += cc.find_library(dep, required: opt_gbackend, dirs: cc_lib_dirs)
```
cboesch-gpsw, line 333: I know this is a hack we have in the vulkan branch, but it is incorrect: if we do that, it replaces the search dir, which means the libraries won't be searched on the default paths anymore, so it will likely fail typically in an independent build system.

libnodegl/meson.build
```diff
@@ -305,6 +308,19 @@ if host_system in ['darwin', 'iphone']
...
+cc_lib_dirs = [ get_option('prefix') / get_option('libdir') ]
+
+if host_system == 'windows' and cc.get_id() == 'msvc'
+
+  #TODO: set correct path
+  windows_sdk_dir = 'C:\\Program Files (x86)\\Windows Kits\\10'
+  windows_sdk_library_directories = [ windows_sdk_dir + '\\lib\\10.0.18362.0\\um\\x64' ]
+
+  cc_lib_dirs += windows_sdk_library_directories
+  lib_include_dirs += include_directories('..\\external\\win64\\gl\\include')
```
cboesch-gpsw, line 320: `..` is incorrect, the builddir could be anywhere: the root makefile is a user of an independently working build system, the dependency doesn't go the other way around.

ngl-tools/meson.build
```diff
@@ -42,9 +42,13 @@ add_project_arguments(
 # Main dependencies
 #
 tool_deps = [
-  cc.find_library('m', required: false),
   dependency('libnodegl'),
 ]
+
+if host_system != 'windows'
+  tool_deps += cc.find_library('m', required: false)
+endif
```
cboesch-gpsw, line 50: why do you need this?

ngl-tools/meson.build
```diff
@@ -115,6 +119,12 @@ tools_specs = {
   },
 }
 
+if host_system == 'windows'
+  tool_install_dir = 'Scripts'
```
cboesch-gpsw, line 123: "Scripts"? Those are binary programs

libnodegl/node_buffer.c
```diff
@@ -127,15 +126,16 @@ static int buffer_init_from_filename(struct ngl_node *node)
...
+    if (!s->fp) {
         LOG(ERROR, "could not open '%s'", s->filename);
         return NGL_ERROR_IO;
     }
 
-    off_t filesize = lseek(s->fd, 0, SEEK_END);
-    off_t ret      = lseek(s->fd, 0, SEEK_SET);
-    if (filesize < 0 || ret < 0) {
+    fseek(s->fp, 0, SEEK_END);
+    int filesize = ftell(s->fp);
```
cboesch-gpsw, line 136: this breaks with large files

libnodegl/utils.c
```diff
@@ -47,20 +51,21 @@ char *ngli_strdup(const char *s)
...
 
-int64_t ngli_gettime(void)
-{
-    struct timeval tv;
-
-    gettimeofday(&tv, NULL);
-    return 1000000 * (int64_t)tv.tv_sec + tv.tv_usec;
-}
-
 int64_t ngli_gettime_relative(void)
```
cboesch-gpsw, line 54: This belongs in a dedicated commit, prototype removal is forgotten btw.

libnodegl/backends/gl/glincludes.h
```diff
@@ -57,17 +57,14 @@
...
+# include <gl/GL.h>
 # include <GL/glcorearb.h>
 # include <GL/glext.h>
 # define NGL_OGL3_COMPAT_INCLUDES 1
 #endif
 
-#ifdef _WIN32
-#define NGLI_GL_APIENTRY WINAPI
-#else
 #define NGLI_GL_APIENTRY
```
cboesch-gpsw, line 67: mmh what's going on here? the commit description needs to explain that

libnodegl/api.c
```diff
@@ -70,7 +70,7 @@ static int get_default_platform(void)
     return NGL_PLATFORM_MACOS;
 #elif defined(TARGET_ANDROID)
     return NGL_PLATFORM_ANDROID;
-#elif defined(TARGET_MINGW_W64)
+#elif defined(TARGET_WINDOWS)
```
cboesch-gpsw, line 73: you can't do that in an isolated commit: it needs to be associated with changes in the `meson.build`, in the CI and in the documentation

ngl-tools/ipc.c
```diff
@@ -22,10 +22,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 #ifdef _WIN32
 #include <winsock2.h>
+#define ssize_t int
```
cboesch-gpsw, line 27: why do you define `ssize_t` if you remove it from the code? also, this is defined in mingw, but not msvc apparently

pynodegl/setup.py
```diff
@@ -36,13 +36,18 @@ def __init__(self, pkg_config_bin='pkg-config'):
...
+        cflags = pkgcfg_cflags.split()
+        ldflags = pkgcfg_libs.split()
+        self.include_dirs = [f[2:] for f in cflags if f.startswith('-I')]
+        self.library_dirs = [f[2:] for f in ldflags if f.startswith('-L')]
+        self.libraries = []
+        for f in ldflags:
+            if f.startswith('-l'):
+                self.libraries.append(f[2:])
+            elif f.endswith('.lib'):
+                self.libraries.append(f[:-4])
```
cboesch-gpsw, line 50: This change doesn't belong in `.gitignore` commit

cboesch-gpsw (changes requested):

external/scripts/sync.sh
jmoguillansky-gpsw: Thanks I will update...
Regarding the git history I kept two versions of the file but once we sort out the strategy I'll squash the commits.

external/scripts/sync.sh
jmoguillansky-gpsw: Thanks, good tip.  I'll update...

external/scripts/sync.sh
jmoguillansky-gpsw: 3 months, but there's several issues with github action artifacts: 1) require security token, 2) different url for each artifact for each run, thst we don't control.  I think if we decide to host prebuilt binaries we should upload them to an ftp server (via github action)

Makefile
jmoguillansky-gpsw: Lets discuss the strategy first regarding 3rdparty libs.  If everyone agrees, I'll submit PR

Makefile
jmoguillansky-gpsw: So we can ignore the details / complexity of the main makefile and develop this code in isolation.  We can put this in a gnu makefile format but I think a bash script is simpler?

libnodegl/utils.c
jmoguillansky-gpsw: Thanks I'll update...

libnodegl/backends/gl/glincludes.h
jmoguillansky-gpsw: Lots of compile errors if I keep "winapi"... with this change the gl backend works.  Is "winapi" related to symbol visibility and dlls?

ngl-tools/ipc.c
jmoguillansky-gpsw: It was either define it here or put ifdefs everywhere that recv / send is called and use int on windows and ssize _t on unix.  This was basically a trick to reduce the code changes.  Alternatively we could use macros or abstraction APIs

pynodegl/setup.py
jmoguillansky-gpsw: I'll clean up the history...

Makefile
jmoguillansky-gpsw: The sdk path was hardcoded (C:/...)

libnodegl/meson.build
jmoguillansky-gpsw: For some reason msvc finds the library but meson doesn't.  I had to add the search paths explicitly.  Alternatively we could just pass the lib names directly as params

libnodegl/meson.build
jmoguillansky-gpsw: Meson converts to absolute path.  It actually complained when I tried to pass it an absolute path if I remember correctly

ngl-tools/meson.build
jmoguillansky-gpsw: I can probably remove it...

ngl-tools/meson.build
jmoguillansky-gpsw: Yes its not the best naming.  When we run python -m venv on windows it sets up a scripts folder but no bin folder.  We can add a folder "bin" but we need to add it to path env variable in the activate script...

libnodegl/node_buffer.c
jmoguillansky-gpsw: Will replace with long int

libnodegl/api.c
jmoguillansky-gpsw: We should define "NGL_PLATFORM_WINDOWS" if TARGET_WINDOWS.  Not sure if we need two versions: "NGL_PLATFORM_WINDOWS and "NGL_PLATFORM_MINGW"?

Makefile
jmoguillansky-gpsw: PR submitted

external/scripts/sync.sh
jmoguillansky-gpsw: updated.  Using set -xu

external/scripts/sync.sh
jmoguillansky-gpsw: updated

external/scripts/sync.sh
jmoguillansky-gpsw: updated: using binaries from git repo.  They should survive forever

libnodegl/node_buffer.c
jmoguillansky-gpsw: already resolved

libnodegl/utils.c
jmoguillansky-gpsw: updated

libnodegl/backends/gl/glincludes.h
jmoguillansky-gpsw: This introduced compile errors.  No need to do this

libnodegl/api.c
jmoguillansky-gpsw: We have 3 macros:
TARGET_WINDOWS, TARGET_MINGW_W64, and TARGET_MSVC.
For this particular logic, it can be applied to both msvc and mingw (hence TARGET_WINDOWS)
We have the corresponding changes to meson.build and CI.

ngl-tools/ipc.c
jmoguillansky-gpsw: already resolved

pynodegl/setup.py
jmoguillansky-gpsw: updated

.gitattributes
```diff
@@ -0,0 +1 @@
+* -text
```
jmoguillansky-gpsw, line 1: The CI server is converting the line endings (e.g. \n to \r\n) automatically for various files.  I had to disable this

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,69 @@ jobs:
...
+          vcpkgArguments: pthreads opengl-registry ffmpeg[ffmpeg] sdl2
+          cleanAfterBuild: false
+
+      - name: vcpkg integrate install
+        run: |
+          cd vcpkg && ./vcpkg.exe integrate install
+
+      - name: Build
+        run: |
+          make -j8 TARGET_OS=Windows PKG_CONFIG="D:\\\\a\\\\gopro-lib-node.gl\\\\gopro-lib-node.gl\\\\pkg-config.exe" VCPKG_DIR="D:\\\\a\\\\gopro-lib-node.gl\\\\gopro-lib-node.gl\\\\vcpkg"
```
jmoguillansky-gpsw, line 70: can probably use single quotes here instead of lots of escape characters, still learning bash scripting...
https://wiki.bash-hackers.org/

.gitignore
```diff
@@ -1,2 +1,5 @@
-builddir
+/builddir
```
cboesch-gpsw, line 1: this looks unrelated

.gitignore
```diff
@@ -1,2 +1,5 @@
...
 nodegl-env/
+/sxplayer*
+/external/win64
+external/token.cfg
```
cboesch-gpsw, line 5: this is inconsistent and doesn't look like it doesn't belong in this commit but in the one introducing the external w64 stuff

.gitignore
```diff
@@ -1,2 +1,5 @@
-builddir
+/builddir
 nodegl-env/
+/sxplayer*
```
cboesch-gpsw, line 3: you can drop that

ngl-tools/meson.build
cboesch-gpsw: yes but `required` is set to `false` so why is it an issue?

libnodegl/backends/gl/glincludes.h
cboesch-gpsw: Mmh well then, `NGLI_GL_APIENTRY` needs to be drop entirely: it's sole purpose was to work with windows

ngl-tools/meson.build
```diff
@@ -120,6 +120,8 @@ tools_specs = {
 }
 
 if host_system == 'windows'
+  #Python -m venv on windows creates a Scripts folder instead of a bin folder
+  #TODO: add a bin folder and add it to path in activate script
```
cboesch-gpsw, line 124: 1. this diff doesn't belong in "nodegl: use TARGET_WINDOWS instead of TARGET_MINGW"
2. cosmetic: please add a space after `#`
3. what is the plan wrt the TODO?

Makefile
```diff
@@ -181,6 +181,8 @@ endif
 nodegl-install: nodegl-setup
 ifeq ($(TARGET_OS),Windows)
 	(cmd.exe /C $(ACTIVATE) \&\& $(MESON_COMPILE) -C builddir\\libnodegl \&\& $(MESON_INSTALL) -C builddir\\libnodegl)
+	# patch libnodegl.pc TODO: remove
```
cboesch-gpsw, line 196: what's your plan for this?

Makefile
```diff
@@ -225,13 +225,25 @@ else
 endif
 
 tests: nodegl-tests tests-setup
+ifeq ($(TARGET_OS),Windows)
+	(cmd.exe /C $(ACTIVATE) \&\& meson test $(MESON_TESTS_SUITE_OPTS) -C builddir\\tests)
```
cboesch-gpsw: if you include `cmd.exe /C` in `ACTIVATE`, does it help getting rid of the condition?

libnodegl/meson.build
```diff
@@ -344,7 +344,7 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
   foreach dep : gbackend_deps
     all_dep_found = all_dep_found and dep.found()
   endforeach
-
+  message('backend: ' + gbackend_name + ' enabled: ' + all_dep_found.to_string())
```
cboesch-gpsw, line 351: `message('backend: @0@ enabled: @1@'.format(gbackend_name, all_dep_found.to_string())` maybe, for consistency with the other messages we have in other `meson.build` files

Makefile
```diff
@@ -63,6 +63,7 @@ RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX)/lib
 
 ifeq ($(TARGET_OS),Windows)
 MESON_SETUP   = meson setup --backend vs --prefix="$(W_PREFIX)" --pkg-config-path=$(PREFIX)\\Lib\\pkgconfig -Drpath=true
+MESON_SETUP_NINJA   = meson setup --backend ninja --prefix="$(W_PREFIX)" --pkg-config-path=$(PREFIX)\\Lib\\pkgconfig -Drpath=true
```
cboesch-gpsw: why is this commit separated from the rest? isn't it needed at the moment windows support is introduced?

Makefile
```diff
@@ -200,9 +202,9 @@ endif
 
 sxplayer-install: external-download $(PREFIX)
 ifeq ($(TARGET_OS),Windows)
-	(cmd.exe /C $(ACTIVATE) \&\& $(MESON_SETUP) --default-library static external\\sxplayer builddir\\sxplayer \&\& $(MESON_COMPILE) -C builddir\\sxplayer \&\& $(MESON_INSTALL) -C builddir\\sxplayer)
+	(cmd.exe /C $(ACTIVATE) \&\& $(MESON_SETUP) --default-library shared external\\sxplayer builddir\\sxplayer \&\& $(MESON_COMPILE) -C builddir\\sxplayer \&\& $(MESON_INSTALL) -C builddir\\sxplayer)
```
cboesch-gpsw: same here, it looks like you want this squashed into the windows build commit

libnodegl/meson.build
```diff
@@ -373,14 +373,6 @@ lib_link_args += cc.get_supported_link_arguments([
...
 ])
 
-lib_name_prefix = []
-lib_name_suffix = []
-if host_system == 'windows'
-  # Follow Windows library convention
-  lib_name_prefix = ''
-  lib_name_suffix = 'lib'
-endif
-
```
cboesch-gpsw: same here: i think this belongs in the commit adding windows build support

Note: "update X" is not a good commit message, a commit pretty much always updates something

libnodegl/meson.build
```diff
@@ -62,8 +62,12 @@ conf_data.set10('DEBUG_GL', 'gl' in debug_opts)
...
 conf_data.set10('DEBUG_SCENE', 'scene' in debug_opts)
 
-if host_system == 'windows' and cc.get_id() != 'msvc'
-  conf_data.set10('TARGET_MINGW_W64', true)
+if host_system == 'windows'
+  if cc.get_id() == 'msvc'
+    conf_data.set10('TARGET_MSVC', true)
+  else
+    conf_data.set10('TARGET_MINGW_W64', true)
+  endif
```
cboesch-gpsw, line 70: please rebase, this should be available upstream

.gitattributes
cboesch-gpsw: please write that in the commit description (and add a "git" prefix to the commit message)

Makefile
```diff
@@ -219,12 +225,12 @@ external-download:
...
-	$(PYTHON) -m venv $(PREFIX)
+	(cmd.exe /C $(PYTHON) -m venv $(PREFIX))
 	(cmd.exe /C copy external\\win64\\pkg-config.exe nodegl-env\\Scripts)
 	(cmd.exe /C mkdir $(PREFIX)\\Lib\\pkgconfig)
-	(cmd.exe /C pushd external\\win64\\ffmpeg_x64-windows \&\& python.exe scripts/install.py "$(W_PREFIX)" \& popd)
-	(cmd.exe /C pushd external\\win64\\pthreads_x64-windows \&\& python.exe scripts/install.py "$(W_PREFIX)" \& popd)
-	(cmd.exe /C pushd external\\win64\\sdl2_x64-windows \&\& python.exe scripts/install.py "$(W_PREFIX)" \& popd)
+	(cmd.exe /C pushd external\\win64\\ffmpeg_x64-windows \&\& $(PYTHON) scripts/install.py "$(W_PREFIX)" \& popd)
+	(cmd.exe /C pushd external\\win64\\pthreads_x64-windows \&\& $(PYTHON) scripts/install.py "$(W_PREFIX)" \& popd)
+	(cmd.exe /C pushd external\\win64\\sdl2_x64-windows \&\& $(PYTHON) scripts/install.py "$(W_PREFIX)" \& popd)
```
cboesch-gpsw: again i think this commit belongs in another

libnodegl/utils.c
```diff
@@ -23,11 +23,14 @@
 #include <pthread.h>
 
 #ifdef _WIN32
+#define POW10_9 1000000000
+#include <Windows.h>
```
cboesch-gpsw, line 27: > Resolves compile error "No target architecture"

What's going on here? What following include requires `Windows.h` to be included first? Can you add the explanation in the commit description?

Also, we don't mention "bugfix" like this. "utils: fix Windows.h include order" is more consistent with other commits.

libnodegl/log.c
```diff
@@ -22,11 +22,11 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include "config.h"
 #if !defined(TARGET_IPHONE) && !defined(TARGET_ANDROID) && !defined(TARGET_WINDOWS)
 #include <unistd.h>
 #endif
 
-#include "config.h"
```
cboesch-gpsw, line 29: Aw. We need to merge that one quickly. Please mention in the commit description that it completes the initial attempt in 1494a18c11b2466e3c2c89fa133205bc3ac9be90

Makefile
```diff
@@ -48,7 +48,18 @@ TESTS_SUITE ?=
...
+VCPKG_DIR ?= C:\\vcpkg
+PKG_CONFIG ?= C:\\msys64\\usr\\bin\\pkg-config.exe
+SET_PKG_CONFIG_PATH = PKG_CONFIG_PATH="$(W_PREFIX)\Lib\pkgconfig" WSLENV=PKG_CONFIG_PATH/w
+CMD = $(SET_PKG_CONFIG_PATH) cmd.exe /C
+else
+ACTIVATE = $(PREFIX)/bin/activate
+endif
+
+ifeq ($(TARGET_OS), Windows)
+  ifneq ($(shell $(CMD) $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON_MAJOR))
```
cboesch-gpsw: again, please squash all of this in the same commit; we don't need to see all your iteration attempts.

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,69 @@ jobs:
...
+#        run: |
+#          # Wait for connection
+#          while [ -S /tmp/tmate.sock ]; do
+#            sleep 1
+#          done
+
+        
+#      - name: Run tests
+#        run: |
+#          make -j8 tests PYTHON="C:/hostedtoolcache/windows/Python/3.9.0/x64/python.exe" TARGET_OS=Windows
```
cboesch-gpsw, line 98: No dead code in the repository please: put that in your personal notes or write a documentation somewhere

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,69 @@ jobs:
...
+          distribution: Ubuntu-20.04
+
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          sudo apt -y update
+          sudo apt -y install build-essential zip
+          pip.exe install meson
+          wget https://github.com/jmoguillansky-gpsw/gopro-pkg/blob/master/external/win64/pkg-config.exe?raw=true -O pkg-config.exe
```
cboesch-gpsw, line 54: can we have something more official that this?

ngl-tools/meson.build
jmoguillansky-gpsw: it also makes it more clear to the developer which libraries are required for different platforms.  
I can remove this change if you prefer.

ngl-tools/meson.build
jmoguillansky-gpsw: 3: We can post a issue to python developers.  This is a python issue.  
For example, meson.exe is also in Scripts folder in python system path:

ls /mnt/c/Users/jmoguillansky/AppData/Local/Programs/Python/Python38/Scripts/

cygdb.exe   cythonize.exe         easy_install.exe  ngl-test-script.py  ngl-viewer-script.py  pip.exe     pip3.exe              pyside2-lupdate.exe  pyside2-uic.exe
cython.exe  easy_install-3.8.exe  meson.exe         ngl-test.exe        ngl-viewer.exe        pip3.8.exe  pyside2-designer.exe  pyside2-rcc.exe      wheel.exe

Makefile
jmoguillansky-gpsw: It seems that currently meson pkg-config export doesn't support lib paths with spaces.  
I will post an issue on meson github...

Makefile
jmoguillansky-gpsw: There's two differences between windows and others:
1) we use cmd.exe
2) we use different slashes (e.g. forward slash instead of backslashes).
It would be great if we can conslidate the code and avoid platform-specific code, but this can be a future improvement.  
It's not so easy to do this in a makefile, easier in a high level programming language like Python, etc.

libnodegl/meson.build
jmoguillansky-gpsw: ok I'll update...

Makefile
jmoguillansky-gpsw: the unit tests target regarding meson setup using ninja.  All the other targets can be built with visual studio backend.

.github/workflows/ci_win.yml
jmoguillansky-gpsw: Windows doesn't provide a pkg-config windows executable.  
I extracted it from msys2, and put it for now in a personal repo. 
There doesn't really seem to be a standalone windows pre-built binary available anywhere.  
We can either build this from source as part of node.gl build, or we can provide this in a public ftp server (or something equivalent).

libnodegl/utils.c
jmoguillansky-gpsw: I get the compile error:
C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\winnt.h(173,1): fatal error C1189: #error:  "No Target Architecture" [C:\Users\jmoguillansky\Documents\2020\Git\gopro-lib-node.gl-meson-windows-master\builddir\libnodegl\tes
t_utils@exe.vcxproj]

I put in the commit message it fixes compile error "No Target Architecture".  
Is this sufficient detail?

libnodegl/backends/gl/glincludes.h
jmoguillansky-gpsw: Can you please help with that change?  
If I recall correctly, glfunctions.h is autogenerated, I don't have to modify it by hand?

libnodegl/meson.build
cboesch-gpsw: I think that workaround is OK, thanks

ngl-tools/meson.build
cboesch-gpsw: We're just following the recommended practice from here: https://mesonbuild.com/howtox.html#add-math-library-lm-portably

The purpose of `required: false` really is all about this: we want as little platform specific logic and be feature/availability oriented. Please don't change something that works for something more complex.

jmoguillansky-gpsw:
Updated

jmoguillansky-gpsw (commented):
Note that these files were autogenerated on windows platform.  
This is why the line endings are different.

libnodegl/backends/gl/glincludes.h
mbouron-gpsw: NGLI_GL_APIENTRY is still needed. The build issue with MSVC is related to the invalid syntax we currently use (which is  accepted by gcc) for declaring the OpenGL function pointers. https://github.com/gopro/gopro-lib-node.gl/pull/166 should addresses the issue.

.github/workflows/ci_win.yml
jbnoel-gpsw: I agree with @cboesch-gpsw, we need to limit the hacks.
Ideas:
- prebuilt with github action like other dependencies ?
- https://stackoverflow.com/questions/1710922/how-to-install-pkg-config-in-windows ?

libnodegl/meson.build
```diff
@@ -491,38 +491,24 @@ endif
...
-gl_xml = custom_target(
-  'gl.xml',
-  command: [find_program('curl'), 'https://raw.githubusercontent.com/KhronosGroup/OpenGL-Registry/master/xml/gl.xml', '-o', '@OUTPUT@'],
-  capture: false,
-  output: 'gl.xml',
+message('** generate gl wrappers **')
+message('gl.xml path: ' + meson.current_build_dir() / 'gl.xml')
+run_command('curl',
+  'https://raw.githubusercontent.com/KhronosGroup/OpenGL-Registry/master/xml/gl.xml',
+  '-o', meson.current_build_dir() / 'gl.xml'
```
cboesch-gpsw, line 498: We don't want to do that at every run; please keep the custom target. It doesn't matter if it doesn't work on Windows, just call the script directly if needed.

pynodegl/.gitignore
```diff
@@ -4,3 +4,5 @@ pynodegl.c
 pynodegl.egg-info
 pynodegl.*.so
 .eggs
+*.pyd
+*.dll
```
cboesch-gpsw, line 8: Again "update X" is not a good commit message; please squash this into the corresponding commit

Makefile
```diff
@@ -19,36 +19,71 @@
...
 V           ?=
+NODEGL_DEFAULT_LIBRARY ?= shared
 
-ifneq ($(shell $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON_MAJOR))
-$(error "Python $(PYTHON_MAJOR) not found")
+ifeq ($(TARGET_OS), Windows)
+VCVARS64 ?= "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+ACTIVATE = $(VCVARS64) \&\& $(PREFIX)\\Scripts\\activate.bat
+VCPKG_DIR ?= C:\\vcpkg
+PKG_CONFIG ?= C:\\msys64\\usr\\bin\\pkg-config.exe
```
cboesch-gpsw, line 55: Can't we do something about these hardcoded path?

Makefile
```diff
@@ -119,48 +166,110 @@ pynodegl-utils-install: pynodegl-utils-deps-install
...
 
 pynodegl-install: pynodegl-deps-install
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& pip -v install -e .\\pynodegl)
+	# Install DLLs / Libs
+	($(CMD) xcopy /Y builddir\\sxplayer\\*.dll pynodegl\\.)
+	($(CMD) xcopy /Y /C builddir\\libnodegl\\*.dll pynodegl\\.)
+	($(CMD) xcopy /Y builddir\\sxplayer\\*.dll $(PREFIX)\\Scripts\\.)
+	($(CMD) xcopy /Y /C builddir\\libnodegl\\*.dll $(PREFIX)\\Scripts\\.)
+	($(CMD) xcopy /Y /C builddir\\libnodegl\\*.lib $(PREFIX)\\Lib\\.)
```
cboesch-gpsw, line 183: The `( ... )` form is a subshell and is only needed when you use the "activate", you can get rid of them

jmoguillansky-gpsw:
Regarding hardcoded path, thats the default value.  The user can override (and I override it locally because I use community edition)

jmoguillansky-gpsw:
Regarding pkg config: yes I'm currently working on a proper solution but its non trivial.  I was able to build "pkgconf" from source: this is supposed to be a drop in replacememt for pkg config but there's compatibility issues on Windows.  It seems I need to patch the source code too... This was meant as a temporary workaround.

jmoguillansky-gpsw:
What about using run_command on msvc and custom target on others?

libnodegl/utils.c
cboesch-gpsw: Yes but why re-ordering fix it? What includes set the architecture, and what is complaining about this missing architecture?

.gitattributes
cboesch-gpsw: Also, doesn't this affect all binaries? Like, is git going to consider our images in the doc as text now?

jmoguillansky-gpsw:
Winnt.h which indirectly included by some other include…
Not so easy to trace the culprit

jmoguillansky-gpsw:
gitattributes(5) (kernel.org)<https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html>

text
This attribute enables and controls end-of-line normalization. When a text file is normalized, its line endings are converted to LF in the repository. To control what line ending style is used in the working directory, use the eol attribute for a single file and the core.eol configuration variable for all text files. Note that setting core.autocrlf to true or input overrides core.eol (see the definitions of those options in git-config(1)<https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-config.html>).

We’re disabling this feature (the minus in front) so we’re disabling end-of-line normalization. It won’t affect any binaries, etc.  All this does is turn off end-of-line normalization of text files.

Makefile
```diff
@@ -89,14 +124,26 @@ all: ngl-tools-install pynodegl-utils-install
 	@echo "    Install completed."
 	@echo
 	@echo "    You can now enter the venv with:"
+ifeq ($(TARGET_OS),Windows)
+	@echo "        . $(PREFIX)\\Scripts\\activate.bat"
```
cboesch-gpsw, line 128: this file doesn't exist

Makefile
```diff
@@ -19,36 +19,71 @@
...
-$(error "Python $(PYTHON_MAJOR) not found")
+ifeq ($(TARGET_OS), Windows)
+VCVARS64 ?= "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+ACTIVATE = $(VCVARS64) \&\& $(PREFIX)\\Scripts\\activate.bat
+VCPKG_DIR ?= C:\\vcpkg
+PKG_CONFIG ?= C:\\msys64\\usr\\bin\\pkg-config.exe
+SET_PKG_CONFIG_PATH = PKG_CONFIG_PATH="$(W_PREFIX)\Lib\pkgconfig" WSLENV=PKG_CONFIG_PATH/w
+CMD = $(SET_PKG_CONFIG_PATH) cmd.exe /C
+else
+ACTIVATE = $(PREFIX)/bin/activate
```
cboesch-gpsw, line 59: why not integrate `cmd.exe` into `ACTIVATE` to help the different if branches?

Makefile
```diff
@@ -89,14 +124,26 @@ all: ngl-tools-install pynodegl-utils-install
...
+ifeq ($(TARGET_OS),Windows)
+	@echo "        . $(PREFIX)\\Scripts\\activate.bat"
+else
 	@echo "        . $(ACTIVATE)"
+endif
 	@echo
 
 ngl-tools-install: nodegl-install
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& $(MESON_SETUP) ngl-tools builddir\\ngl-tools \&\& $(MESON_COMPILE) -C builddir\\ngl-tools \&\& $(MESON_INSTALL) -C builddir\\ngl-tools)
```
cboesch-gpsw, line 136: why do you need to escape the `&`?

jmoguillansky-gpsw:
It exists on windows

libnodegl/backends/gl/glincludes.h
jmoguillansky-gpsw: I tried #166 but still get compile errors (see jmoguillansky/meson-windows-master-tmp-gl-apientry)...

libnodegl/meson.build
jmoguillansky-gpsw: how about if we use "run_command" on msvc and "custom_target" for others, and put a comment that this is a temporary solution?

Makefile
jmoguillansky-gpsw: those are the default params, the user can override

Makefile
jmoguillansky-gpsw: It exists on windows (with msvc)

Makefile
jmoguillansky-gpsw: cmd is running more commands than just activate.  It's doing activate and meson setup and meson compile, etc.  Putting cmd inside activate is a bit misleading

Makefile
jmoguillansky-gpsw: good question.  This is one of the tricky integration issues.  I'm passing ampersand to be processed by cmd executable instead of by bash shell (hence using the escape character).  I had to deal with various issues regarding slashes and ampersands etc.

.github/workflows/ci_win.yml
jmoguillansky-gpsw: Regarding pkg config: yes I'm currently working on a proper solution but its non trivial.  I was able to build "pkgconf" from source: this is supposed to be a drop in replacememt for pkg config but there's compatibility issues on Windows.  It seems I need to patch the source code too... This was meant as a temporary workaround.

FYI:
when I build with "pkg-config" from msys2:

Called `C:\Users\jmoguillansky\Documents\2020\Git\gopro-lib-node.gl-meson-windows-master\nodegl-env\Scripts\pkg-config.EXE sdl2 --libs` -> 0
-L -L/manual-link -lSDL2 -lSDL2main -ldinput8 -lshell32 -lsetupapi -ladvapi32 -luuid -lversion -loleaut32 -lole32 -limm32 -lwinmm -lgdi32 -luser32 -lSDL2

When I build with "pkgconf" (built from source):
Called `C:\Users\jmoguillansky\Documents\2020\Git\gopro-lib-node.gl-meson-windows-master\nodegl-env\Scripts\pkg-config.EXE sdl2 --libs` -> 0
-L -lSDL2 -L/manual-link -lSDL2main

I need to add param "--static".  There's various other incompatibilities like backslashes instead of forward slashes, etc

.gitattributes
jmoguillansky-gpsw: gitattributes(5) (kernel.org)<https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html>

text
This attribute enables and controls end-of-line normalization. When a text file is normalized, its line endings are converted to LF in the repository. To control what line ending style is used in the working directory, use the eol attribute for a single file and the core.eol configuration variable for all text files. Note that setting core.autocrlf to true or input overrides core.eol (see the definitions of those options in git-config(1)<https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-config.html>).

We’re disabling this feature (the minus in front) so we’re disabling end-of-line normalization. It won’t affect any binaries, etc.  All this does is turn off end-of-line normalization of text files.

ubitux: So it means someone working on windows may start submitting files with CLRF line endings?

Unless I'm mistaken, this is only needed for writing file, so it's not about the clone itself. Why do you need to write files in the first place? Is that related to the OpenGL wrapper thing?

Makefile
ubitux: My bad I misunderstood.

Makefile
ubitux: `CMD = .` then?

Makefile
ubitux: > we use different slashes (e.g. forward slash instead of backslashes).

I though windows was supporting `/` nowadays?

> It's not so easy to do this in a makefile, easier in a high level programming language like Python, etc.

I'm not opposed to having Windows handled outside the Makefile (we moved away from make for the build for a reason). I don't know if Python would be a great language for bootstrapping packages etc on Windows (because it requires an initial installation for the user and we will need to create a virtual env from another Python script, it might get clumsy), but I wouldn't be particularly opposed to that. Isn't Powershell moer appropriate? (I really don't know).

If you were under the impression that I was asking for Windows integration within the Makefile, I'm sorry. The goal of that Makefile is to bootstrap the environment: we kept make because on Unix it's extremely convenient. The keyword here is "convenient", so the question to answer on windows is: "what would be the most convenient way for a user to bootstrap the environment reliably?".

jmoguillansky-gpsw:
No what it means is that if you submit a text file with /n on windows, git will convert to /r/n.  We don't want git changing the line endings we can manage them ourselves

jmoguillansky-gpsw:
Regarding supporting backslashes: not command prompt and I'm not familiar with powershell.  I used bash and Makefile because its the minimal porting effort.

jmoguillansky-gpsw:
I don't think we should have makefile for one platform and something else for other platforms, ideally we use the same thing for both.  I'm ok with Makefile.

Makefile
jmoguillansky-gpsw: sure I'll update

jmoguillansky-gpsw: let me know if this is better: https://github.com/gopro/gopro-lib-node.gl/pull/152/commits/7ca08730e1720b34116ee94f0e5418e9a91fc3fc

jmoguillansky-gpsw:
some recent updates:
1) added params --includedir and --libdir when calling meson setup to point to $(PREFIX)\Include and $(PREFIX)\Lib.  
This also fixes the pkg-config generation issue: now the generated pkg-config files point to correct location.
2) updated pkg-config patch:
set params "--keep-system-cflags and --keep-system-libs" automatically.  This resolves an issue where various pkg-config cflags and libs params were being filtered out by pkg-config (pkg-config thinks these directories are system paths but they're not).

cboesch-gpsw:
Merged with https://github.com/gopro/gopro-lib-node.gl/pull/177

PR #153, opened 2020-12-01
Port misc code to windows

jmoguillansky-gpsw:


libnodegl/log.c
```diff
@@ -22,8 +22,9 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
+#ifndef _WIN32
```
cboesch-gpsw, line 25: please keep this check coherent with the check at line 45: `#if !defined(TARGET_IPHONE) && !defined(TARGET_ANDROID) && !defined(TARGET_MINGW_W64)`

But ideally, I'd suggest to make a dedicated commit in which you add `conf_data.set10('HAVE_ISATTY', cc.has_function('isatty'))` to the `meson.build` and protect the include+code with `#if HAVE_ISATTY`

libnodegl/node_camera.c
```diff
@@ -23,7 +23,6 @@
 #include <stddef.h>
 #include <stdio.h>
 #include <string.h>
-#include <unistd.h>
```
cboesch-gpsw, line 26: Can we have a dedicated commit for this: "camera: remove unused unistd.h include"?

That way, with the `isatty` change isolated as well, you can reword the current commit to something more explicit: "internal: replace read/write with fread/fwrite", and in the description: "This removes the unix specific unistd.h include, not supported with MSVC."

libnodegl/utils.c
```diff
@@ -47,14 +47,6 @@ char *ngli_strdup(const char *s)
     return r;
 }
 
-int64_t ngli_gettime(void)
```
cboesch-gpsw, line 50: please add the "unused" qualifier in the commit message.

libnodegl/utils.c
```diff
@@ -27,7 +27,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef _WIN32
+#include <Windows.h>
```
cboesch-gpsw, line 31: We are using the non-capitalized form in `libnodegl/backends/gl/glcontext_wgl.c`: which one is correct?

libnodegl/utils.c
```diff
@@ -49,10 +53,20 @@ char *ngli_strdup(const char *s)
 
 int64_t ngli_gettime_relative(void)
 {
+#ifdef _WIN32
+    FILETIME t0;
+    GetSystemTimeAsFileTime(&t0);
```
cboesch-gpsw, line 58: This is not appropriate for a monotonic clock: you probably need to use `QueryPerformance{Frequency,Counter}`. `GetSystemTimeAsFileTime` would only be appropriate for the real time.

ngl-tools/common.c
```diff
@@ -23,27 +23,45 @@
...
+#else
+#include <sys/time.h>
+#endif
 
 #include "common.h"
 
 int64_t gettime(void)
 {
+#ifdef _WIN32
+    return gettime_relative();
```
cboesch-gpsw, line 41: This will become invalid when you will fix `gettime_relative()` to be properly monotonic.

ngl-tools/common.c
```diff
@@ -23,27 +23,45 @@
...
+    t1.LowPart = t0.dwLowDateTime;
+    t1.HighPart = t0.dwHighDateTime;
+    int64_t t2 = t1.QuadPart;
+    return t2 / 10;
+#else
     struct timespec ts;
-    int ret = clock_gettime(CLOCK_MONOTONIC, &ts);
-    return ret != 0 ? 0 : ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return 1000000 * (int64_t)ts.tv_sec + ts.tv_nsec / 1000;
```
cboesch-gpsw, line 63: What is this change for?

ngl-tools/ngl-serialize.c
```diff
@@ -19,12 +19,17 @@
  * under the License.
  */
 
+#ifdef _WIN32
+#define STDOUT_FILENO _fileno(stdout)
+#define STDERR_FILENO _fileno(stderr)
```
cboesch-gpsw, line 24: I believe you want those only in `_MSC_VER`

ngl-tools/opts.c
```diff
@@ -22,7 +22,6 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/time.h>
```
cboesch-gpsw, line 25: A dedicated commit for this unused include please

ngl-tools/ngl-serialize.c
```diff
@@ -19,12 +19,17 @@
...
+#else
+#include <unistd.h>
```
cboesch-gpsw, line 26: All system includes need to be after the definition of `_POSIX_C_SOURCE`

ngl-tools/ngl-ipc.c
```diff
@@ -28,11 +28,12 @@
 #ifdef _WIN32
 #include <winsock2.h>
 #include <ws2tcpip.h>
+#define ssize_t int
```
cboesch-gpsw, line 31: I'm uncomfortable setting `ssize_t` for `_WIN32` (it's defined on mingw), I'd rather have it for MSVC only. That said, in this file, I think you can make `recv(s->upload_fd, s->upload_buffer, UPLOAD_CHUNK_SIZE, 0)` return an int: `UPLOAD_CHUNK_SIZE` will never require more than what an int can store, so it's fine to downcast on platforms that have a `ssize_t` returned.

ngl-tools/ipc.c
```diff
@@ -22,10 +22,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <unistd.h>
 #ifdef _WIN32
 #include <winsock2.h>
+#define ssize_t int
```
cboesch-gpsw, line 27: Similar to my other comment in `ngl-ipc`, I think it's best if we use an `int` here as well. Please make a dedicated commit for that one and the change in `ngl-ipc`: a commit specially about `ssize_t`

cboesch-gpsw (changes requested):

libnodegl/utils.c
jmoguillansky-gpsw: Windows.h (with uppercase)

ngl-tools/common.c
jmoguillansky-gpsw: I'll remove it.  Probably got there by accident

libnodegl/utils.c
cboesch-gpsw: Alright: would you mind adding a commit to fix the one in `glcontext_wgl.c`?

libnodegl/log.c
jmoguillansky-gpsw: I can add the following logic: 
in meson.build:
define TARGET_WINDOWS + TARGET_MINGW_W64 on windows / mingw,
define TARGET_WINDOWS + TARGET_MSVC on windows /  msvc
use: #if !defined(TARGET_IPHONE) && !defined(TARGET_ANDROID) && !defined(TARGET_WINDOWS)

libnodegl/meson.build
```diff
@@ -62,8 +62,12 @@ conf_data.set10('DEBUG_GL', 'gl' in debug_opts)
 conf_data.set10('DEBUG_MEM', 'mem' in debug_opts)
 conf_data.set10('DEBUG_SCENE', 'scene' in debug_opts)
 
-if host_system == 'windows' and cc.get_id() != 'msvc'
-  conf_data.set10('TARGET_MINGW_W64', true)
+if host_system == 'windows'
+  if cc.get_id() != 'msvc'
```
mbouron-gpsw, line 66: Maybe use this form:
`if cc.get_id() == 'msvc'
  conf_data.set10('TARGET_MSVC', true)
else
  conf_data.set10('TARGET_MINGW_W64', true)
endif`

libnodegl/log.c
```diff
@@ -22,7 +22,9 @@
 #include <stdarg.h>
 #include <stdio.h>
 #include <stdlib.h>
+#if !defined(TARGET_IPHONE) && !defined(TARGET_ANDROID) && !defined(TARGET_WINDOWS)
 #include <unistd.h>
+#endif
```
mbouron-gpsw, line 27: This should be part of a separate commit (it is not tied to the usage of open/read/write)).

libnodegl/log.c
```diff
@@ -42,7 +44,7 @@ static void default_callback(void *arg, int level, const char *filename, int ln,
 
     const char *color_start = "", *color_end = "";
 
-#if !defined(TARGET_IPHONE) && !defined(TARGET_ANDROID) && !defined(TARGET_MINGW_W64)
+#if !defined(TARGET_IPHONE) && !defined(TARGET_ANDROID) && !defined(TARGET_WINDOWS)
```
mbouron-gpsw, line 47: Same as above.

libnodegl/node_buffer.c
```diff
@@ -127,15 +126,16 @@ static int buffer_init_from_filename(struct ngl_node *node)
...
+    s->fp = fopen(s->filename, "r");
+    if (!s->fp) {
         LOG(ERROR, "could not open '%s'", s->filename);
         return NGL_ERROR_IO;
     }
 
-    off_t filesize = lseek(s->fd, 0, SEEK_END);
-    off_t ret      = lseek(s->fd, 0, SEEK_SET);
-    if (filesize < 0 || ret < 0) {
+    fseek(s->fp, 0, SEEK_END);
```
mbouron-gpsw, line 135: Shouldn't the return of fseek be checked ?

libnodegl/node_buffer.c
```diff
@@ -155,7 +155,7 @@ static int buffer_init_from_filename(struct ngl_node *node)
     if (!s->data)
         return NGL_ERROR_MEMORY;
 
-    ssize_t n = read(s->fd, s->data, s->data_size);
+    size_t n = fread(s->data, 1, s->data_size, s->fp);
     if (n < 0) {
```
mbouron-gpsw, line 159: The error condition should be adjusted as well, fread cannot return a negative number. Quoting the documenation:
`fread() does not distinguish between end-of-file and error, and callers must use feof(3) and ferror(3) to determine which occurred.`

libnodegl/utils.c
```diff
@@ -49,10 +54,23 @@ char *ngli_strdup(const char *s)
 
 int64_t ngli_gettime_relative(void)
 {
+#ifdef _WIN32
+//reference: https://github.com/mirror/mingw-w64/blob/master/mingw-w64-libraries/winpthreads/src/clock.c
```
mbouron-gpsw, line 58: Please align the comment with the code bellow.

libnodegl/utils.c
```diff
@@ -49,10 +54,23 @@ char *ngli_strdup(const char *s)
...
+    LARGE_INTEGER pf, pc;
+    QueryPerformanceFrequency(&pf);
+    QueryPerformanceCounter(&pc);
+    int64_t tv_sec = pc.QuadPart / pf.QuadPart;
+    int64_t tv_nsec = (int) (((pc.QuadPart % pf.QuadPart) * POW10_9 + (pf.QuadPart >> 1)) / pf.QuadPart);
```
mbouron-gpsw, line 63: extra space between (int) and (((pc.QuadPart % pf.QuadPart) * POW10_9 + (pf.QuadPart >> 1)) / pf.QuadPart);

libnodegl/utils.c
```diff
@@ -49,10 +54,23 @@ char *ngli_strdup(const char *s)
...
+        tv_nsec -= POW10_9;
+    }
+    return tv_sec * 1000000 + tv_nsec / 1000;
+#else
     struct timespec ts;
-
-    clock_gettime(CLOCK_MONOTONIC, &ts);
-    return 1000000 * (int64_t)ts.tv_sec + ts.tv_nsec / 1000;
+    int ret = clock_gettime(CLOCK_MONOTONIC, &ts);
+    return ret != 0 ? 0 : ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
```
mbouron-gpsw, line 72: This should be part of another commit that checks the return of clock_gettime().

ngl-tools/common.c
```diff
@@ -23,27 +23,56 @@
...
 
     gettimeofday(&tv, NULL);
     return 1000000 * (int64_t)tv.tv_sec + tv.tv_usec;
+#endif
 }
 
 int64_t gettime_relative(void)
 {
+#ifdef _WIN32
+//reference: https://github.com/mirror/mingw-w64/blob/master/mingw-w64-libraries/winpthreads/src/clock.c
```
mbouron-gpsw, line 60: Please align the comment with the code below.

ngl-tools/common.c
```diff
@@ -23,27 +23,56 @@
...
+    LARGE_INTEGER pf, pc;
+    QueryPerformanceFrequency(&pf);
+    QueryPerformanceCounter(&pc);
+    int64_t tv_sec = pc.QuadPart / pf.QuadPart;
+    int64_t tv_nsec = (int) (((pc.QuadPart % pf.QuadPart) * POW10_9 + (pf.QuadPart >> 1)) / pf.QuadPart);
```
mbouron-gpsw, line 65: Extra space between (int) and (((pc.QuadPart % pf.QuadPart) * POW10_9 + (pf.QuadPart >> 1)) / pf.QuadPart);

ngl-tools/common.c
```diff
@@ -85,23 +114,23 @@ char *get_text_file_content(const char *filename)
...
+        const int needed = pos + BUF_SIZE + 1;
         void *new_buf = realloc(buf, needed);
         if (!new_buf) {
             free(buf);
             buf = NULL;
             goto end;
         }
         buf = new_buf;
-        const ssize_t n = read(fd, buf + pos, BUF_SIZE);
+        const int n = fread(buf + pos, 1, BUF_SIZE, fp);
```
mbouron-gpsw, line 133: int -> size_t

ngl-tools/common.c
```diff
@@ -85,23 +114,23 @@ char *get_text_file_content(const char *filename)
...
         if (n < 0) {
```
mbouron-gpsw, line 134: The error condition should be adjusted (n cannot be negative with fread()).

ngl-tools/ipc.c
```diff
@@ -189,11 +189,11 @@ int ipc_send(int fd, const struct ipc_pkt *pkt)
...
 }
 
-static int readbuf(int fd, uint8_t *buf, ssize_t size)
+static int readbuf(int fd, uint8_t *buf, int size)
 {
-    ssize_t nr = 0;
+    int nr = 0;
     while (nr != size) {
-        const ssize_t n = recv(fd, buf + nr, size - nr, 0);
+        const int n = recv(fd, buf + nr, size - nr, 0);
```
mbouron-gpsw, line 196: n should still be a ssize_t.

libnodegl/utils.c
mbouron-gpsw: Alse please add a space between // and the comment

libnodegl/node_buffer.c
mbouron-gpsw: Does this handle files >= 4gb ?

jmoguillansky-gpsw: we use ftell to get the filesize right after

libnodegl/log.c
ubitux: or you can just use `defined(_WIN32)` for both

libnodegl/node_buffer.c
jmoguillansky-gpsw: the "data_size" variable is a 32 bit int, so I'm not sure this current code would handle 4 gb files.  Not sure we have to worry about this issue at this point?

libnodegl/node_buffer.c
jmoguillansky-gpsw: updated

libnodegl/utils.c
jmoguillansky-gpsw: updated

libnodegl/utils.c
jmoguillansky-gpsw: The main motivation was to have the same code as ngl-tools/common.c but you're right checking the return value could be helpful

libnodegl/utils.c
jmoguillansky-gpsw: updated

libnodegl/log.c
jmoguillansky-gpsw: Ok, but let's resolve the higher-priority issues first

ngl-tools/common.c
jmoguillansky-gpsw: Updated

ngl-tools/common.c
jmoguillansky-gpsw: updated

ngl-tools/common.c
jmoguillansky-gpsw: updated

ngl-tools/common.c
jmoguillansky-gpsw: I think this code can be simplified significantly:
Seek to the end of the file, get the file size, and then read the entire contents with one call to fread?

jmoguillansky-gpsw: nm I see, you may be reading from stdin when filename is empty string

jmoguillansky-gpsw: When you read from stdin, how do you know that end of data is reached?  This code assumes that the only input to stdin is the buffer data and nothing else?

jmoguillansky-gpsw: Might be better if you pass the data size and then the data

ubitux: > When you read from stdin, how do you know that end of data is reached? This code assumes that the only input to stdin is the buffer data and nothing else?

`read()` returns `0` at `EOF`:
```c
        if (n == 0) {
            buf[pos] = 0;
            break;
        }
```

libnodegl/node_buffer.c
mbouron-gpsw: We don't have to worry about supporting files >= 4gb for now, I was just curious about how it is handled with these functions. I think we should check the return of fseek() so we indicate that an error happened at the file level, otherwise we will reach the error block below:
``` c
    if (s->data_size != s->count * s->data_stride) {
        LOG(ERROR,
            "element count (%d) and data stride (%d) does not match data size (%d)",
            s->count,
            s->data_stride,
            s->data_size);
        return NGL_ERROR_INVALID_DATA;
    }
```

which is not too indicative about the original error.

libnodegl/utils.c
mbouron-gpsw: Makes sense, just split theses changes in a separate commit and it is fine.

ngl-tools/common.c
jmoguillansky-gpsw: yes I know I understand how the code works.  I meant is this the right logic?  With this logic we can't pass additional data after the buffer in stdin.

libnodegl/log.c
jmoguillansky-gpsw: TARGET_WINDOWS is basically synonymous with _WIN32.  I can remove TARGET_WINDOWS?

libnodegl/log.c
jmoguillansky-gpsw: should this be TARGET_MSVC?  Windows / msvc doesn't support unistd.h, and also related to open/write/write

libnodegl/log.c
mbouron-gpsw: Isn't this one about isatty() ?

libnodegl/utils.c
jmoguillansky-gpsw: updated

libnodegl/node_buffer.c
```diff
@@ -155,8 +157,8 @@ static int buffer_init_from_filename(struct ngl_node *node)
     if (!s->data)
         return NGL_ERROR_MEMORY;
 
-    ssize_t n = read(s->fd, s->data, s->data_size);
-    if (n < 0) {
+    size_t n = fread(s->data, 1, s->data_size, s->fp);
+    if (n != s->data_size) {
         LOG(ERROR, "could not read '%s': %zd", s->filename, n);
```
mbouron-gpsw, line 162: The error message should be changed to LOG(ERROR, "could not read '%s'", s->filename);

ngl-tools/common.c
```diff
@@ -85,37 +114,37 @@ char *get_text_file_content(const char *filename)
...
-        const ssize_t n = read(fd, buf + pos, BUF_SIZE);
-        if (n < 0) {
+        const size_t n = fread(buf + pos, 1, BUF_SIZE, fp);
+        if (ferror(n)) {
             free(buf);
             buf = NULL;
             goto end;
         }
-        if (n == 0) {
+        if (feof(fp)) {
```
mbouron-gpsw, line 139: You want to increment pos before checking for EOF otherwise buf will be incomplete.

jmoguillansky-gpsw: there's already pos += n later.  If we hit EOF we goto end, and pos isn't used there?

libnodegl/log.c
jmoguillansky-gpsw: Sorry you're right.  updated

ngl-tools/common.c
mbouron-gpsw: If you check for EOF before executing pos +=n, the null character will be inserted in the wrong place truncating the buffer.

ngl-tools/common.c
```diff
@@ -85,37 +114,37 @@ char *get_text_file_content(const char *filename)
...
         if (!new_buf) {
             free(buf);
             buf = NULL;
             goto end;
         }
         buf = new_buf;
-        const ssize_t n = read(fd, buf + pos, BUF_SIZE);
-        if (n < 0) {
+        const size_t n = fread(buf + pos, 1, BUF_SIZE, fp);
+        if (ferror(n)) {
```
mbouron-gpsw, line 134: ferror(n) -> ferror(fp)

ngl-tools/common.c
jmoguillansky-gpsw: sorry you're right.  updated

ngl-tools/common.c
jmoguillansky-gpsw: thanks, updated

ngl-tools/ngl-ipc.c
```diff
@@ -114,7 +110,7 @@ static int craft_packet(struct ctx *s, struct ipc_pkt *pkt)
         const size_t name_size = name_len + 1;
 
         const char *filename = s->uploadfile + name_size;
-        s->upload_fd = open(filename, O_RDONLY | O_BINARY);
+        s->upload_fd = open(filename, O_RDONLY);
```
mbouron-gpsw, line 113: This should be ported to fopen().

ngl-tools/ngl-ipc.c
```diff
@@ -255,7 +251,7 @@ static int handle_response(struct ctx *s, const struct ipc_pkt *pkt)
     if (s->upload_fd > 0) {
         ipc_pkt_reset(s->send_pkt);
 
-        const ssize_t n = read(s->upload_fd, s->upload_buffer, UPLOAD_CHUNK_SIZE);
+        const int n = recv(s->upload_fd, s->upload_buffer, UPLOAD_CHUNK_SIZE, 0);
```
mbouron-gpsw, line 254: fread().

ngl-tools/ngl-ipc.c
mbouron-gpsw: The ngl-desktop.c counterpart should also be ported to fopen()/fwrite().

jmoguillansky-gpsw: strange that there's no compiler errors on windows

jmoguillansky-gpsw: updated

ngl-tools/ngl-ipc.c
jmoguillansky-gpsw: updated

mbouron-gpsw:
Merged in https://github.com/gopro/gopro-lib-node.gl/pull/153.

PR #157, opened 2020-12-04
tools: use read/write instead of recv/send where appropriate

mbouron-gpsw:
The recv and send functions must only be used on sockets (otherwise they
return ENOTSOCK).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,10 +110,14 @@ pynodegl-utils-install: pynodegl-utils-deps-install
 # available on PyPi.
 #
 # We do not pull the requirements on Windows because of various issues:
-# - PySide2 can't be pulled
+# - PySide2 can't be pulled (required to be installed by the user outside the
+#   Python virtualenv)
 # - Pillow fails to find zlib (required to be installed by the user outside the
 #   Python virtualenv)
-# - ngl-control can not currently work because of temporary files handling
+# - ngl-control can not currently work because of:
+#     - temporary files handling
+#     - subprocess usage, passing fd is not supported on Windows
+#     - subprocess usage, Windows cannot execute directly hooks shell scripts
 #
 # Still, we want the module to be installed so we can access the scene()
 # decorator and other related utils.
--- a/doc/howto/installation.md
+++ b/doc/howto/installation.md
@@ -49,7 +49,9 @@ On Windows, the bootstrap is slightly more complex:
 pacman -Syuu  # and restart the shell
 pacman -S git make
 pacman -S mingw-w64-x86_64-{toolchain,ffmpeg,python}
+pacman -S mingw-w64-x86_64-python-watchdog
 pacman -S mingw-w64-x86_64-python3-{pillow,pip}
+pacman -S mingw-w64-x86_64-pyside2-qt5
 pacman -S mingw-w64-x86_64-meson
 make TARGET_OS=MinGW-w64
 ```
--- a/ngl-tools/ngl-desktop.c
+++ b/ngl-tools/ngl-desktop.c
@@ -376,6 +376,33 @@ static int handle_commands(struct ctx *s, int fd)
     }
 }
 
+static void close_socket(int socket)
+{
+    /*
+     * On Solaris and MacOS, close() is the only way to terminate the
+     * blocking accept(). The shutdown() call will fail with errno ENOTCONN
+     * on these systems.
+     *
+     * On Linux though, shutdown() is required to terminate the blocking
+     * accept(), because close() will not. The reason for this difference
+     * of behaviour is to prevent a possible race scenario where the same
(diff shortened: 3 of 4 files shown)


doc/howto/installation.md
```diff
@@ -49,7 +49,9 @@ On Windows, the bootstrap is slightly more complex:
 pacman -Syuu  # and restart the shell
 pacman -S git make
 pacman -S mingw-w64-x86_64-{toolchain,ffmpeg,python}
+pacman -S mingw-w64-x86_64-python-watchdog
 pacman -S mingw-w64-x86_64-python3-{pillow,pip}
+pacman -S mingw-w64-x86_64-pyside2-qt5
```
ubitux, line 54: Any reason to ask users to install Qt etc if ngl-control doesn't work anyway?

ngl-tools/ngl-desktop.c
```diff
@@ -377,6 +377,13 @@ static int handle_commands(struct ctx *s, int fd)
 
 static void close_conn(struct ctx *s, int conn_fd)
 {
+#if _WIN32
+    closesocket(conn_fd);
```
ubitux, line 381: Can you explain what happens when a user timeout on MinGW and why this is necessary?

ngl-tools/ngl-desktop.c
```diff
@@ -377,6 +377,13 @@ static int handle_commands(struct ctx *s, int fd)
...
+#else
+    if (shutdown(conn_fd, SHUT_RDWR) < 0)
+        perror("shutdown");
```
ubitux, line 384: Is this required? If so, why?

Shouldn't we factor that out with what's in the `main`?

Makefile
```diff
@@ -110,10 +110,14 @@ pynodegl-utils-install: pynodegl-utils-deps-install
 # available on PyPi.
 #
 # We do not pull the requirements on Windows because of various issues:
-# - PySide2 can't be pulled
+# - PySide2 can't be pulled (required to be installed by the user outside the
+#   Python virtualenv
```
ubitux, line 114: missing `)`

mbouron-gpsw: Fixed.

doc/howto/installation.md
mbouron-gpsw: It makes the environement ready for ngl-control even it does not work on it yet and it eases the process for a developer who wants to port ngl-control to MinGW / Windows.

ngl-tools/ngl-desktop.c
```diff
@@ -219,7 +219,7 @@ static int handle_tag_filepart(struct ctx *s, const uint8_t *data, int size)
         return ipc_pkt_add_rtag_fileend(s->send_pkt, s->upload_path);
     }
 
-    const ssize_t n = send(s->upload_fd, data, size, 0);
+    const ssize_t n = write(s->upload_fd, data, size);
```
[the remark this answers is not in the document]
mbouron-gpsw: Done.

ngl-tools/ngl-desktop.c
mbouron-gpsw: Done.

ngl-tools/ngl-desktop.c
mbouron-gpsw: I've factored it out. It is not required on all platform, but as I said in the commit description it does not hurt to call it on all the platforms as it makes sense to me, we want to disallow all transmissions before closing the socket.

ngl-tools/ngl-desktop.c
```diff
@@ -634,13 +635,13 @@ int main(int argc, char *argv[])
          * See https://bugzilla.kernel.org/show_bug.cgi?id=106241 for more
          * information.
          */
+        if (shutdown(s.sock_fd, SHUT_RDWR) < 0)
+            perror("shutdown");
```
cboesch-gpsw: You need to update the comment stating "On Windows, both shutdown() and close() won't work, we have to use a Windows specific one: closesocket()". What I meant originally is that `shutdown()` nor `close()` terminate the blocking `accept()`, but you can add that it's recommended for a graceful shutdown by Microsoft.

ngl-tools/ngl-desktop.c
cboesch-gpsw: You have a typo in the commit message: "This allows to call the proper function **to the** close the socket" (in the description of "tools/desktop: use close_socket() to shutdown client connections")

mbouron-gpsw: Fixed.

ngl-tools/ngl-desktop.c
mbouron-gpsw: Updated.

cboesch-gpsw (approved):

PR #164, opened 2021-01-05
api: add skew node

cboesch-gpsw:


libnodegl/node_skew.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+static int skew_init(struct ngl_node *node)
+{
+    struct skew_priv *s = node->priv_data;
+    static const float zvec[3] = {0};
+    if (!memcmp(s->axis, zvec, sizeof(s->axis))) {
+        LOG(ERROR, "(0.0, 0.0, 0.0) is not a valid axis");
+        return NGL_ERROR_INVALID_ARG;
+    }
```
mbouron-gpsw, line 53: Should we also normalize the axis before using the compute the shear matrix ?

libnodegl/node_skew.c
```diff
@@ -0,0 +1,114 @@
...
+    return ngli_node_update(child, t);
+}
+
+#define OFFSET(x) offsetof(struct skew_priv, x)
+static const struct node_param skew_params[] = {
+    {"child",   PARAM_TYPE_NODE, OFFSET(trf.child),
+                .flags=PARAM_FLAG_NON_NULL,
+                .desc=NGLI_DOCSTRING("scene to skew")},
+    {"factors", PARAM_TYPE_VEC3,  OFFSET(factors),
```
mbouron-gpsw, line 91: 1. Are the factors between -1 and 1 ? If so, we should probably add checks for that.
2. Is it preferred to uses factors instead of angles ?

libnodegl/node_skew.c
ubitux: Done for consistency with the rotate node, but the axis merely is a bit/flag. We may consider the idea of changing the type of this option in the future.

libnodegl/node_skew.c
ubitux: That's a really good point. In this new iteration, I replaced the factors with angles because they are indeed more user friendly, and I didn't set any limit for those (it's handy for users to be able to circle around, just like the rotation).

Also, I added the anchor support (pretty much copied from the rotate node), and `NGLI_RAD2DEG()` in the math utils (which I use in both rotate and skew).

mbouron-gpsw (approved):

PR #165, opened 2021-01-05
pgcraft: fix integer based types transmission between stages

cboesch-gpsw:
Integers can not be interpolated like rational types are, so GLSL
require a qualifier such as "flat" to be used.

With MESA, it fixes the following error:
  error: if a fragment input is (or contains) an integer, then it must
  be qualified with 'flat'


ubitux:
> If I understand correctly, you're passing a uniform color to vertex shader, then copying it to a vertex shader output int, then it gets transferred to per fragment input int? Why not just pass the uniform color directly to fragment shader?

It's just a test. I spotted the issue while transferring integer data from the vertex shader to the fragment shader; typically in my case it was the `gl_VertexID`. Maybe I could use that in the test, but it doesn't matter what's the data, the scope here is to test the integer transfer itself.

mbouron-gpsw (approved):

PR #166, opened 2021-01-07
glfunctions: fix functions declarations on MSCV

mbouron-gpsw:
Fixes a build issue on MSCV caused by an invalid syntax used for OpenGL
functions declarations.

This issue is not present on MinGW because gcc accepts this invalid
syntax.


jmoguillansky-gpsw:
I tested this fix locally but it didn't work:

C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\um\GL\gl.h(1157,16): error C2054: expected '(' to follow 'WINGDIAPI' [C:\Users\jmoguillansky\Documents\2020\Git\gopro-lib-node.gl-meson-windows-master\builddir\libnodegl\nodegl
@sha.vcxproj]

jmoguillansky-gpsw:
Thanks this fix worked.  There was one additional change I had to make:
in glincludes.h include Windows.h before including GL.h

PR #173, opened 2021-01-22
internal: use __func__ instead of __FUNCTION__

mbouron-gpsw:
__func__ is standardised and is part of C99 whereas __FUNCTION__ is not.

See https://tool.oschina.net/uploads/apidocs/gcc-4.7.1-manual/Function-Names.html.

### DIFF
--- a/libnodegl/log.h
+++ b/libnodegl/log.h
@@ -26,7 +26,7 @@
 #include "utils.h"
 
 #ifndef __PRETTY_FUNCTION__
-#define __PRETTY_FUNCTION__ __FUNCTION__
+#define __PRETTY_FUNCTION__ __func__
 #endif
 
 #define LOG(log_level, ...) ngli_log_print(NGL_LOG_##log_level, __FILE__, __LINE__, __PRETTY_FUNCTION__, __VA_ARGS__)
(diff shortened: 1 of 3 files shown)


libnodegl/log.h
```diff
@@ -26,10 +26,10 @@
 #include "utils.h"
 
 #ifndef __PRETTY_FUNCTION__
-#define __PRETTY_FUNCTION__ __FUNCTION__
+#define __PRETTY_FUNCTION__ __func__
```
cboesch-gpsw, line 29: If you're not using `__PRETTY_FUNCTION__` anymore below, you can probably remove the ifdefery, especially since actually check for `__PRETTY_FUNCTION__` might be wrong (could be something else than a preproc define according to GCC doc)

mbouron: __PRETTY_FUNCTION__ will be useful when calling the log function from C++ code which is something that can happen in the future.

ubitux: Yeah but you're not using it in the log anymore

mbouron: oh, that's a mistake then.

mbouron: Fixed.

cboesch-gpsw (approved):

PR #176, opened 2021-01-25
Numbered logging

cboesch-gpsw:


libnodegl/log.c
```diff
@@ -28,19 +28,36 @@
...
         [NGL_LOG_DEBUG]   = "DEBUG",
         [NGL_LOG_VERBOSE] = "VERBOSE",
         [NGL_LOG_INFO]    = "INFO",
         [NGL_LOG_WARNING] = "WARNING",
         [NGL_LOG_ERROR]   = "ERROR",
     };
-    vsnprintf(logline, sizeof(logline), fmt, vl);
+
+    /* we need a copy because it may be re-used a 2nd time */
+    va_list vl2;
```
mbouron-gpsw, line 48: vl2 -> vl_copy

cboesch-gpsw: Renamed

libnodegl/backends/gl/program_gl.c
```diff
@@ -285,8 +285,15 @@ int ngli_program_gl_init(struct program *s, const char *vertex, const char *frag
         ngli_glShaderSource(gl, shader, 1, &shaders[i].src, NULL);
         ngli_glCompileShader(gl, shader);
         ret = program_check_status(gl, shader, GL_COMPILE_STATUS);
-        if (ret < 0)
+        if (ret < 0) {
+            LOG(ERROR, "failed to compile:");
+            char *s_with_numbers = ngli_numbered_lines(shaders[i].src);
+            if (s_with_numbers) {
+                LOG(ERROR, "\n%s", s_with_numbers);
```
mbouron-gpsw, line 292: Maybe: LOG(ERROR, "failed to compile:\n%s", s_with_numbers); and remove the call to LOG() above.

libnodegl/utils.c
```diff
@@ -194,3 +195,53 @@ int ngli_get_filesize(const char *filename, int64_t *size)
...
+    struct bstr *b = ngli_bstr_create();
+    if (!b)
+        return NULL;
+
+    const int nb_lines = count_lines(s);
+    const int nb_digits = count_digits(nb_lines);
+    int line = 1;
+    while (*s) {
+        const size_t len = strcspn(s, "\n");
+        ngli_bstr_printf(b, "%*d %.*s\n", nb_digits, line++, (int)len, s);
```
mbouron-gpsw, line 231: Do we really need to cast to int ? can't we use unsigned directly ?

libnodegl/backends/gl/program_gl.c
ubitux: Don't we want to print something in case the shader fails to compile but the driver doesn't provide a message? I think we observed that situation recently on iOS.

libnodegl/utils.c
ubitux: it's not about signs, it's about `sizeof(size_t)` > `sizeof(int)`; the format string for `size_t` is annoying to get right in a portable way.

mbouron-gpsw: Ok.

libnodegl/backends/gl/program_gl.c
mbouron-gpsw: The error we faced recently did trigger the logs. Not seeing the logs in the debug console was an issue on the user end. In this case, we won't see the error message only if ngli_numbered_lines() fails to allocate memory which is really unlikely.

cboesch-gpsw: Ah I'm mistaken with a crash I had in MESA while constructing the log message: https://gitlab.freedesktop.org/mesa/mesa/-/issues/2837

Do we still want to have no error message at all in this situation?

cboesch-gpsw: Ah my bad the string can't be `NULL` here, will fix

cboesch-gpsw: Fixed

mbouron-gpsw: Thx.

mbouron-gpsw (approved):

PR #177, opened 2021-02-01
Jbelaston/msvc support

jbelaston-gpsw:
- PR making a final point to this first iteration regarding nodegl windows build work
- Adjustments added based on `jmoguillansky/meson-windows-master` (see #152) work and `mbouron/msvc-support` adjustments.

### DIFF
--- a/.github/workflows/ci_win.yml
+++ b/.github/workflows/ci_win.yml
@@ -30,3 +30,42 @@ jobs:
         run: |
           $env:CHERE_INVOKING = 'yes'  # Preserve the current working directory
           C:\msys64\usr\bin\bash -lc "make -j$(($(nproc)+1)) TARGET_OS=MinGW-w64"
+
+  msvc:
+
+    runs-on: windows-latest
+
+    defaults:
+      run:
+        shell: wsl-bash {0}
+
+    steps:
+      - uses: Vampire/setup-wsl@v1
+        with:
+          distribution: Ubuntu-20.04
+
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          sudo apt -y update
+          sudo apt -y install build-essential
+          pip.exe install meson
+
+      - name: Restore from cache and install vcpkg
+        uses: lukka/run-vcpkg@v6
+        with:
+          vcpkgGitCommitId: 595777db2332a3442b73f9af9f656355f207aec9
+          vcpkgTriplet: x64-windows
+          vcpkgArguments: pthreads opengl-registry ffmpeg[ffmpeg,ffprobe] sdl2
+          cleanAfterBuild: false
+
+      - name: vcpkg integrate install
+        run: |
+          cd vcpkg && ./vcpkg.exe integrate install
+
+      - name: Build
+        run: |
+          make -j$(($(nproc)+1)) TARGET_OS=Windows \
+          VCPKG_DIR='D:\\a\\gopro-lib-node.gl\\gopro-lib-node.gl\\vcpkg' \
+          VCVARS64='"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"'
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@
 # under the License.
 #
 
-PREFIX ?= $(PWD)/nodegl-env
+PREFIX ?= $(PWD)/venv
 
 PYTHON_MAJOR = 3
 
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,8 @@
 # under the License.
 #
 
-PREFIX ?= $(PWD)/venv
+PREFIX          ?= venv
+PREFIX_FULLPATH  = $(PWD)/$(PREFIX)
 
 PYTHON_MAJOR = 3
 
@@ -41,11 +42,11 @@ ifneq ($(shell $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON
 $(error "Python $(PYTHON_MAJOR) not found")
 endif
 
-ACTIVATE = . $(PREFIX)/bin/activate
+ACTIVATE = . $(PREFIX_FULLPATH)/bin/activate
 
-RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX)/lib
+RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX_FULLPATH)/lib
 
-MESON_SETUP   = meson setup --prefix=$(PREFIX) --pkg-config-path=$(PREFIX)/lib/pkgconfig -Drpath=true
+MESON_SETUP   = meson setup --prefix=$(PREFIX_FULLPATH) --pkg-config-path=$(PREFIX_FULLPATH)/lib/pkgconfig -Drpath=true
 # MAKEFLAGS= is a workaround for the issue described here:
 # https://github.com/ninja-build/ninja/issues/1139#issuecomment-724061270
 MESON_COMPILE = MAKEFLAGS= meson compile
@@ -124,9 +125,9 @@ ifneq ($(TARGET_OS),MinGW-w64)
 endif
 
 pynodegl-install: pynodegl-deps-install
-	($(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
+	($(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX_FULLPATH)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
 
-pynodegl-deps-install: $(PREFIX) nodegl-install
+pynodegl-deps-install: $(PREFIX_FULLPATH) nodegl-install
 	($(ACTIVATE) && pip install -r ./pynodegl/requirements.txt)
 
 nodegl-install: nodegl-setup
@@ -135,7 +136,7 @@ nodegl-install: nodegl-setup
 nodegl-setup: sxplayer-install
 	($(ACTIVATE) && $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir/libnodegl)
 
-sxplayer-install: external-download $(PREFIX)
+sxplayer-install: external-download $(PREFIX_FULLPATH)
 	($(ACTIVATE) && $(MESON_SETUP) external/sxplayer builddir/sxplayer && $(MESON_COMPILE) -C builddir/sxplayer && $(MESON_INSTALL) -C builddir/sxplayer)
 
 external-download:
@@ -145,7 +146,7 @@ external-download:
 # We do not pull meson from pip on Windows for the same reasons we don't pull
 # Pillow and PySide2. We require the users to have it on their system.
 #
-$(PREFIX):
+$(PREFIX_FULLPATH):
 ifeq ($(TARGET_OS),MinGW-w64)
 	$(PYTHON) -m venv --system-site-packages $@
 else
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ ifneq ($(shell $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON
 $(error "Python $(PYTHON_MAJOR) not found")
 endif
 
-ACTIVATE = $(PREFIX)/bin/activate
+ACTIVATE = . $(PREFIX)/bin/activate
 
 RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX)/lib
 
@@ -89,14 +89,14 @@ all: ngl-tools-install pynodegl-utils-install
 	@echo "    Install completed."
 	@echo
 	@echo "    You can now enter the venv with:"
-	@echo "        . $(ACTIVATE)"
+	@echo "        $(ACTIVATE)"
 	@echo
 
 ngl-tools-install: nodegl-install
-	(. $(ACTIVATE) && $(MESON_SETUP) ngl-tools builddir/ngl-tools && $(MESON_COMPILE) -C builddir/ngl-tools && $(MESON_INSTALL) -C builddir/ngl-tools)
+	($(ACTIVATE) && $(MESON_SETUP) ngl-tools builddir/ngl-tools && $(MESON_COMPILE) -C builddir/ngl-tools && $(MESON_INSTALL) -C builddir/ngl-tools)
 
 pynodegl-utils-install: pynodegl-utils-deps-install
-	(. $(ACTIVATE) && pip -v install -e ./pynodegl-utils)
+	($(ACTIVATE) && pip -v install -e ./pynodegl-utils)
 
 #
 # pynodegl-install is in dependency to prevent from trying to install pynodegl
@@ -120,23 +120,23 @@ pynodegl-utils-install: pynodegl-utils-deps-install
 #
 pynodegl-utils-deps-install: pynodegl-install
 ifneq ($(TARGET_OS),MinGW-w64)
-	(. $(ACTIVATE) && pip install -r ./pynodegl-utils/requirements.txt)
+	($(ACTIVATE) && pip install -r ./pynodegl-utils/requirements.txt)
 endif
 
 pynodegl-install: pynodegl-deps-install
-	(. $(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
+	($(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
 
 pynodegl-deps-install: $(PREFIX) nodegl-install
-	(. $(ACTIVATE) && pip install -r ./pynodegl/requirements.txt)
+	($(ACTIVATE) && pip install -r ./pynodegl/requirements.txt)
 
 nodegl-install: nodegl-setup
-	(. $(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl && $(MESON_INSTALL) -C builddir/libnodegl)
+	($(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl && $(MESON_INSTALL) -C builddir/libnodegl)
 
 nodegl-setup: sxplayer-install
-	(. $(ACTIVATE) && $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir/libnodegl)
+	($(ACTIVATE) && $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir/libnodegl)
 
 sxplayer-install: external-download $(PREFIX)
-	(. $(ACTIVATE) && $(MESON_SETUP) external/sxplayer builddir/sxplayer && $(MESON_COMPILE) -C builddir/sxplayer && $(MESON_INSTALL) -C builddir/sxplayer)
+	($(ACTIVATE) && $(MESON_SETUP) external/sxplayer builddir/sxplayer && $(MESON_COMPILE) -C builddir/sxplayer && $(MESON_INSTALL) -C builddir/sxplayer)
 
 external-download:
 	$(MAKE) -C external
@@ -150,20 +150,20 @@ ifeq ($(TARGET_OS),MinGW-w64)
 	$(PYTHON) -m venv --system-site-packages $(PREFIX)
 else
 	$(PYTHON) -m venv $(PREFIX)
-	(. $(ACTIVATE) && pip install meson ninja)
+	($(ACTIVATE) && pip install meson ninja)
 endif
 
 tests: nodegl-tests tests-setup
-	(. $(ACTIVATE) && meson test $(MESON_TESTS_SUITE_OPTS) -C builddir/tests)
+	($(ACTIVATE) && meson test $(MESON_TESTS_SUITE_OPTS) -C builddir/tests)
 
 tests-setup: ngl-tools-install pynodegl-utils-install
-	(. $(ACTIVATE) && $(MESON_SETUP) builddir/tests tests)
+	($(ACTIVATE) && $(MESON_SETUP) builddir/tests tests)
 
 nodegl-tests: nodegl-install
-	(. $(ACTIVATE) && meson test -C builddir/libnodegl)
+	($(ACTIVATE) && meson test -C builddir/libnodegl)
 
 nodegl-%: nodegl-setup
-	(. $(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl $(subst nodegl-,,$@))
+	($(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl $(subst nodegl-,,$@))
 
 clean_py:
 	$(RM) pynodegl/nodes_def.pyx
@@ -186,9 +186,9 @@ clean: clean_py
 # We don't use `meson coverage` here because of
 # https://github.com/mesonbuild/meson/issues/7895
 coverage-html:
-	(. $(ACTIVATE) && ninja -C builddir/libnodegl coverage-html)
+	($(ACTIVATE) && ninja -C builddir/libnodegl coverage-html)
 coverage-xml:
-	(. $(ACTIVATE) && ninja -C builddir/libnodegl coverage-xml)
+	($(ACTIVATE) && ninja -C builddir/libnodegl coverage-xml)
 
 .PHONY: all
 .PHONY: ngl-tools-install
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,11 @@
 #
 
 PREFIX          ?= venv
-PREFIX_FULLPATH  = $(PWD)/$(PREFIX)
+ifeq ($(TARGET_OS),Windows)
+PREFIX_FULLPATH = $(shell wslpath -wa .)\$(PREFIX)
+else
+PREFIX_FULLPATH = $(PWD)/$(PREFIX)
+endif
 
 PYTHON_MAJOR = 3
 
@@ -29,7 +33,11 @@ PYTHON_MAJOR = 3
 #
 DEBUG      ?= no
 COVERAGE   ?= no
+ifeq ($(TARGET_OS),Windows)
+PYTHON     ?= python.exe
+else
 PYTHON     ?= python$(if $(shell which python$(PYTHON_MAJOR) 2> /dev/null),$(PYTHON_MAJOR),)
+endif
 export TARGET_OS ?= $(shell uname -s)
 
 DEBUG_GL    ?= no
@@ -38,18 +46,50 @@ DEBUG_SCENE ?= no
 TESTS_SUITE ?=
 V           ?=
 
-ifneq ($(shell $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON_MAJOR))
+$(info PYTHON: $(PYTHON))
+$(info PREFIX: $(PREFIX))
+$(info PREFIX_FULLPATH: $(PREFIX_FULLPATH))
+
+ifeq ($(TARGET_OS),Windows)
+# Initialize VCVARS64 and VCPKG_DIR to a default value
+# Note: the user should override this environment variable if needed
+VCVARS64 ?= "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
+VCPKG_DIR ?= C:\\vcpkg
+PKG_CONF_DIR = external\\pkgconf\\build
+CMD = PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG="$(PREFIX_FULLPATH)\\Scripts\\pkg-config.exe" PKG_CONFIG_PATH="$(VCPKG_DIR)\\installed\\x64-windows\\lib\\pkgconfig" WSLENV=PKG_CONFIG/w:PKG_CONFIG_PATH/w:PKG_CONFIG_ALLOW_SYSTEM_LIBS/w:PKG_CONFIG_ALLOW_SYSTEM_CFLAGS/w cmd.exe /C
+else
+CMD =
+endif
+
+ifneq ($(shell $(CMD) $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON_MAJOR))
 $(error "Python $(PYTHON_MAJOR) not found")
 endif
 
+ifeq ($(TARGET_OS),Windows)
+ACTIVATE = $(CMD) $(VCVARS64) \&\& "$(PREFIX_FULLPATH)\\Scripts\\activate.bat"
+else
 ACTIVATE = . $(PREFIX_FULLPATH)/bin/activate
+endif
 
 RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX_FULLPATH)/lib
 
-MESON_SETUP   = meson setup --prefix=$(PREFIX_FULLPATH) --pkg-config-path=$(PREFIX_FULLPATH)/lib/pkgconfig -Drpath=true
-# MAKEFLAGS= is a workaround for the issue described here:
+ifeq ($(TARGET_OS),Windows)
+MESON_SETUP_PARAMS  = \
+    --prefix="$(PREFIX_FULLPATH)" --bindir="$(PREFIX_FULLPATH)\\Scripts" --includedir="$(PREFIX_FULLPATH)\\Include" \
+    --libdir="$(PREFIX_FULLPATH)\\Lib" --pkg-config-path="$(VCPKG_DIR)\\installed\x64-windows\\lib\\pkgconfig;$(PREFIX_FULLPATH)\\Lib\\pkgconfig" -Drpath=true
+MESON_SETUP         = meson setup --backend vs $(MESON_SETUP_PARAMS)
+MESON_SETUP_NINJA   = meson setup --backend ninja $(MESON_SETUP_PARAMS)
+else
+MESON_SETUP         = meson setup --prefix=$(PREFIX_FULLPATH) --pkg-config-path=$(PREFIX_FULLPATH)/lib/pkgconfig -Drpath=true
+endif
+# MAKEFLAGS= is a workaround (not working on Windows due to incompatible Make
+# syntax) for the issue described here:
 # https://github.com/ninja-build/ninja/issues/1139#issuecomment-724061270
+ifeq ($(TARGET_OS),Windows)
+MESON_COMPILE = meson compile
+else
 MESON_COMPILE = MAKEFLAGS= meson compile
+endif
 MESON_INSTALL = meson install
 ifeq ($(COVERAGE),yes)
 MESON_SETUP += -Db_coverage=true
@@ -90,14 +130,32 @@ all: ngl-tools-install pynodegl-utils-install
 	@echo "    Install completed."
 	@echo
 	@echo "    You can now enter the venv with:"
+ifeq ($(TARGET_OS),Windows)
+	@echo "        (via Windows Command Prompt)"
+	@echo "            cmd.exe"
+	@echo "            $(PREFIX)\\\Scripts\\\activate.bat"
+	@echo "        (via Windows PowerShell)"
+	@echo "            powershell.exe"
+	@echo "            $(PREFIX)\\\Scripts\\\Activate.ps1"
+else
 	@echo "        $(ACTIVATE)"
+endif
 	@echo
 
 ngl-tools-install: nodegl-install
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& $(MESON_SETUP) ngl-tools builddir\\ngl-tools \&\& $(MESON_COMPILE) -C builddir\\ngl-tools \&\& $(MESON_INSTALL) -C builddir\\ngl-tools)
+	$(CMD) xcopy /Y builddir\\ngl-tools\\*.dll "$(PREFIX_FULLPATH)\\Scripts\\."
+else
 	($(ACTIVATE) && $(MESON_SETUP) ngl-tools builddir/ngl-tools && $(MESON_COMPILE) -C builddir/ngl-tools && $(MESON_INSTALL) -C builddir/ngl-tools)
+endif
 
 pynodegl-utils-install: pynodegl-utils-deps-install
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& pip -v install -e pynodegl-utils)
+else
 	($(ACTIVATE) && pip -v install -e ./pynodegl-utils)
+endif
 
 #
 # pynodegl-install is in dependency to prevent from trying to install pynodegl
@@ -120,34 +178,67 @@ pynodegl-utils-install: pynodegl-utils-deps-install
 # decorator and other related utils.
 #
 pynodegl-utils-deps-install: pynodegl-install
-ifneq ($(TARGET_OS),MinGW-w64)
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& pip install -r pynodegl-utils\\requirements.txt)
+else ifneq ($(TARGET_OS),MinGW-w64)
 	($(ACTIVATE) && pip install -r ./pynodegl-utils/requirements.txt)
 endif
 
 pynodegl-install: pynodegl-deps-install
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& pip -v install -e .\\pynodegl)
+	$(CMD) xcopy /Y builddir\\sxplayer\\*.dll pynodegl\\.
+	$(CMD) xcopy /Y /C builddir\\libnodegl\\*.dll pynodegl\\.
+else
 	($(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX_FULLPATH)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
+endif
 
 pynodegl-deps-install: $(PREFIX) nodegl-install
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& pip install -r pynodegl\\requirements.txt)
+else
 	($(ACTIVATE) && pip install -r ./pynodegl/requirements.txt)
+endif
 
 nodegl-install: nodegl-setup
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& $(MESON_COMPILE) -C builddir\\libnodegl \&\& $(MESON_INSTALL) -C builddir\\libnodegl)
+else
 	($(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl && $(MESON_INSTALL) -C builddir/libnodegl)
+endif
 
 nodegl-setup: sxplayer-install
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir\\libnodegl)
+else
 	($(ACTIVATE) && $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir/libnodegl)
+endif
 
-sxplayer-install: external-download $(PREFIX)
+pkg-config-install: external-download $(PREFIX)
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& $(MESON_SETUP) -Dtests=false external\\pkgconf builddir\\pkgconf \&\& $(MESON_COMPILE) -C builddir\\pkgconf \&\& $(MESON_INSTALL) -C builddir\\pkgconf)
+	($(CMD) copy "$(PREFIX_FULLPATH)\\Scripts\\pkgconf.exe" "$(PREFIX_FULLPATH)\\Scripts\\pkg-config.exe")
+endif
+
+sxplayer-install: external-download pkg-config-install $(PREFIX)
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& $(MESON_SETUP) external\\sxplayer builddir\\sxplayer \&\& $(MESON_COMPILE) -C builddir\\sxplayer \&\& $(MESON_INSTALL) -C builddir\\sxplayer)
+else
 	($(ACTIVATE) && $(MESON_SETUP) external/sxplayer builddir/sxplayer && $(MESON_COMPILE) -C builddir/sxplayer && $(MESON_INSTALL) -C builddir/sxplayer)
+endif
 
 external-download:
 	$(MAKE) -C external
 
 #
-# We do not pull meson from pip on Windows for the same reasons we don't pull
+# We do not pull meson from pip on MinGW for the same reasons we don't pull
 # Pillow and PySide2. We require the users to have it on their system.
 #
 $(PREFIX):
-ifeq ($(TARGET_OS),MinGW-w64)
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(PYTHON) -m venv "$@")
+	($(ACTIVATE) \&\& pip install meson ninja)
+else ifeq ($(TARGET_OS),MinGW-w64)
 	$(PYTHON) -m venv --system-site-packages $@
 else
 	$(PYTHON) -m venv $@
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -19,36 +19,74 @@
...
 TESTS_SUITE ?=
 V           ?=
+NODEGL_DEFAULT_LIBRARY ?= shared
 
-ifneq ($(shell $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON_MAJOR))
-$(error "Python $(PYTHON_MAJOR) not found")
+ifeq ($(TARGET_OS), Windows)
+# Initialize VCVARS64 to a default value
+# Note: the user should override this environment variable to point to the correct path containing vcvars64.bat
+VCVARS64 ?= "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
```
jmoguillansky-gpsw, line 54: TODO: we can set this variable automatically but this can be a future improvement (future PR)

Makefile
```diff
@@ -119,51 +176,104 @@ pynodegl-utils-install: pynodegl-utils-deps-install
...
+endif
 
 external-download:
 	$(MAKE) -C external
 
-#
-# We do not pull meson from pip on Windows for the same reasons we don't pull
-# Pillow and PySide2. We require the users to have it on their system.
-#
 $(PREFIX):
```
jmoguillansky-gpsw, line 235: TODO: if this target fails, make will think it passed because the folder is already created / existing.  Also, what if the user chooses /usr/local as the PREFIX variable... Make will think the target already passed when in fact the install didn't finish yet.  This can be a future improvement / PR

Makefile
```diff
@@ -119,51 +176,104 @@ pynodegl-utils-install: pynodegl-utils-deps-install
...
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& meson test -C builddir\\libnodegl)
+else
+	($(CMD) $(ACTIVATE) && meson test -C builddir/libnodegl)
+endif
 
 nodegl-%: nodegl-setup
-	(. $(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl $(subst nodegl-,,$@))
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& $(MESON_COMPILE) -C builddir/libnodegl $(subst nodegl-,,$@))
```
jmoguillansky-gpsw, line 273: TODO: set builddir as a user param (future PR, merge from jmoguillansky/muti-backend)

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,40 @@ jobs:
         run: |
           $env:CHERE_INVOKING = 'yes'  # Preserve the current working directory
           C:\msys64\usr\bin\bash -lc "make -j$(($(nproc)+1)) TARGET_OS=MinGW-w64"
+
+  win-msvc:
```
mrobertseidowsky-gpsw, line 34: `msvc` as job name is sufficient IMHO

jmoguillansky-gpsw:
Tested locally: unit tests working

Summary of Failures:
200/215 ngl-tests:opengl+texture / cubemap_from_mrt_msaa        FAIL           1.02s (exit status 1)
200/215 ngl-tests:opengl+texture / cubemap_from_mrt_msaa        FAIL           1.02s (exit status 1)
Ok:                 214
Expected Fail:      0
Fail:               1
Unexpected Pass:    0
Skipped:            0
Timeout:            0

Makefile
jbelaston-gpsw: yep I also put it int the commit message... A classical way consists to use `vswhere.exe`

external/Makefile
```diff
@@ -24,7 +24,18 @@ CURL ?= curl
...
 SXPLAYER_VERSION ?= 9.7.0
 
-all: sxplayer
+all: sxplayer pkg-config
+
+pkgconf:
+ifeq ($(TARGET_OS), Windows)
+	$(CURL) -L https://codeload.github.com/pkgconf/pkgconf/zip/78a77dd497c4ba057613e6e93aebe89824fa71f8 -o pkgconf.zip
+	unzip pkgconf.zip && mv pkgconf-* pkgconf
+#	cd pkgconf && patch -p1 < ../patches/pkgconf/0001-args-patch-command-line-params.patch
```
mbouron, line 33: This comment can be removed.

Makefile
```diff
@@ -119,51 +176,104 @@ pynodegl-utils-install: pynodegl-utils-deps-install
...
+	($(CMD) $(ACTIVATE) && pip install -r ./pynodegl-utils/requirements.txt)
 endif
 
 pynodegl-install: pynodegl-deps-install
-	(. $(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& pip -v install -e .\\pynodegl)
+	# Install DLLs / Libs
+	$(CMD) xcopy /Y builddir\\sxplayer\\*.dll pynodegl\\.
+	$(CMD) xcopy /Y /C builddir\\libnodegl\\*.dll pynodegl\\.
```
mbouron, line 190: It seems that the 3 lines are not required for the python binding (and can be removed).

jmoguillansky-gpsw:
one more suggestion (can be a future PR):

https://github.com/gopro/gopro-lib-node.gl/blob/60c28a525f725d941f3ddc96c4b034e9dba54943/Makefile#L96

for debug mode, I suggest to set default configuration to debug instead of debugoptimized.  
It makes debugging easier.

jmoguillansky-gpsw:
For future PR:
we can consider using meson "subprojects" for internal and 3rd party modules
https://github.com/gopro/gopro-lib-node.gl/blob/60c28a525f725d941f3ddc96c4b034e9dba54943/Makefile#L225

One benefit is that all the projects will show up in the same IDE "solution file" (or xcode project file), etc and can provide a better IDE experience.  
We can have a top-level meson.build file and reference nodegl, ngl-tools, etc as subprojects

Makefile
```diff
@@ -19,36 +19,74 @@
...
 #
 DEBUG      ?= no
 COVERAGE   ?= no
+export TARGET_OS  ?= $(shell uname -s)
+
+ifeq ($(TARGET_OS),Windows)
+PYTHON     ?= python.exe
+PREFIX     ?= nodegl-env
+W_PWD = $(shell wslpath -wa .)
+W_PREFIX ?= $(W_PWD)\$(PREFIX)
```
cboesch-gpsw, line 35: style: please keep the alignment

libnodegl/build_scripts/copyfile.py
```diff
@@ -0,0 +1,16 @@
...
+import os
+import sys
+import shutil
+
+if len(sys.argv) < 2:
+    exit(1)
+
+dst = sys.argv[-1]
+if len(sys.argv) > 2:
+    if not os.path.isdir(dst):
```
mrobertseidowsky-gpsw, line 12: on one line with `&` ?

libnodegl/build_scripts/copyfile.py
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import shutil
+
```
mrobertseidowsky-gpsw, line 6: can you add an helper about what type of parameter is expected ? something like "please provide source and destination repositories" in case it's missing ?

Makefile
```diff
@@ -19,36 +19,74 @@
...
+ACTIVATE = $(VCVARS64) \&\& $(PREFIX)\\Scripts\\activate.bat
+# Initialize VCPKG_DIR to a default value
+# Note: the user should override this environment variable to point to the correct path containing vcpkg
+VCPKG_DIR ?= C:\\vcpkg
+PKG_CONF_DIR = external\\pkgconf\\build
+# Set PKG_CONFIG and PKG_CONFIG_PATH environment variables when invoking command shell
+CMD = PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG="$(W_PREFIX)\\Scripts\\pkg-config.exe" PKG_CONFIG_PATH="$(VCPKG_DIR)\\installed\\x64-windows\\lib\\pkgconfig" WSLENV=PKG_CONFIG/w:PKG_CONFIG_PATH/w:PKG_CONFIG_ALLOW_SYSTEM_LIBS/w:PKG_CONFIG_ALLOW_SYSTEM_CFLAGS/w cmd.exe /C
+else
+ACTIVATE = . $(PREFIX)/bin/activate
+CMD =
```
cboesch-gpsw, line 64: Don't you want `CMD = .` and `ACTIVATE` unchanged? This change could be isolated in its commit as it applies on master and while reduce the number of negative lines from this diff significantly

Makefile
```diff
@@ -19,36 +19,74 @@
...
+  --prefix="$(W_PREFIX)" --bindir="$(W_PREFIX)\\Scripts" --includedir="$(W_PREFIX)\\Include" \
+  --libdir="$(W_PREFIX)\\Lib" --pkg-config-path="$(VCPKG_DIR)\\installed\x64-windows\\lib\\pkgconfig;$(W_PREFIX)\\Lib\\pkgconfig" -Drpath=true
+else
 MESON_SETUP   = meson setup --prefix=$(PREFIX) --pkg-config-path=$(PREFIX)/lib/pkgconfig -Drpath=true
+endif
 # MAKEFLAGS= is a workaround for the issue described here:
 # https://github.com/ninja-build/ninja/issues/1139#issuecomment-724061270
+ifeq ($(TARGET_OS),Windows)
+MESON_COMPILE = meson compile
+else
```
cboesch-gpsw, line 87: The default with `MAKEFLAGS` doesn't work?

Makefile
```diff
@@ -19,36 +19,74 @@
 # under the License.
```
mrobertseidowsky-gpsw, line 19: spelling mistake in git commit messsage:
=> `Windows Command Promt ` => `Prompt`

mrobertseidowsky-gpsw: btw, we have to signed our commit ? I refered to `Signed-off-by: Matthieu Bouron <matthieu.bouron@gmail.com>
`

Makefile
```diff
@@ -89,14 +127,33 @@ all: ngl-tools-install pynodegl-utils-install
...
+	@echo "        $(ACTIVATE)"
+endif
 	@echo
 
 ngl-tools-install: nodegl-install
-	(. $(ACTIVATE) && $(MESON_SETUP) ngl-tools builddir/ngl-tools && $(MESON_COMPILE) -C builddir/ngl-tools && $(MESON_INSTALL) -C builddir/ngl-tools)
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& $(MESON_SETUP) ngl-tools builddir\\ngl-tools \&\& $(MESON_COMPILE) -C builddir\\ngl-tools \&\& $(MESON_INSTALL) -C builddir\\ngl-tools)
+	# Install DLLs / Libs
+	$(CMD) xcopy /Y builddir\\ngl-tools\\*.dll $(PREFIX)\\Scripts\\.
```
cboesch-gpsw, line 146: what dll is there in ngl-tools?

external/Makefile
```diff
@@ -24,7 +24,18 @@ CURL ?= curl
 
 SXPLAYER_VERSION ?= 9.7.0
 
-all: sxplayer
+all: sxplayer pkg-config
+
+pkgconf:
+ifeq ($(TARGET_OS), Windows)
+	$(CURL) -L https://codeload.github.com/pkgconf/pkgconf/zip/78a77dd497c4ba057613e6e93aebe89824fa71f8 -o pkgconf.zip
```
cboesch-gpsw, line 31: Why not use https://github.com/pkgconf/pkgconf/archive/pkgconf-1.7.3.tar.gz ?

So that:
1. we know what version this is
2. we don't need `zip` installed (like in the CI)

external/Makefile
```diff
@@ -24,7 +24,18 @@ CURL ?= curl
...
+	unzip pkgconf.zip && mv pkgconf-* pkgconf
```
cboesch-gpsw, line 32: Please chain the dependencies like with `sxplayer`: `pkgconf` → `pkgconf-$(PKGCONF-VERSION)` → `pkgconf-$(PKGCONF-VERSION)`

external/Makefile
```diff
@@ -24,7 +24,18 @@ CURL ?= curl
...
+#	cd pkgconf && patch -p1 < ../patches/pkgconf/0001-args-patch-command-line-params.patch
+	rm -rf pkgconf/.git
```
cboesch-gpsw, line 34: you don't need to do this if you pick from the tarball above

libnodegl/meson.build
```diff
@@ -405,7 +405,7 @@ pkg.generate(
 # Specifications
 #
 
-cp = find_program('cp')
+copyfile = find_program('./build_scripts/copyfile.py')
```
cboesch-gpsw, line 408: Do you need the `./`? I don't think the current directory is what you think it is here

doc/howto/installation.md
```diff
@@ -93,3 +141,12 @@ pip install ./pynodegl
...
 ```
+
+## Known Limitations
+
+- [Windows only] Even if it should work using Cygwin environment, no test has been done.
+- [Windows only] Even if both WSL versions have been tested, we encourage to
+use version 1 which seems for now faster and stable in our use case. Indeed
+we are accessing to executable files mounted from Windows, which is a drawback
+for WSL 2 linux kernel needing files entirely stored in Linux file system to
+be efficent (more info [here](https://docs.microsoft.com/en-us/windows/wsl/compare-versions#exceptions-for-using-wsl-1-rather-than-wsl-2))
```
cboesch-gpsw, line 152: You can defer a link in markdown. Also, a link with "here" is not recommended because out of context it means nothing. Instead, you can do:

```markdown
for [WSL 2 linux kernel needing files entirely stored in Linux file system to
be efficient][wsl1-vs-wsl2]

[wsl1-vs-wsl2]: https://docs.microsoft.com/en-us/windows/wsl/compare-versions#exceptions-for-using-wsl-1-rather-than-wsl-2
```

Also mind the "efficent" typo you have

Makefile
mbouron-gpsw: We tend to sign-off a commit when we do modifications to it (and when we are not the original author).

.github/workflows/ci_win.yml
jbelaston-gpsw: :+1:  done

external/Makefile
jbelaston-gpsw: Done

Makefile
jbelaston-gpsw: yep. you're right.
Done

libnodegl/build_scripts/copyfile.py
jbelaston-gpsw: Yep. Now, using `and`, we have:
```
if len(sys.argv) > 2 and not os.path.isdir(dst):
    exit(1)
```

libnodegl/build_scripts/copyfile.py
jbelaston-gpsw: Done :+1:

libnodegl/meson.build
jbelaston-gpsw: Oh yep, you're right. It's not needed.
Done

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,40 @@ jobs:
...
+          vcpkgArguments: pthreads opengl-registry ffmpeg[ffmpeg,ffprobe] sdl2
+          cleanAfterBuild: false
+
+      - name: vcpkg integrate install
+        run: |
+          cd vcpkg && ./vcpkg.exe integrate install
+
+      - name: Build
+        run: |
+          make -j$(($(nproc)+1)) TARGET_OS=Windows PKG_CONFIG='D:\\a\\gopro-lib-node.gl\\gopro-lib-node.gl\\pkg-config.exe' VCPKG_DIR='D:\\a\\gopro-lib-node.gl\\gopro-lib-node.gl\\vcpkg'
```
mrobertseidowsky-gpsw, line 69: you can split the line with  `&& ^` at the end ?

Makefile
jmoguillansky-gpsw: Personally I don't think so.  Dot is an alias for source which is used to set variables in the parent shell.  Its used with activate but not with other commands.  But feel free to modify it however you like

Makefile
jmoguillansky-gpsw: I see that this is a workaround to enable multithreaded meson compile?  Probably it should work too and maybe will help since we're using wsl

doc/howto/installation.md
```diff
@@ -1,62 +1,110 @@
...
+    ```
+- Enter the virtual environment with:
+    ```shell
+    . /home/user/node.gl/nodegl-env/bin/activate
+    ```
+
+## Quick user installation on Windows (MSVC toolchain)
+
+- Install **[Python](https://www.python.org/downloads/windows/)** 3.x using the Windows Installer
+- Install **[CMake](https://cmake.org/download/)** 3.x or greater using the Windows Installer
```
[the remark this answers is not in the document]
jmoguillansky-gpsw: 3rd party libs

doc/howto/installation.md
```diff
@@ -1,62 +1,110 @@
...
+- Install **[Microsoft Visual Studio](https://visualstudio.microsoft.com/downloads/)** 2019 v16.3 or greater and make sure the
+following components are included:
+    - Desktop development with C++
+    - MSVC v142 - VS 2019 C++ x64/x86 build tools
+    - Windows 10 SDK (10.0.18362.0)
+- Download **[Vcpkg](https://github.com/microsoft/vcpkg)** anywhere on your
+files system (avoid root directory since it is not recommended on Windows)
+- From Windows PowerShell (run as administrator), bootstrap Vcpkg and install
```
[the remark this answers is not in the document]
jmoguillansky-gpsw: It shouldn't require administrator priviledge (another reason we should remove vcpkg integrate install for pthreads dependency)

doc/howto/installation.md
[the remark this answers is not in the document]
cboesch-gpsw: which ones?

.github/workflows/ci_win.yml
jbelaston-gpsw: Done

doc/howto/installation.md
jbelaston-gpsw: Oh great! thanks for the tip
Done.

doc/howto/installation.md
```diff
@@ -93,3 +141,12 @@ pip install ./pynodegl
 pip install -r ./pynodegl-utils/requirements.txt
 pip install ./pynodegl-utils
 ```
+
+## Known Limitations
+
+- [Windows only] Even if it should work using Cygwin environment, no test has been done.
+- [Windows only] Even if both WSL versions have been tested, we encourage to
+use version 1 which seems for now faster and stable in our use case. Indeed
+we are accessing to executable files mounted from Windows, which is a drawback
```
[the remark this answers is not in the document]
jbelaston-gpsw: done.

doc/howto/installation.md
```diff
@@ -1,62 +1,110 @@
...
+    pacman -S mingw-w64-x86_64-pyside2-qt5
+    pacman -S mingw-w64-x86_64-meson
+    ```
+- From MinGW64, build with:
+    ```shell
+    make -j8 TARGET_OS=MinGW-w64
+    ```
+- Enter the virtual environment with:
+    ```shell
+    . /home/user/node.gl/nodegl-env/bin/activate
```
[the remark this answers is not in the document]
jbelaston-gpsw: Oh yep you're right! But that means we should remove it also in `Quick user installation on Linux and MacOS` section. 
What do you think ?

doc/howto/installation.md
```diff
@@ -1,62 +1,110 @@
...
+  - **GNU/Make**
+  - **FFmpeg** (and its libraries for compilation)
+  - **Python 3.x** (you will need the package with the build headers as well,
+    typically named with a `-devel` suffix on Debian based systems)
+  - **pip** (Python 3 version)
+  - **Graphviz**
+  - **SDL2**
+- Build with:
+    ```shell
+    make -j8
```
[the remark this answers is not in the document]
jbelaston-gpsw: Correct. Done.

doc/howto/installation.md
```diff
@@ -1,62 +1,110 @@
...
-- **GNU/Make**
-- **FFmpeg** (and its libraries for compilation)
-- **Python 3.x** (you will need the package with the build headers as well,
-  typically named with a `-devel` suffix on Debian based systems)
-- **pip** (Python 3 version)
-- **Graphviz**
-- **SDL2**
+- MacOS 10.15 or greater
+- Ubuntu 20.04 or greater
+- Arch Linux
```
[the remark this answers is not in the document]
jbelaston-gpsw: Yeah... Upon consideration, keep it as it was originally makes more sense :+1: 
I removed this part.

doc/howto/installation.md
```diff
@@ -1,62 +1,110 @@
 # Installation
 
-## Dependencies
+This page includes the system requirements as well as the steps needed for
+building and running the complete `node.gl` stack.
 
-Make sure you have installed on your system the essential list of dependencies
-for building and running the complete **node.gl** stack:
+## Recommended Operating Systems
```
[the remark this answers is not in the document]
jbelaston-gpsw: Section removed. See my next comment. :wink:

doc/howto/installation.md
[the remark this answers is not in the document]
jbelaston-gpsw: Oh ok. In fact, maybe it's not clear and I have to reword it.

Not the official recommendation from https://github.com/microsoft/vcpkg:

```
Prerequisites: 
- Windows 7 or newer
- Git
- Visual Studio 2015 Update 3 or greater with the English language pack

First, download and bootstrap vcpkg itself; it can be installed anywhere, but generally we recommend
using vcpkg as a submodule for CMake projects, and installing it globally for Visual Studio projects. 
We recommend somewhere like C:\src\vcpkg or C:\dev\vcpkg, since otherwise you may run into path
issues for some port build systems.
```

doc/howto/installation.md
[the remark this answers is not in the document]
jbelaston-gpsw: Ok so I retried a setup from scratch without installing CMake. 
Result:
- @jmoguillansky-gpsw you're right. Vcpkg requires it when installing 3rd party libs (ffmpeg)
- But it automatically manage it (download & install)
So I think it's better to remove this step from the doc (in order to not induce user to believe that nodegl directly need it, which is false).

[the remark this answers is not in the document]
jbelaston-gpsw: Done

Makefile
jbelaston-gpsw: :+1 
See https://jira.gopro.com/browse/MAGMA-1667

jbelaston-gpsw:
> one more suggestion (can be a future PR):
> 
> https://github.com/gopro/gopro-lib-node.gl/blob/60c28a525f725d941f3ddc96c4b034e9dba54943/Makefile#L96
> 
> for debug mode, I suggest to set default configuration to debug instead of debugoptimized.
> It makes debugging easier. This can also be a configurable user param

:+1: 
Request saved into https://jira.gopro.com/browse/MAGMA-1668

jbelaston-gpsw:
> nodegl-install: nodegl-setup

Definitely agree with you! Something I also talked to @mbouron-gpsw  and @cboesch-gpsw .
It's really common for windows user using Visual Studio IDE to use the solution (`*.sln`) as a super project referring all sub projects... Today, we doesn't really take advantage of `*.sln` since each project (`sxplayer`, `libnodefgl`, etc...) has one solution referring its own project... really not interesting... But yep, something for the future....

jbelaston-gpsw:
> For future PR:
> we can consider using meson "subprojects" for internal and 3rd party modules
> 
> https://github.com/gopro/gopro-lib-node.gl/blob/60c28a525f725d941f3ddc96c4b034e9dba54943/Makefile#L205
> 
> 
> https://github.com/gopro/gopro-lib-node.gl/blob/60c28a525f725d941f3ddc96c4b034e9dba54943/Makefile#L225
> 
> One benefit is that all the projects will show up in the same IDE "solution file" (or xcode project file), etc and can provide a better IDE experience.
> We can have a top-level meson.build file and reference nodegl, ngl-tools, etc as subprojects

Definitely agree with you! Something I also talked to @mbouron-gpsw and @cboesch-gpsw .
It's really common for windows user using Visual Studio IDE to use the solution (*.sln) as a super project referring all sub projects... Today, we doesn't really take advantage of *.sln since each project (sxplayer, libnodefgl, etc...) has one solution referring its own project... really not interesting... But yep, something for the future....

Makefile
jbelaston-gpsw: `SDL2.dll` and `pthreadCV3.dll`
They are both needed for ngl-player

doc/howto/installation.md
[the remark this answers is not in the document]
jmoguillansky-gpsw: ?  In microsoft's documentation, vcpkg install doesn't require admin, only vcpkg integrate install.  There's reasons to avoid integrate install:
1) modifies user's global visual studio configuration
2) sets global linker args with visual studio, not with other build systems like ninja or cmake

Also fyi: vcpkg integrate install adds the linker arg for all visual studio projects, slowing down linking for all projects....  the current vcpkg logic is not ideal


https://github.com/microsoft/vcpkg/issues/15452

external/Makefile
jbelaston-gpsw: Done

external/Makefile
jbelaston-gpsw: Done

external/Makefile
jbelaston-gpsw: Done

doc/howto/installation.md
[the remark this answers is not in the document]
jbelaston-gpsw: Done.

external/Makefile
```diff
@@ -23,8 +23,9 @@ TAR  ?= tar
 CURL ?= curl
 
 SXPLAYER_VERSION ?= 9.7.0
+PKGCONF_VERSION ?= 1.7.3
 
-all: sxplayer
+all: sxplayer pkgconf
```
ubitux, line 28: can you do something like:
```make
DEPENDENCIES = sxplayer
ifeq ($(TARGET_OS),Windows)
DEPENDENCIES += pkgconf
endif
all: $(DEPENDENCIES)
```

And remove the `if`/`endif` in the `pkgconf` rule?

Rationale: we don't want `pkgconf` on most platforms (but still want to be able to trigger the rule everywhere)

Makefile
jbelaston-gpsw: Ok so:
- I well isolated the inclusion of "." into $(ACTIVATE) in a separate commit.
- I still let CMD set  to empty in the case of Linux/MacOS for now. The best would be to fully remove it Linux/MacOS case.

Makefile
mbouron-gpsw: Not sure i was clear enough, I was talking about the 3 lines bellow my comment. On my setup, I only need to copy the dll dependencies to the pynodegl folder.

jmoguillansky-gpsw:
For debugging the dll loading issue I recommend using windows process monitor from sysinternals.  We may be able to set PATH env variable .  Don't we want these dlls in nodegl-env/Scripts folder?  They may be used outside pynodegl etc.

Makefile
jmoguillansky-gpsw: For debugging the dll loading issue I recommend using windows process monitor from sysinternals.  We may be able to set PATH env variable to modify dll loading path) .  Don't we want these dlls in $(prefix)/Scripts folder (it should be prefix/bin but python's venv naming conventions are strange on Windows)?  They may be used outside pynodegl etc.

Makefile
jbelaston-gpsw: It doesn't work: I get following error:
`'MAKEFLAGS' is not recognized as an internal or external command, operable program or batch file.`

Makefile
mbouron-gpsw: If they are used outside pynodegl ok but I don't think they belong in the post install of pynodegl.

jbelaston-gpsw: Ok so, just to make it clear, we are talking about these copies:
```
pynodegl-install: pynodegl-deps-install
ifeq ($(TARGET_OS),Windows)
	...
	$(CMD) xcopy /Y builddir\\sxplayer\\*.dll $(PREFIX)\\Scripts\\.
	$(CMD) xcopy /Y /C builddir\\libnodegl\\*.dll $(PREFIX)\\Scripts\\.
	$(CMD) xcopy /Y /C builddir\\libnodegl\\*.lib $(PREFIX)\\Lib\\.
	...
```
So, yep for my point of view too, definitely not required our work which consists today to manage the build/binding (and not the integration ). I also checked on my setup and I confirmed than without these copies, we are good (not needed for the python biding).
Lines removed! :+1:

external/Makefile
jbelaston-gpsw: Oh yeah! I like it :+1: Thanks for tip.
Done (`if`/`endif` also removed :+1: )

doc/howto/installation.md
```diff
@@ -1,62 +1,102 @@
...
+    .\vcpkg.exe install pthreads:x64-windows opengl-registry:x64-windows ffmpeg[ffmpeg,ffprobe]:x64-windows sdl2:x64-windows
+    .\vcpkg.exe integrate install
+    ```
+- Enable **[WSL (Windows Subsystem for Linux)](https://docs.microsoft.com/en-us/windows/wsl/install-win10)**
+on your system and install Ubuntu (recommend 20.04) from Microsoft Store.
+- In WSL, install other needed dependencies as 
+following:
+    ```shell
+    sudo apt -y update
+    sudo apt -y install build-essential zip
```
cboesch-gpsw, line 77: I think you don't need `zip` anymore

doc/howto/installation.md
```diff
@@ -1,62 +1,102 @@
...
+    ```
+- Add `C:\dev\vcpkg\installed\x64-windows\tools\ffmpeg` path to your windows system %PATH% environment variable (ffmpeg
+and ffprobe binaries must be available in order to run the tests)
```
cboesch-gpsw, line 80: inline backquote on `%PATH%`, `ffmpeg` and `ffprobe` please

doc/howto/installation.md
```diff
@@ -93,3 +133,14 @@ pip install ./pynodegl
 pip install -r ./pynodegl-utils/requirements.txt
 pip install ./pynodegl-utils
 ```
+
+## Known Limitations
+
+- [Windows only] Even if it should work using Cygwin environment, no test has been done.
+- [Windows only] Even if both WSL versions have been tested, we encourage to
```
cboesch-gpsw, line 140: You should avoid the use of special characters like `[ ]` since it's used for markup. You could add a `### On Windows` sub-section eventually.

Or you could just add a `### Known limitations` in the Windows sub-section.

doc/howto/installation.md
```diff
@@ -1,62 +1,102 @@
...
+    ```
+
+## Quick user installation on Windows (MSVC toolchain)
+
+- Install **[Python](https://www.python.org/downloads/windows/)** 3.x using the Windows Installer
+- Install **[Microsoft Visual Studio](https://visualstudio.microsoft.com/downloads/)** 2019 v16.3 or greater and make sure the
+following components are included:
+    - Desktop development with C++
+    - MSVC v142 - VS 2019 C++ x64/x86 build tools
+    - Windows 10 SDK (10.0.18362.0)
```
cboesch-gpsw, line 61: Are we going to have to update those regularly? Isn't it enough to say "MSVC - VS 2019 C++ x64/x86 build tools" and "Windows 10 SDK"?

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,42 @@ jobs:
...
+      - uses: Vampire/setup-wsl@v1
+        with:
+          distribution: Ubuntu-20.04
+
+      - uses: actions/checkout@v2
+
+      - name: Install dependencies
+        run: |
+          sudo apt -y update
+          sudo apt -y install build-essential zip
```
cboesch-gpsw, line 52: `zip` not needed anymore?

.github/workflows/ci_win.yml
```diff
@@ -30,3 +30,42 @@ jobs:
...
+          pip.exe install meson
+
+      - name: Restore from cache and install vcpkg
+        uses: lukka/run-vcpkg@v6
+        with:
+          vcpkgGitCommitId: 595777db2332a3442b73f9af9f656355f207aec9
```
cboesch-gpsw, line 58: Do we need this?

Makefile
```diff
@@ -19,36 +19,74 @@
...
 DEBUG_GL    ?= no
 DEBUG_MEM   ?= no
 DEBUG_SCENE ?= no
 TESTS_SUITE ?=
 V           ?=
+NODEGL_DEFAULT_LIBRARY ?= shared
 
-ifneq ($(shell $(PYTHON) -c "import sys;print(sys.version_info.major)"),$(PYTHON_MAJOR))
-$(error "Python $(PYTHON_MAJOR) not found")
+ifeq ($(TARGET_OS), Windows)
```
cboesch-gpsw, line 51: We don't have extra spaces after `,` in other conditions, please keep it that way

Makefile
```diff
@@ -19,36 +19,74 @@
...
 #
 DEBUG      ?= no
 COVERAGE   ?= no
+export TARGET_OS  ?= $(shell uname -s)
+
+ifeq ($(TARGET_OS),Windows)
+PYTHON     ?= python.exe
+PREFIX     ?= nodegl-env
+W_PWD       = $(shell wslpath -wa .)
+W_PREFIX   ?= $(W_PWD)\$(PREFIX)
```
cboesch-gpsw, line 35: what's the difference between `PREFIX` and `W_PREFIX`? It looks like `PREFIX` is absolute except on windows where it's relative and `W_PREFIX` is only on windows but is absolute...? This is extremely confusing and error prone IMO, it's going to be a real pain to maintain the correct behaviour.

Makefile
```diff
@@ -19,36 +19,74 @@
...
+$(info PYTHON: $(PYTHON))
+$(info PREFIX: $(PREFIX))
+$(info W_PREFIX: $(W_PREFIX))
```
cboesch-gpsw, line 38: There is no reason to print those only under windows scope (let's try to have a maximum of common code to reduce the risk of breakage). For `W_PREFIX`, maybe we should have `REL_PREFIX` and `PREFIX` (always absolute) in both systems.

Makefile
cboesch-gpsw: OK, please add a comment then

external/.gitignore
```diff
@@ -1 +1,2 @@
 sxplayer*
+pkgconf*
```
cboesch-gpsw, line 2: alphabetical order please

external/Makefile
```diff
@@ -23,8 +23,14 @@ TAR  ?= tar
 CURL ?= curl
 
 SXPLAYER_VERSION ?= 9.7.0
+PKGCONF_VERSION ?= pkgconf-1.7.3
```
cboesch-gpsw, line 26: version should be `1.7.3` only

Makefile
cboesch-gpsw: I'd rather go in the direction where `PREFIX` is not customizable by the user: this is not meant to be a full-featured package but rather a standalone in-place user setup. If you want to have it on your system or cleanly installed, you should rely on proper packages (which will never use this root Makefile but rather redo what it does with proper integration of the deps etc).

external/Makefile
jbelaston-gpsw: (seen with @mbouron)
That was also my first intention but I got an error since, contrary to sxplayer archives which are named `sxplayer-XXX`, pkgconfig archives are named `pkgconfig-pkgconfig-XXX ` :smile:
Why... ? Because...
Tag Version is "X.X.X" for sxplayer and "pkgconfig-X.X.X" for pkgconfig...

ubitux: Well then maybe use https://distfiles.dereferenced.org/pkgconf/ which seems to be the recommended URL: https://github.com/pkgconf/pkgconf#release-tarballs

Makefile
jbelaston-gpsw: Done

doc/howto/installation.md
```diff
@@ -1,62 +1,102 @@
...
-- **GCC** or **Clang**
-- **GNU/Make**
-- **FFmpeg** (and its libraries for compilation)
-- **Python 3.x** (you will need the package with the build headers as well,
-  typically named with a `-devel` suffix on Debian based systems)
-- **pip** (Python 3 version)
-- **Graphviz**
-- **SDL2**
+- [Windows only] Hyper-V Virtualization enabled in BIOS (especially if using
+WSL2)
```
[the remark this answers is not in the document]
jbelaston-gpsw: Done.

doc/howto/installation.md
jbelaston-gpsw: Correct. Done

doc/howto/installation.md
jbelaston-gpsw: :+1: Done.

doc/howto/installation.md
jbelaston-gpsw: I mode this section under `## Quick user installation on Windows (MSVC toolchain)`

doc/howto/installation.md
jbelaston-gpsw: Oh ok, as previously discussed, let's do it with a generic approach for now.
Done :+1:

.github/workflows/ci_win.yml
jbelaston-gpsw: Correct. Not needed anymore.

Makefile
jbelaston-gpsw: Done.

external/.gitignore
jbelaston-gpsw: done

external/Makefile
jbelaston-gpsw: Oh great! Well observed (I will read the doc the next time :wink:) !
Thanks. Done

.github/workflows/ci_win.yml
jbelaston-gpsw: Yep it seems we need it. I get the following error if I remove it:
`'Error: 'vcpkgGitCommitId' input parameter must be provided when the specified vcpkg... is not a submodule.`
Also, it seems not trivial to specify a git tag...

Here an example of output:

```
2021-02-04T19:56:26.8308716Z   vcpkgArtifactIgnoreEntries: !.git
**/*
!installed
!/vcpkg
!vcpkg.exe
!vcpkgArtifactIgnoreEntries

2021-02-04T19:56:26.8309705Z   vcpkgGitURL: https://github.com/microsoft/vcpkg.git
2021-02-04T19:56:26.8310394Z   doNotUpdateVcpkg: false
2021-02-04T19:56:26.8310799Z   useShell: true
2021-02-04T19:56:26.8311166Z   doNotCache: false
2021-02-04T19:56:26.8311505Z ##[endgroup]
2021-02-04T19:56:27.0232119Z tool: C:\Program Files\Git\bin\git.exe
2021-02-04T19:56:27.0233179Z Fetching the Git commit id at 'D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg' ...
2021-02-04T19:56:27.0733730Z 56e30b011a870447acb62aa89f6db20023f1731b
2021-02-04T19:56:27.1721656Z vcpkg identified at commitId='56e30b011a870447acb62aa89f6db20023f1731b', adding it to the cache's key.
2021-02-04T19:56:27.1725760Z Cache's key = 'localGitId=42-args=-664396695-os=1553167195-appendedKey=42-triplet=1258145250'.
2021-02-04T19:56:27.1738622Z ##[group]Restore vcpkg and its artifacts from cache
2021-02-04T19:56:27.2022595Z Cache's key = 'localGitId=42-args=-664396695-os=1553167195-appendedKey=42-triplet=1258145250', paths = 'D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg,!D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg\packages,!D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg\buildtrees,!D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg\downloads'
2021-02-04T19:56:27.2024358Z Running restore-cache...
2021-02-04T19:56:27.2707487Z Cache miss.
2021-02-04T19:56:27.2708384Z ##[endgroup]
2021-02-04T19:56:27.2714712Z ⏱ elapsed: 0.189 seconds
2021-02-04T19:56:27.2719568Z ##[group]Set output env vars
2021-02-04T19:56:27.2726112Z Set the output variable 'RUNVCPKG_VCPKG_ROOT_OUT' to value: D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg
2021-02-04T19:56:27.2733571Z Set the environment variable 'RUNVCPKG_VCPKG_TRIPLET' to value: x64-windows
2021-02-04T19:56:27.2735356Z Set the output variable 'RUNVCPKG_VCPKG_TRIPLET_OUT' to value: x64-windows
2021-02-04T19:56:27.2736070Z ##[endgroup]
2021-02-04T19:56:27.2736628Z ⏱ elapsed: 0.002 seconds
2021-02-04T19:56:27.3037496Z tool: C:\Program Files\Git\bin\git.exe
2021-02-04T19:56:27.3038933Z Fetching the Git commit id at 'D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg' ...
2021-02-04T19:56:27.3289242Z 56e30b011a870447acb62aa89f6db20023f1731b
2021-02-04T19:56:27.3337522Z ##[group]Check whether vcpkg repository is up to date
2021-02-04T19:56:27.3340378Z Checking whether vcpkg's repository is updated to commit id '56e30b011a870447acb62aa89f6db20023f1731b' ...
2021-02-04T19:56:27.3563154Z tool: C:\Program Files\Git\bin\git.exe
2021-02-04T19:56:27.8071296Z error: pathspec 'D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg' did not match any file(s) known to git
2021-02-04T19:56:27.8192258Z Current commit id of vcpkg: '56e30b011a870447acb62aa89f6db20023f1731b'.
2021-02-04T19:56:27.8193032Z ##[endgroup]
2021-02-04T19:56:27.8193592Z ⏱ elapsed: 0.483 seconds
2021-02-04T19:56:27.8203801Z Error: 'vcpkgGitCommitId' input parameter must be provided when the specified vcpkg directory (D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg) is not a submodule.
2021-02-04T19:56:27.8205311Z     at VcpkgRunner.<anonymous> (D:\a\_actions\lukka\run-vcpkg\v6\dist\index.js:5708:31)
2021-02-04T19:56:27.8206091Z     at Generator.next (<anonymous>)
2021-02-04T19:56:27.8206777Z     at fulfilled (D:\a\_actions\lukka\run-vcpkg\v6\dist\index.js:5451:58)
2021-02-04T19:56:27.8207612Z     at processTicksAndRejections (internal/process/task_queues.js:93:5)
2021-02-04T19:56:27.8214094Z ##[error]run-vcpkg action execution failed: 'Error: 'vcpkgGitCommitId' input parameter must be provided when the specified vcpkg directory (D:\a\gopro-lib-node.gl\gopro-lib-node.gl\vcpkg) is not a submodule.
```

Makefile
jmoguillansky-gpsw: W_PREFIX is the windows full path to prefix, while PREFIX is the linux path to prefix.  It should be ok if PREFIX is relative or absolute, however, relative is easier to work with since we're using WSL and the paths are wsl-specific (e.g. /mnt/c/...) , they won't work in mingw64, etc.  

I think the reason is some bash commands needed to use the linux path while the windows commands needed to use the windows path.  
Also there's a target in the makefile: nodegl-%.  Does that target conflict with the prefix target?

Makefile
```diff
@@ -119,51 +181,101 @@ pynodegl-utils-install: pynodegl-utils-deps-install
...
 
 nodegl-tests: nodegl-install
-	(. $(ACTIVATE) && meson test -C builddir/libnodegl)
+ifeq ($(TARGET_OS),Windows)
+	($(CMD) $(ACTIVATE) \&\& meson test -C builddir\\libnodegl)
+else
+	($(CMD) $(ACTIVATE) && meson test -C builddir/libnodegl)
+endif
 
 nodegl-%: nodegl-setup
```
jmoguillansky-gpsw, line 273: Why are we using a wildcard target here?

mbouron-gpsw: To support the different libnodegl custom targets (updatedoc, updatespec, updateglwrappers) as make nodegl-{updatedoc,updatespecs,updateglwrappers).

Makefile
cboesch-gpsw: > There's a target in the makefile: nodegl-%. Does that target conflict with the prefix target?
> If I remember correctly, if you type make nodegl-env to run the prefix target, on some platforms that would run the nodegl-% target instead of the prefix target?

That's actually a very good point and likely a source of error. We need to rename either the venv directory or the prefix of that rule.

jbelaston-gpsw: So here the final work:
- rebase on `cboesch/make-venv` (`nodegl-env` bas been replaced by `nodegl-venv`)
- PREFIX split to PREFIX + PREFIX_FULLPATH (introduced in a separate commit)
- Use PREFIX_FULLPATH everywhere except for the target
Done.

Makefile
jbelaston-gpsw: Done.

cboesch-gpsw (approved):

Makefile
jmoguillansky-gpsw: It might be better to rename this target rule to something that doesn't overlap with nodegl-tests, nodegl-install, etc.
For example:
nodegl-target-%: nodegl-setup
ifeq ($(TARGET_OS),Windows)
	($(ACTIVATE) \&\& $(MESON_COMPILE) -C builddir\\libnodegl $(subst nodegl-target-,,$@))
else
	($(ACTIVATE) && $(MESON_COMPILE) -C builddir/libnodegl $(subst nodegl-target-,,$@))
endif

mbouron-gpsw (approved):

Makefile
```diff
@@ -89,14 +130,32 @@ all: ngl-tools-install pynodegl-utils-install
...
+else
+	@echo "        $(ACTIVATE)"
+endif
 	@echo
 
 ngl-tools-install: nodegl-install
-	(. $(ACTIVATE) && $(MESON_SETUP) ngl-tools builddir/ngl-tools && $(MESON_COMPILE) -C builddir/ngl-tools && $(MESON_INSTALL) -C builddir/ngl-tools)
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& $(MESON_SETUP) ngl-tools builddir\\ngl-tools \&\& $(MESON_COMPILE) -C builddir\\ngl-tools \&\& $(MESON_INSTALL) -C builddir\\ngl-tools)
+	$(CMD) xcopy /Y builddir\\ngl-tools\\*.dll "$(PREFIX_FULLPATH)\\Scripts\\."
```
jmoguillansky-gpsw, line 148: I think we can remove this line... meson install should now copy the dlls over to the correct path...
MESON_SETUP_PARAMS
--bindir

Makefile
```diff
@@ -119,51 +178,100 @@ pynodegl-utils-install: pynodegl-utils-deps-install
...
+	($(ACTIVATE) \&\& pip install -r pynodegl-utils\\requirements.txt)
+else ifneq ($(TARGET_OS),MinGW-w64)
+	($(ACTIVATE) && pip install -r ./pynodegl-utils/requirements.txt)
 endif
 
 pynodegl-install: pynodegl-deps-install
-	(. $(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
+ifeq ($(TARGET_OS),Windows)
+	($(ACTIVATE) \&\& pip -v install -e .\\pynodegl)
+	$(CMD) xcopy /Y builddir\\sxplayer\\*.dll pynodegl\\.
```
jmoguillansky-gpsw, line 190: TODO: we can probably remove these lines as well, and set PATH env variable instead

external/Makefile
```diff
@@ -37,7 +43,7 @@ all: sxplayer
 # than the prerequisite directory of the sxplayer rule. If this isn't true, the
 # symlink will be re-recreated on the next `make` call
 sxplayer: sxplayer-$(SXPLAYER_VERSION)
-ifneq ($(TARGET_OS),MinGW-w64)
+ifneq ($(TARGET_OS),$(filter $(TARGET_OS),MinGW-w64 Windows))
 	ln -snf $< $@
 else
 	cp -r $< $@
```
jmoguillansky-gpsw, line 49: TODO: we should be able to use symlink in mingw and msys (works in my environment)?

Makefile
jmoguillansky-gpsw: TODO:
It would be nice if we can improve the error handling for this target: know when the target actually completes.  
maybe at the end of venv target we can touch a file, i.e. venv.done, and then other targets can depend on venv.done instead of venv.

Makefile
```diff
@@ -37,18 +46,50 @@ DEBUG_SCENE ?= no
...
+ACTIVATE = $(CMD) $(VCVARS64) \&\& "$(PREFIX_FULLPATH)\\Scripts\\activate.bat"
+else
+ACTIVATE = . $(PREFIX_FULLPATH)/bin/activate
+endif
 
-RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX)/lib
+RPATH_LDFLAGS ?= -Wl,-rpath,$(PREFIX_FULLPATH)/lib
 
-MESON_SETUP   = meson setup --prefix=$(PREFIX) --pkg-config-path=$(PREFIX)/lib/pkgconfig -Drpath=true
-# MAKEFLAGS= is a workaround for the issue described here:
```
jmoguillansky-gpsw, line 49: when running make -j8 from command line, it tries to run separate make targets in parallel.  
So it will try to build sxplayer and build some other library in parallel.  And each of these builds use multiple cores.
The problem is that both of them write to the console at the same time, etc.
I think what we really want is:
the user should run make without -j8.
The makefile should internally call meson compile -jN where N is the number of cores.

external/Makefile
ubitux: Nah it doesn't work well when re-evaluated a 2nd time

jmoguillansky-gpsw: you mean you can't overwrite the symbolic link?
Works fine for me:

jmoguillansky@DESKTOP-S7C1I2O:/tmp$ touch t1 t2
jmoguillansky@DESKTOP-S7C1I2O:/tmp$ ln -snf t1 t
jmoguillansky@DESKTOP-S7C1I2O:/tmp$ ln -snf t2 t

cboesch-gpsw:
```make
linkfile: file
	ln -snf $< $@
file:
	touch $@
```

Try to run `make` twice with this `Makefile`

jmoguillansky-gpsw: jmoguillansky@DESKTOP-S7C1I2O:/tmp$ make
touch file
ln -snf file linkfile
jmoguillansky@DESKTOP-S7C1I2O:/tmp$ make
make: 'linkfile' is up to date.

Makefile
```diff
@@ -19,7 +19,12 @@
 # under the License.
 #
 
-PREFIX ?= $(PWD)/nodegl-env
+PREFIX          ?= venv
+ifeq ($(TARGET_OS),Windows)
```
jmoguillansky-gpsw, line 23: You should define TARGET_OS before...

PR #178, opened 2021-02-05
texture/buffer: add and use explicit usage flags

mbouron-gpsw:


libnodegl/backends/gl/texture_gl.c
```diff
@@ -253,7 +253,8 @@ static int texture_init_fields(struct texture *s)
     struct glcontext *gl = gctx_gl->glcontext;
     const struct texture_params *params = &s->params;
 
-    if (params->usage & NGLI_TEXTURE_USAGE_ATTACHMENT_ONLY) {
+    if (params->usage == NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT ||
+        params->usage == NGLI_TEXTURE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT) {
```
cboesch-gpsw, line 257: aren't you supposed to use `&` here?

libnodegl/backends/gl/texture_gl.c
```diff
@@ -430,7 +431,8 @@ int ngli_texture_gl_upload(struct texture *s, const uint8_t *data, int linesize)
 
     /* texture with external storage (including wrapped textures and render
      * buffers) cannot update their content with this function */
-    ngli_assert(!s->external_storage && !(params->usage & NGLI_TEXTURE_USAGE_ATTACHMENT_ONLY));
+    ngli_assert(!s->external_storage);
+    ngli_assert(params->usage & NGLI_TEXTURE_USAGE_TRANSFER_DST_BIT);
```
cboesch-gpsw, line 435: Earlier you replaced `NGLI_TEXTURE_USAGE_ATTACHMENT_ONLY` with `NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT` and `NGLI_TEXTURE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT`; why is this replaced with `NGLI_TEXTURE_USAGE_TRANSFER_DST_BIT` here?

libnodegl/hwupload.c
```diff
@@ -87,6 +87,9 @@ static int init_hwconv(struct ngl_node *node)
     params.format = NGLI_FORMAT_R8G8B8A8_UNORM;
     params.width  = mapped_image->params.width;
     params.height = mapped_image->params.height;
+    params.usage  = NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT |
+                    NGLI_TEXTURE_USAGE_SAMPLED_BIT          |
+                    NGLI_TEXTURE_USAGE_STORAGE_BIT;
```
cboesch-gpsw, line 92: style: `|` below the `=`?

libnodegl/backends/gl/texture_gl.c
mbouron-gpsw: == is intended, you want a renderbuffer only when there is no other usage than color or depth_stencil attachment.

libnodegl/backends/gl/texture_gl.c
mbouron-gpsw: NGLI_TEXTURE_USAGE_ATTACHMENT_ONLY meant either NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT or NGLI_TEXTURE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT.

Uploading data to a texture is only possible if the texture is flagged with NGLI_TEXTURE_USAGE_TRANSFER_DST_BIT (which does not exclude that the texture can also be flagged as a color or depth/stencil attachment).

libnodegl/hwupload.c
mbouron-gpsw: I removed this chunk which is not necesarry. params is copied from texture_priv->params which already contains the proper usage flags. I modified the following commit (texture: infer textures usage from the graph) to add NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT (since it is not set by default anymore in texture_prefetch()).

mbouron-gpsw: I've fixed the occurrence of this pattern.

cboesch-gpsw (approved):

libnodegl/node_texture.c
```diff
@@ -284,6 +284,10 @@ static int texture_prefetch(struct ngl_node *node)
         case NGL_NODE_BUFFERVEC2:
         case NGL_NODE_BUFFERVEC4: {
             struct buffer_priv *buffer = s->data_src->priv_data;
+            if (buffer->block) {
```
jmoguillansky-gpsw, line 292: should this be ngli_assert(buffer->block == 0)?
when the texture data source is a buffer, block should always be 0
You can't use a uniform buffer as a texture, etc

libnodegl/backends/gl/gctx_gl.c
```diff
@@ -108,6 +108,7 @@ static int wrap_capture_cvpixelbuffer(struct gctx *s,
         .format = NGLI_FORMAT_B8G8R8A8_UNORM,
         .width  = width,
         .height = height,
+        .usage  = NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT,
```
jmoguillansky-gpsw, line 111: why?  You're sampling the buffer, not rendering to it with the GPU

libnodegl/backends/gl/texture_gl.c
jmoguillansky-gpsw: shouldn't this be something more like if color attachment and not sampled or if depth_stencil attachment ?

libnodegl/texture.h
```diff
@@ -46,7 +46,14 @@ enum {
...
 };
 
-#define NGLI_TEXTURE_USAGE_ATTACHMENT_ONLY (1 << 0)
+enum {
+    NGLI_TEXTURE_USAGE_TRANSFER_SRC_BIT             = 1 << 0,
+    NGLI_TEXTURE_USAGE_TRANSFER_DST_BIT             = 1 << 1,
+    NGLI_TEXTURE_USAGE_SAMPLED_BIT                  = 1 << 2,
+    NGLI_TEXTURE_USAGE_STORAGE_BIT                  = 1 << 3,
+    NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT         = 1 << 4,
+    NGLI_TEXTURE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT = 1 << 5,
```
jmoguillansky-gpsw, line 55: we may want to add transient attachment bit (can be a future improvement...)

jmoguillansky-gpsw: I'm not sure if these will map to metal backend....  I think metal infers these based on the usage of the texture (like GL)

libnodegl/node_texture.c
```diff
@@ -253,16 +253,10 @@ static int texture_prefetch(struct ngl_node *node)
...
-    if (ngli_format_has_depth(params->format) || ngli_format_has_stencil(params->format)) {
-        params->usage |= NGLI_TEXTURE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT;
-    } else {
-        params->usage |= NGLI_TEXTURE_USAGE_STORAGE_BIT;
-        params->usage |= NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT;
-    }
+    params->usage |= NGLI_TEXTURE_USAGE_TRANSFER_DST_BIT | NGLI_TEXTURE_USAGE_SAMPLED_BIT;
+
+    if (params->mipmap_filter != NGLI_MIPMAP_FILTER_NONE)
+        params->usage |= NGLI_TEXTURE_USAGE_TRANSFER_SRC_BIT;
```
jmoguillansky-gpsw, line 259: how come you're removing the flags depth_stencil_attachment, storage, color_attachment?

libnodegl/node_texture.c
```diff
@@ -253,16 +253,10 @@ static int texture_prefetch(struct ngl_node *node)
     if (gctx->features & NGLI_FEATURE_TEXTURE_STORAGE)
         params->immutable = 1;
 
-    params->usage = NGLI_TEXTURE_USAGE_TRANSFER_SRC_BIT
-                  | NGLI_TEXTURE_USAGE_TRANSFER_DST_BIT
-                  | NGLI_TEXTURE_USAGE_SAMPLED_BIT;
-
-    if (ngli_format_has_depth(params->format) || ngli_format_has_stencil(params->format)) {
```
jmoguillansky-gpsw: how come you're removing the cases related to the flags depth_stencil_attachment, storage, color_attachment

libnodegl/pass.c
```diff
@@ -181,7 +181,7 @@ static int register_block(struct pass *s, const char *name, struct ngl_node *blo
      * Select buffer type. We prefer UBO over SSBO, but in the following
      * situations, UBO is not possible.
      */
-    int type = block->type == NGLI_TYPE_NONE ? NGLI_TYPE_UNIFORM_BUFFER : block->type;
+    int type = NGLI_TYPE_UNIFORM_BUFFER;
     if (block->layout == NGLI_BLOCK_LAYOUT_STD430) {
         LOG(DEBUG, "block %s has a std430 layout, declaring it as SSBO", name);
         type = NGLI_TYPE_STORAGE_BUFFER;
```
jmoguillansky-gpsw, line 188: I think a storage buffer can also be std140.  
But from a higher level perspective:
why are we inferring the buffer type passed on the layout parameter?
Isn't it simpler if the user directly indicates the storage type instead of setting a layout value and pgcraft inferring the buffer type?

libnodegl/nodes.h
```diff
@@ -92,6 +92,10 @@ struct ngl_ctx {
     struct darray modelview_matrix_stack;
     struct darray projection_matrix_stack;
     struct darray activitycheck_nodes;
+    /* The initialization of the block and buffer nodes is delayed because
+     * their usage is only known at the end of the init stage */
+    struct darray pending_init_buffer_nodes;
```
jmoguillansky-gpsw, line 97: why do you need this extra data structure?
If you're initializing the object on demand, when the usage is known, can you just check if the object is NULL, and if so, then initialize it there?

libnodegl/nodes.h
```diff
@@ -491,6 +495,7 @@ int ngli_prepare_draw(struct ngl_ctx *s, double t);
 void ngli_node_draw(struct ngl_node *node);
 
 int ngli_node_attach_ctx(struct ngl_node *node, struct ngl_ctx *ctx);
+int ngli_node_attach_ctx_internal(struct ngl_node *node, struct ngl_ctx *ctx);
```
jmoguillansky-gpsw, line 498: what's the difference between these two functions?
Can you add a comment describing the purpose?

libnodegl/buffer.h
```diff
@@ -25,8 +25,13 @@
 struct gctx;
 
 enum {
-    NGLI_BUFFER_USAGE_STATIC,
-    NGLI_BUFFER_USAGE_DYNAMIC,
+    NGLI_BUFFER_USAGE_DYNAMIC_BIT        = 1 << 0,
```
jmoguillansky-gpsw, line 28: I'm not sure if this flag (DYNAMIC_BIT) belongs here.  
This is more of a usage hint to the driver.  
Maybe this should be a separate enum NGLI_BUFFER_USAGE_HINT_DYNAMIC_BIT.
Also, this doesn't really map to vulkan.  
I think the closest equivalent in Vulkan would be HOST_VISIBLE_BIT?  
https://developer.arm.com/documentation/101897/0200/cpu-overheads/vulkan-cpu-memory-mapping

This is basically a hint to the GL driver to store the buffer in CPU memory instead of GPU memory (since the CPU is writing to it a lot)?

jmoguillansky-gpsw: Maybe once we remove GL backend we can remove this usage hint too...

jmoguillansky-gpsw: what happens if we remove BUFFER_USAGE_DYNAMIC and use GL_STATIC_DRAW in GL backend for everything?  What's the impact on performance?

libnodegl/pass.c
```diff
@@ -182,13 +182,16 @@ static int register_block(struct pass *s, const char *name, struct ngl_node *blo
      * situations, UBO is not possible.
      */
     int type = NGLI_TYPE_UNIFORM_BUFFER;
+    int usage = NGLI_BUFFER_USAGE_UNIFORM_BUFFER_BIT;
```
jmoguillansky-gpsw, line 185: do we need both?  Isn't this sort of redundant?
Can we just remove "type" and use "usage" instead?

libnodegl/pass.c
```diff
@@ -182,13 +182,16 @@ static int register_block(struct pass *s, const char *name, struct ngl_node *blo
...
     if (block->layout == NGLI_BLOCK_LAYOUT_STD430) {
         LOG(DEBUG, "block %s has a std430 layout, declaring it as SSBO", name);
         type = NGLI_TYPE_STORAGE_BUFFER;
```
jmoguillansky-gpsw, line 188: There must be a better way to detect the type of the buffer than from the layout declaration...?

libnodegl/backends/gl/gctx_gl.c
mbouron-gpsw: We are rendering to it, the video buffer acts as a render target.

libnodegl/texture.h
mbouron-gpsw: This will be added later on.

libnodegl/node_texture.c
mbouron-gpsw: They are inferred from the graph, ie: the parent nodes are responsible from settings the correct usage flags.

libnodegl/node_texture.c
mbouron-gpsw: We don't want to assert when the user doesn't use the API properly.

libnodegl/pass.c
mbouron-gpsw: Yes a storage buffer can also be std140. We are selecting the buffer type based on the layout because the nodegl API let the user select the layout of the block node. We were considering removing the layout parameter from the block node. We would still need to set the buffer type depending on the block size and if the buffer is marked as writable in the resource props. A future improvement could be either specifying the block type to the block node or adding a "storage" hint to the resource props (like we have a writable property).

libnodegl/buffer.h
mbouron-gpsw: I kept the flag only for OpenGL compatibility and wanted to avoid adding another parameter for now. As you said, this usage hint does not to map the vulkan backend at all.

If we use GL_STATIC_DRAW all the time, the driver will complain when the buffer is updated more than once. A possible solution would be use GL_DYNAMIC_DRAW all the time and remove the flag but I don't know the performance implications as it's highly driver / gpu dependent. And yes, this flag will be removed once we get rid of OpenGL. I would keep it this way for now so we don't introduce any potential performance regressions on the OpenGL backend.

libnodegl/pass.c
mbouron-gpsw: Yes as I said above, we should remove probably remove the layout parameter from the block node and either add a type parameter to the block node or a "storage" field to the resource props.

libnodegl/pass.c
mbouron-gpsw: It depends if we want to keep the possibility to bind the same buffer differently depending on the pipeline and It also depends on the way we infer the type from the graph (ie: having layout removed from the block node + either adding a type parameter to the block node or adding a "storage" hint to the resource props of the pipeline).

libnodegl/nodes.h
mbouron-gpsw: I will come up with another approach (i'm currently testing it locally) that do not require another data structure and simplify the diff.

libnodegl/nodes.h
mbouron-gpsw: This chunk has been removed in favor of a cleaner solution to defer pass buffers initialization.

libnodegl/nodes.h
mbouron-gpsw: PR updated. Deferring pass buffer initialization is now local to pass.c.

libnodegl/backends/gl/texture_gl.c
mbouron: Renderbuffers cannot be sampled nor used as an image nor you can't upload data to it and these texture usage flags do not map to OpenGL.
For example, allowing TRANSFER_DST with renderbuffers means that the user would expect ngli_texture_upload() to function properly which is not possible. 
IMHO, it's safer to use renderbuffers only if the usage is COLOR_ATTACHMENT or DEPTH_STENCIL_ATTACHMENT.

jmoguillansky-gpsw: Sounds good.  Maybe consider adding a comment (this code might be a bit confusing)

libnodegl/node_texture.c
jmoguillansky-gpsw: Maybe this parameter checking should be done at a higher level, like api.c?

libnodegl/backends/gl/gctx_gl.c
jmoguillansky-gpsw: What are you rendering to it?

libnodegl/pass.c
jmoguillansky-gpsw: I don't think we needa buffer "type".  A buffer can have one or more usages.  e.g. in vulkan we have vkbuffer and in metal, mtlbuffer

mbouron-gpsw: I agree. Regarding the removal of type in favor of usage, that would require more changes to pgcraft. We can't use pgcraft_block.buffer.usage directly as it could be UNIFORM,STORAGE or UNIFORM+STORAGE and we would still need to a have a dedicated field in the pgcraft_block structure, for example we can have pgcraft_block.usage instead of pgcraft_block.type but it would require to check that only one bit flag is set. It would also require to do the same for textures for consistency. IMHO this is not a good idea and it makes more sense to have separate enums. Also we'll probably want to avoid passing the actual resources (buffers, textures) to pgcraft at some point.

libnodegl/node_texture.c
mbouron-gpsw: We currently enforce restrictions directly in the nodes code (See node_media.c:228, pass.c:376, node_animated.c:185 for examples).
Moving these checks to a higher level would require to either add nodes specific code (that perform the checks) to nodes.c or add some way to traverse in api.c + nodes specific code.

PR #179, opened 2021-02-05
limits: rename max_work_group_{counts,sizes} to max_work_group_{count,size} 

mbouron-gpsw:


ubitux (approved):
"Makes the field names match the Khronos nomenclature." I think

LGTM otherwise

mbouron-gpsw:
> "Makes the field names match the Khronos nomenclature." I think
> 
> LGTM otherwise

Fixed. Thanks.

PR #181, opened 2021-02-07
Noise node

cboesch-gpsw:


libnodegl/node_noise.c
```diff
@@ -0,0 +1,149 @@
...
+}
+
+static int init_noise_generators(struct noise_priv *s, int n)
+{
+    /*
+     * Every generator is instanciated the same, except for the seed: the seed
+     * offset is defined to create a large gap between every components to keep
+     * the overlap to the minimum possible
+     */
+    const uint32_t seed_offset = 0xffffffff / n;
```
mbouron-gpsw, line 111: UINT_MAX ?

tests/data.py
```diff
@@ -338,3 +340,26 @@ def data_integer_iovars(cfg):
...
     render.update_vert_resources(color_u32=ngl.UniformIVec4(value=(0x50, 0x80, 0xa0, 0xff)))
     return render
+
+
+@test_fingerprint(nb_keyframes=30, tolerance=1)
+@scene()
+def data_noise(cfg):
+    cfg.aspect_ratio = (1, 1)
+    cfg.duration = 3
+    vert = '''void main() {
```
mbouron-gpsw, line 350: Can you put the `{` on a new line ?

tests/data.py
```diff
@@ -338,3 +340,26 @@ def data_integer_iovars(cfg):
...
+    ngl_out_pos = ngl_projection_matrix * ngl_modelview_matrix * ngl_position;
+    ngl_out_pos += vec4(wiggle, 0, 0);
+}
+'''
+    frag = '''void main() {
```
mbouron-gpsw, line 355: Same as above.

libnodegl/noise.h
```diff
@@ -0,0 +1,53 @@
...
+};
+
+typedef float (*interp_func_type)(float t);
+
+struct noise_params {
+    double amplitude;
+    int octaves;
+    double lacunarity;
+    double gain;
+    uint32_t seed;
```
mbouron-gpsw, line 41: You are using a uint32_t for the seed but you do not explicit sized types for `int octaves;` and `int function;`, can you change them to int32_t ? so the whole file is consistent. The same show also probably apply to `noise.c` ?

tests/data.py
cboesch-gpsw: It was kept that consistent with how it's currently done in this file. Adding a line break also forces me to put a line break after `'''` which I find unnecessary. I don't agree with this cosmetic but whatever, changed.

libnodegl/node_noise.c
cboesch-gpsw: replaced with UINT32_MAX

libnodegl/noise.h
cboesch-gpsw: we are iterating with `for (int i = 0; i < p->octaves; i++) {` for `octaves` because the size doesn't really matter, only being an integer matters. For `function`, `int` is what we use everywhere for enums so again, consistency with the rest of the codebase. `seed` is special: it needs to be unsigned (we need it to overflow), and exactly 32-bits (because of the hashing). So basically `seed` is the only field that requires being specific. I don't mind switching all the other ints but I believe it needs to be broad to the codebase.

mbouron-gpsw: Ok.

mbouron-gpsw (approved):

PR #182, opened 2021-02-08
doc: add Graphic configuration explanation

cboesch-gpsw:


doc/expl/graphicconfig.md
```diff
@@ -0,0 +1,93 @@
...
+Color and alpha blending, depth, stencil and culling are controlled by the
+`GraphicConfig` node.
+
+`node.gl` is aligned with the traditional blending mode concepts from graphics
+APIs. For example, color and alpha blending involves source and destination
+factors for color channels or alpha, and a blending operation (addition,
+subtraction, etc). This document will assume you are familiar with these
+concepts.
+
+For color and alpha blending, you will interested in particular in the
```
sfranz-gpsw, line 12: you will interested  in particular in -> you will be particularly interested in

sfranz-gpsw (approved):
Thank you for the doc! Except for my little comment, LGTM

doc/expl/graphicconfig.md
cboesch-gpsw: thanks, fixed

mbouron-gpsw:
LGTM.

Not directly related to this PR but we use "graphic" in a lot of place of the codebase but it looks like the right term is "graphic*s*". We might want to fix that at some point.

mbouron-gpsw (approved):

ubitux:
> LGTM.
>

Thanks, will merge
 
> Not directly related to this PR but we use "graphic" in a lot of place of the codebase but it looks like the right term is "graphic_s_". We might want to fix that at some point.

The document use "graphic" in the following situations:
1. to quote `GraphicConfig` (which I can't change)
2. in the title as "Graphic configuration" but I'd like to keep it this way for now since it matches `GraphicConfig`
3. for "Graphic state" -- somehow it sounds more correct to me that way but I might be wrong

I reference OTOH "Graphics API"...

So yeah, it's unclear to me what's the correct way to write it, but we should probably change the node name first if it's incorrect.

PR #183, opened 2021-02-09
api: wait for the device to be idle before releasing the graph resources

mbouron-gpsw:


libnodegl/api.c
```diff
@@ -77,8 +77,17 @@ static int get_default_platform(void)
...
 }
 
+#define KEEP_SCENE  0
+#define UNREF_SCENE 1
+
 static int cmd_stop(struct ngl_ctx *s, void *arg)
 {
+    if (s->scene) {
+        ngli_node_detach_ctx(s->scene, s);
+        int unref_scene = *(int *)arg;
```
cboesch-gpsw, line 89: Can you inline the intermediate `unref_scene` or rename it to `flags` or something? (`unref_scene = KEEP_SCENE` is weird)

libnodegl/api.c
```diff
@@ -211,8 +213,10 @@ static int cmd_set_scene(struct ngl_ctx *s, void *arg)
         ngli_node_detach_ctx(s->scene, s);
         ngl_node_unrefp(&s->scene);
     }
-    ngli_rnode_clear(&s->rnode);
+    ngli_rnode_reset(&s->rnode);
 
+    ngli_rnode_init(&s->rnode);
```
cboesch-gpsw, line 227: `clear` is `reset`+`init` already: don't we want to keep the existing `clear` here?

mbouron-gpsw: Using reset + init is more explicit and more in line with our other patterns ? Since ngli_rnode_clear() is not used anymore should i remove it in a separate commit ?

cboesch-gpsw: the reset/clear difference is confusing so I don't mind getting rid of it

libnodegl/api.c
mbouron-gpsw: Fixed by inlining the check.

libnodegl/api.c
mbouron-gpsw: Updated.

cboesch-gpsw (approved):

PR #190, opened 2021-02-17
Jbelaston/automatic vcvars64 detection

jbelaston-gpsw:
Automatic detection and use of vcvars64.bat in the makefile.
This detection is done using Microsoft vswhere.exe utility (more info in the commit descriptions).

Note that I used a powershell script and not a batch (cmd) one since it is really friendly to use and especially is now the modern Windows approach. Would be good to progressively move all our current batches to powershell...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ $(info PREFIX_FULLPATH: $(PREFIX_FULLPATH))
 ifeq ($(TARGET_OS),Windows)
 # Initialize VCVARS64 and VCPKG_DIR to a default value
 # Note: the user should override this environment variable if needed
-VCVARS64 ?= "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
+VCVARS64 ?= "$(shell powershell.exe .\\scripts\\find_vcvars64.ps1)"
 VCPKG_DIR ?= C:\\vcpkg
 PKG_CONF_DIR = external\\pkgconf\\build
 CMD = PKG_CONFIG_ALLOW_SYSTEM_CFLAGS=1 PKG_CONFIG_ALLOW_SYSTEM_LIBS=1 PKG_CONFIG="$(PREFIX_FULLPATH)\\Scripts\\pkg-config.exe" PKG_CONFIG_PATH="$(VCPKG_DIR)\\installed\\x64-windows\\lib\\pkgconfig" WSLENV=PKG_CONFIG/w:PKG_CONFIG_PATH/w:PKG_CONFIG_ALLOW_SYSTEM_LIBS/w:PKG_CONFIG_ALLOW_SYSTEM_CFLAGS/w cmd.exe /C
--- a/.github/workflows/ci_win.yml
+++ b/.github/workflows/ci_win.yml
@@ -69,5 +69,4 @@ jobs:
       - name: Build
         run: |
           make -j$(($(nproc)+1)) TARGET_OS=Windows \
-          VCPKG_DIR='D:\\a\\gopro-lib-node.gl\\gopro-lib-node.gl\\vcpkg' \
-          VCVARS64='"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"'
+          VCPKG_DIR='D:\\a\\gopro-lib-node.gl\\gopro-lib-node.gl\\vcpkg'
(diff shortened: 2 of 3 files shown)


scripts/get_vcvars64.ps1
```diff
@@ -0,0 +1,13 @@
+# Download latest vswhere.exe from microsoft server into windows temp directory
+$vswhereLatest = "https://github.com/Microsoft/vswhere/releases/latest/download/vswhere.exe"
+$vswhere = "$env:temp\vswhere.exe"
+if ([System.IO.File]::Exists($vswhere)) {
+    remove-item $vswhere
+}
+invoke-webrequest $vswhereLatest -OutFile $vswhere
```
cboesch-gpsw, line 7: So this is going to re-download a binary at every `make` call?

How about the following:

In `external/Makefile`, add a `vswhere.exe` target that triggers the download:
```make
vswhere.exe:
	$(CURL) -L https://github.com/Microsoft/vswhere/releases/latest/download/$@ -o $@
```

You add it to the `DEPENDENCIES`.

One thing to make sure of: `$(PREFIX)` in the root `Makefile` will need to have `external-download` in requirement.

jmoguillansky-gpsw: You shouldn't need to download vswhere, its already in the visual studio installer path.

jmoguillansky-gpsw:
> Automatic detection and use of vcvars64.bat in the makefile.
> This detection is done using Microsoft vswhere.exe utility (more info in the commit descriptions).
> 
> Note that I used a powershell script and not a batch (cmd) one since it is really friendly to use and especially is now the modern Windows approach. Would be good to progressively move all our current batches to powershell...

I agree we can use powershell if the script is specific to windows.  If it's cross platform, then maybe better to use bash or python

scripts/get_vcvars64.ps1
jbelaston-gpsw: Yeah you're right. 

In fact, i was not really confident on the fact that `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe` is really fixed as said in the official wiki doc (https://github.com/Microsoft/vswhere/wiki/Installing).
One of my interrogation were "is it possible for a user to specialize the VS installation location ?" and after investigation, the answer is clear: No! Or let's say "Not really": the user can specialize some (shared) components via Visual Studio Installer but the basis is fixed. So clearly no need to download it! Thanks.

Now anyway, after more investigation, further to your comment and discussion with @cboesch-gpsw, i changed for another approach (better than using vswhere): "Visual Studio Setup PowerShell". More info in the commits.

jbelaston-gpsw: @cboesch-gpsw 
I took into account your proposal regarding dependencies link between `PREFIX` and `external-download` targets
See https://github.com/gopro/gopro-lib-node.gl/pull/213
Thanks.

jmoguillansky-gpsw:
lgtm

Makefile
```diff
@@ -53,7 +53,7 @@ $(info PREFIX_FULLPATH: $(PREFIX_FULLPATH))
 ifeq ($(TARGET_OS),Windows)
 # Initialize VCVARS64 and VCPKG_DIR to a default value
 # Note: the user should override this environment variable if needed
-VCVARS64 ?= "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
+VCVARS64 ?= "$(shell powershell.exe .\\scripts\\get_vcvars64.ps1)"
```
jmoguillansky-gpsw, line 56: some minor comments:
1) suggest moving the script to build_scripts/windows
2) I think this can be simplified a little:
VCVARS64 ?= "$(powershell.exe ./build_scripts/windows/get_vcvars64.ps1)"

jbelaston-gpsw: Thanks for this proposals :+1: 
- For the 1. i did it the same way than for `libnodegl/scripts` (to be consistent). if you're ok, we will do your proposal for all `scripts` folders in a separate cleanup commit, another time...
- I well did the renaming. You're right it's better.
- I tried without `shell` and it seems, it doesn't work. So I keep it like that for now.

scripts/get_vcvars64.ps1
jbelaston-gpsw: Finally #213 is closed. More info in this PR.

cboesch-gpsw (approved):

mbouron-gpsw (approved):

PR #191, opened 2021-02-18
nodes: invalidate the node branches on live change

mbouron-gpsw:


libnodegl/node_uniform.c
```diff
@@ -129,8 +128,6 @@ static const struct node_param uniformmat4_params[] = {
 
 static int uniform_update(struct ngl_node *node, double t)
 {
-    struct variable_priv *s = node->priv_data;
-    s->live_changed = 0;
     return 0;
 }
```
cboesch-gpsw, line 125: I believe you can `#define uniform_update NULL` here

libnodegl/node_uniform.c
```diff
@@ -61,20 +52,22 @@ DECLARE_UPDATE_FUNC(mat4,  s->opt.mat)
...
 static int uniformfloat_update_func(struct ngl_node *node)
 {
-    int ret = set_live_changed(node);
-    if (ret < 0)
-        return ret;
     struct variable_priv *s = node->priv_data;
+    if (s->dynamic) {
+        LOG(ERROR, "updating data on a dynamic uniform is unsupported");
+        return NGL_ERROR_INVALID_USAGE;
+    }
```
cboesch-gpsw, line 59: Not a huge fan of inlining N times the same error message; how about renaming the function instead to `check_not_dynamic()` or something?

tests/live.py
```diff
@@ -171,6 +173,59 @@ def scene_func(cfg, seed=0, debug_positions=True, color_tint=False, trf_step=0):
     return scene_func
 
 
+def _get_live_shared_uniform_cuepoints():
+    return dict((('0', (-0.5, -0.5)), ('1', (0.5, 0.5))))
```
cboesch-gpsw, line 177: Why don't you just declare a `_SHARED_UNIFORM_CUEPOINTS`? The function looks unnecessary here

tests/live.py
```diff
@@ -183,5 +238,7 @@ def _bootstrap():
             field_id = '{category}_{type}'.format(**field_info)
             globals()['live_trf_{}_{}'.format(field_id, layout)] = _get_live_trf_function(spec, field_id, layout)
 
+    globals()['live_shared_uniform'] = _get_live_shared_uniform_function()
```
cboesch-gpsw, line 241: You don't need a builder; why not simply declare the function directly?

tests/refs/live_shared_uniform.ref
```diff
@@ -0,0 +1 @@
+0:FF0000FF 1:FF0000FF
```
cboesch-gpsw, line 1: You want at least 2 keyframes so that we see the effect of the live change from the original state

tests/live.py
```diff
@@ -171,6 +173,59 @@ def scene_func(cfg, seed=0, debug_positions=True, color_tint=False, trf_step=0):
...
+    group = ngl.Group()
+    for i in range(2):
+        block = ngl.Block(fields=[color])
+        quad = ngl.Quad((-1 + i, -1 + i, 0), (1, 0, 0), (0, 1, 0))
+        render = ngl.Render(quad, program)
+        render.update_frag_resources(data=block)
+        group.add_children(render)
+    if debug_positions:
+        cuepoints = _get_live_shared_uniform_cuepoints()
+        print(cuepoints)
```
cboesch-gpsw, line 205: please remove this `print`

mbouron-gpsw: Done

tests/live.py
mbouron-gpsw: The builder is needed to keep the uniform declaration local. I'll set the function directly instead of using globals().

tests/refs/live_shared_uniform.ref
mbouron-gpsw: Done.

tests/live.py
mbouron-gpsw: Fixed.

tests/live.py
mbouron-gpsw: Done.

libnodegl/node_uniform.c
mbouron-gpsw: Done.

libnodegl/node_uniform.c
mbouron-gpsw: The function has been renamed to is_live_update_supported().

libnodegl/nodes.c
```diff
@@ -611,6 +619,18 @@ int ngl_node_param_add(struct ngl_node *node, const char *key,
     return ret;
 }
 
+static int node_invalidate_branch(struct ngl_node *node)
```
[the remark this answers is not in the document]
mbouron-gpsw: I instead added the relevant return checks.

libnodegl/nodes.c
```diff
@@ -611,6 +619,18 @@ int ngl_node_param_add(struct ngl_node *node, const char *key,
...
+{
+    node->last_update_time = -1;
+    if (node->class->invalidate)
+        node->class->invalidate(node);
+    struct ngl_node **parents = ngli_darray_data(&node->parents);
+    for (int i = 0; i < ngli_darray_count(&node->parents); i++) {
+        node_invalidate_branch(parents[i]);
+    }
```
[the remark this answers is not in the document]
mbouron-gpsw: `{` `}` are now needed as i've added a check of the return of node_invalidate_branch().

libnodegl/nodes.c
```diff
@@ -230,19 +230,23 @@ static int track_children(struct ngl_node *node)
...
             }
             case PARAM_TYPE_NODEDICT: {
                 struct hmap *hmap = *(struct hmap **)(base_ptr + par->offset);
                 if (!hmap)
                     break;
                 const struct hmap_entry *entry = NULL;
-                while ((entry = ngli_hmap_next(hmap, entry)))
-                    if (!ngli_darray_push(&node->children, &entry->data))
+                while ((entry = ngli_hmap_next(hmap, entry))) {
+                    struct ngl_node *child = entry->data;
```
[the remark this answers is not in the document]
mbouron-gpsw: We were not using const before while adding the element to the children array and i believe the children nodes are not meant to be const (even though the darray api as no notion of constness).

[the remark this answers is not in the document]
cboesch-gpsw: `entry` is const, and `void *ngli_darray_push(struct darray *darray, const void *element)`

We will modify the elements through the array, but the operation of push is const.

libnodegl/node_uniform.c
cboesch-gpsw: Nice; you could take the opportunity to change the prototype to const (it wasn't before because it was mutating the state, but now it's only a check)

mbouron-gpsw: Done in a separate commit.

cboesch-gpsw (approved):

PR #193, opened 2021-02-19
darray: add ngli_darray_set()

jmoguillansky-gpsw:
Signed-off-by: Matthieu Bouron <matthieu.bouron@gmail.com>

### DIFF
--- a/libnodegl/darray.c
+++ b/libnodegl/darray.c
@@ -107,6 +107,15 @@ void *ngli_darray_get(const struct darray *darray, int index)
     return darray->data + index * darray->element_size;
 }
 
+int ngli_darray_set(const struct darray *darray, int index, const void *element)
+{
+    void *dst = ngli_darray_get(darray, index);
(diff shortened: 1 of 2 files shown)


cboesch-gpsw:
What do you need this for?

jmoguillansky-gpsw:
It’s a convenience api to set an item in an array.
We have api to get an item, we should probably have an api to set an item.
I think this is already signed/merged in vulkan , but not yet in master?

cboesch-gpsw:
> It’s a convenience api to set an item in an array. We have api to get an item, we should probably have an api to set an item.

If it's not used, I don't see why. It also means it wouldn't be tested. I believe it was added because we needed it but don't anymore.

> I think this is already signed/merged in vulkan , but not yet in master?

It's been removed because it serves no purpose currently, afaict.

jmoguillansky-gpsw:
you're right, we can close this

PR #195, opened 2021-02-19
nodegl: add extern C to headers

jmoguillansky-gpsw:
Nodegl can be used from both C and C++ code.
This makes it much easier to use Node.GL from C++ code,
(and is a typical approach)


cboesch-gpsw:
Sorry but I totally disagree with this; this is not the responsibility of the C code to be compatible with C++. Add it in every C++ file you need instead.

jmoguillansky-gpsw:
This is the standard approach that C header files use?  Vulkan.h, opengl.h, etc?
It makes it easier to use from C++ code.  Why do you dislike this approach?

PR #199, opened 2021-02-19
buffer: add offset param

jmoguillansky-gpsw:


cboesch-gpsw (approved):
This looks fine; just curious, do you plan to actually rely on this for something?

mbouron-gpsw (approved):

jmoguillansky-gpsw:
> This looks fine; just curious, do you plan to actually rely on this for something?

Yes its important for updating a portion of a buffer

cboesch-gpsw:
> > This looks fine; just curious, do you plan to actually rely on this for something?
> 
> Yes its important for updating a portion of a buffer

Yeah I understand; my question was whether you actually use that feature or not yet.

jmoguillansky-gpsw:
> > > This looks fine; just curious, do you plan to actually rely on this for something?
> > 
> > 
> > Yes its important for updating a portion of a buffer
> 
> Yeah I understand; my question was whether you actually use that feature or not yet.

I don't believe I'm using it yet...  I think we need some logic at a higher level to see which portion of the buffer has changed, or expose this to the user at the node API level

PR #200, opened 2021-02-19
ngl-tools: instruct NSView to create a layer

jmoguillansky-gpsw:
The NSView can optionally be backed by a layer, which is a buffer that
stores the view contents.  We want to back the view by a layer,
so that we can draw directly to it


jmoguillansky-gpsw:
closed, this code is already in vulkan branch

PR #203, opened 2021-02-19
pynodegl-utils: add log messages for test passed / failed

jmoguillansky-gpsw:
Currently, we only have an exit code to indicate if the test passed or failed.  
This makes it more clear to the user if the test passed or failed, especially if running the test manually (not via meson)

### DIFF
--- a/pynodegl-utils/pynodegl_utils/tests/__init__.py
+++ b/pynodegl-utils/pynodegl_utils/tests/__init__.py
@@ -141,7 +141,8 @@ def run():
     test_func = _refgen_map[refgen_opt]
     err = test_func(func_name, tester, ref_filepath, dump)
     if err:
+        sys.stderr.write(f'{func_name} failed\n')
         sys.stderr.write('\n'.join(err) + '\n')
         sys.exit(1)
-
+    print(f'{func_name} passed')
     sys.exit(0)


pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -141,7 +141,8 @@ def run():
     test_func = _refgen_map[refgen_opt]
     err = test_func(func_name, tester, ref_filepath, dump)
     if err:
+        print(f'{func_name} failed')
         sys.stderr.write('\n'.join(err) + '\n')
```
ubitux, line 145: This is an error so it should be part of the `stderr`; maybe add this to the log below

jmoguillansky-gpsw: updated

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -141,7 +141,7 @@ def run():
     test_func = _refgen_map[refgen_opt]
     err = test_func(func_name, tester, ref_filepath, dump)
     if err:
-        sys.stderr.write('\n'.join(err) + '\n')
+        sys.stderr.write(f'{func_name} failed\n'.join(err) + '\n')
```
cboesch-gpsw, line 144: `join` doesn't work that way... Please test your changes

mbouron-gpsw: You can keep it separated:
```python
sys.stderr.write(f'{func_name} failed\n')
sys.stderr.write('\n'.join(err) + '\n')
```

jmoguillansky-gpsw: updated

cboesch-gpsw (commented):
Should be fine but please use `utils/tests:` as prefix

jmoguillansky-gpsw:
> Should be fine but please use `utils/tests:` as prefix

updated

PR #204, opened 2021-02-19
git: disable automatic line ending conversion

jmoguillansky-gpsw:
The main motivation for this PR is to enable using bash scripts on Windows.

This disables automatic conversion of line feeds from /n to /r/n for text files
We cannot use windows style line endings in bash scripts, they specifically have to be unix-style line endings


jbelaston-gpsw (commented):
Hi @jmoguillansky-gpsw 
I well understand the motivation... Now I discussed with @cboesch-gpsw & @mbouron about that (one month ago I think...) and it I think you also had this discussion...  and the conclusion was for now to let the user manage it in its own git configuration (using `autocrlf` git property...)

jmoguillansky-gpsw:
> Hi @jmoguillansky-gpsw
> I well understand the motivation... Now I discussed with @cboesch-gpsw & @mbouron about that (one month ago I think...) and I think you also had this discussion with them... and the conclusion and best approach was to, for now, let the user manage it in its own git configuration (using `autocrlf` git property...)

By dedault, the ci server was converting the line endings to windows style, which is nor acceptable for bash scripts.  With this small change, we disable line ending conversion for bash scripts

PR #205, opened 2021-02-19
pynodegl-utils: add dll search directories

jmoguillansky-gpsw:
The main benefit of this PR is that it allows us to avoid copying the DLLs to multiple directories (and hence simplify the build and runtime environment).  

Currently, when loading a python module, the DLL search path is set to the directory containing the python module.
It doesn't search the directories specified in PATH environment variable.
see https://bugs.python.org/issue43173
With this change, we add the directories in PATH environment variable, along with the venv/Scripts folder as DLL search paths on Windows.

### DIFF
--- a/pynodegl-utils/pynodegl_utils/tests/__init__.py
+++ b/pynodegl-utils/pynodegl_utils/tests/__init__.py
@@ -23,6 +23,15 @@
 import os
 import os.path as op
 import sys
+import platform
+
+
+if platform.system() == 'Windows':
+    env_path = os.getenv('PATH').split(';')
+    for p in env_path:
+        if op.exists(p):
+            os.add_dll_directory(p)
+
 
 from pynodegl_utils.com import load_script
 


pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -23,6 +23,17 @@
 import os
 import os.path as op
 import sys
+import platform
+import pathlib
+
+if platform.system() == 'Windows':
+    prefix = pathlib.Path(__file__).parent.absolute().joinpath('../../../venv')
```
cboesch-gpsw, line 30: hum. I don't think this is acceptable

jmoguillansky-gpsw: you mean hardcoding venv?  how can we pass the prefix variable?  maybe as an environment variable?

jmoguillansky-gpsw: updated.

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -23,6 +23,15 @@
 import os
 import os.path as op
 import sys
+import platform
+import pathlib
```
ubitux, line 27: pathlib is unused

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -23,6 +23,15 @@
...
+
```
ubitux, line 28: missing 2nd line break

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -23,6 +23,15 @@
...
+if platform.system() == 'Windows':
+    env_path = os.getenv("PATH").split(';')
+    for p in env_path:
+        if os.path.exists(p):
+            os.add_dll_directory(p)
```
ubitux, line 33: Can you explain in the commit description the reason why Python doesn't do it itself?

ubitux (changes requested):

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: it's already explained in the link
https://bugs.python.org/issue43173

In 3.8+, the search path for the dependent DLLs of a normally imported extension module includes the following directories:

    * the loaded extension module's directory
    * the application directory (e.g. that of python.exe)
    * the user DLL search directories that get added by 
      SetDllDirectory() and AddDllDirectory(), such as with 
      os.add_dll_directory()
    * %SystemRoot%\System32

Note that the above list does not include the current working directory or %PATH% directories.

ubitux: Unless I'm mistaken, nothing explains here *why* `PATH` is not honored.

jmoguillansky-gpsw: the Python developers explained in the issue tracker that it doesn't honor PATH environment variable when searching for dlls

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: updated

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: updated

pynodegl-utils/pynodegl_utils/tests/__init__.py
ubitux: Yeah. But **why**?

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -23,6 +23,15 @@
 import os
 import os.path as op
 import sys
+import platform
+
+
+if platform.system() == 'Windows':
+    env_path = os.getenv("PATH").split(';')
```
ubitux, line 30: consistent quoting please

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -23,6 +23,15 @@
...
+    for p in env_path:
+        if os.path.exists(p):
```
ubitux, line 32: you can use `op`

ubitux (changes requested):

ubitux:
commit message: please use `utils:` as prefix

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: updated

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: updated

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: > commit message: please use `utils:` as prefix

updated

jmoguillansky-gpsw: > Yeah. But **why**?

I don't know, need to ask the Python developers

jmoguillansky-gpsw:
> commit message: please use `utils:` as prefix

Updated

cboesch-gpsw (changes requested):
You're doing this dll hook at the import time of the `ngl-test` program: why at import time and not in its main? This code could be triggered as soon as the module gets imported (by side effect or not). What's the goal here? You want it only for `ngl-test` or utils wide?

jmoguillansky-gpsw:
> You're doing this dll hook at the import time of the `ngl-test` program: why at import time and not in its main? This code could be triggered as soon as the module gets imported (by side effect or not). What's the goal here? You want it only for `ngl-test` or utils wide?

You're right, we want it utils wide.
I will try this suggestion...

pynodegl-utils/pynodegl_utils/tests/__init__.py
jmoguillansky-gpsw: my guess is that it simplifies the configuration: one less variable to consider (and which the user may modify).  
If we look at the typical approach currently used in Python, all the dlls seem to be duplicated in the python module folder.
For example, for PySide2 (python Qt bindings), all the Qt DLLs are stored twice: once in system's Qt installation and once in PySide2

Maybe it's better to just have duplicate copies of the DLL for now since that seems to be consistent with Python's current approach.
For long term we can consider merging pynodegl and pynodegl-utils into one module (pynodegl) to simplify install, and this way we'll have less duplicate modules.

jbelaston-gpsw (commented):
Is this change needed for easy debug ?

jmoguillansky-gpsw:
> Is this change needed for easy debug ?

No it's to avoid duplicate copies of dlls

PR #207, opened 2021-02-19
ci: (windows) misc

jmoguillansky-gpsw:
Split mingw and msvc tests into separate workflows
Improve vcpkg caching mechanism


cboesch-gpsw:
Isn't this going to break the badge in the README?

.github/workflows/ci_msvc.yml
```diff
@@ -0,0 +1,56 @@
...
+          pip.exe install meson
+          cd /mnt/c/vcpkg
+          ./vcpkg.exe install pthreads:x64-windows opengl-registry:x64-windows ffmpeg[ffmpeg,ffprobe]:x64-windows sdl2:x64-windows
+          # TODO: remove vcpkg integrate install
+          ./vcpkg.exe integrate install
+
+      - name: Build
+        run: |
+          # Currently, Meson has a bug where the include paths need to be on the same drive as the project directory
+          # see https://github.com/mesonbuild/meson/issues/8329
```
cboesch-gpsw, line 48: please do not slip in changes in a code move: preserve the content of the rules exactly the same. These changes must be discussed separately

jmoguillansky-gpsw: I'm using the built-in vcpkg but it's on C drive.  The project is on D drive.  There's a bug in meson where the dependencies need to be on the same drive.

cboesch-gpsw: That's not what my remark is about: you're hiding functional changes in a move commit

jmoguillansky-gpsw: it's not only a move commit, see the commit message.  But I will split it up into atomic commits like you prefer

jbelaston-gpsw: +1
Yeah I think it's really difficult here to do the review of a code which has been moved and updated... :disappointed: 
So, please, here my proposal :slightly_smiling_face: :
- one commit only for the move (without any other change)
- then another one for the change

jbelaston-gpsw (changes requested):

jmoguillansky-gpsw:
closed, already merged in a different PR

PR #208, opened 2021-02-19
build: misc improvements

jmoguillansky-gpsw:
Here's a high level summary of the changes:
1.  set visual studio environment variables once in parent shell, instead of each time a windows executable is invoked. This accelerates and simplifies the build.  The user can either launch wsl from visual studio developer command prompt, or use the helper script provided
2.   set windows build environment variables once, in parent shell as well, instead of each time a command is invoked.  This simplifies the environment, and allows the user to invoke commands manually.  For example, to rebuild nodegl, they can type meson.exe compile -j8 builddir/libnodegl directly, no additional environment variables necessary, etc.  It also simplifies the build log
3.   invoke windows executables directly from a bash shell, instead of invoking from a command prompt, from a bash shell.
      This further allows us to align the code across platforms.
4.    use same code path across platforms
5.    add wildcard rules for compile and install
6.    added more fault tolerance if prefix target fails.
      Previously, make assumes prefix target is successful if the folder is created.
      But, prefix target may encounter an error after folder is created.
      Now, we add a hidden file when prefix target is completed.
7.   added more customization parameters:
      The user can select the builddir, meson backend
8.   fix quotation syntax issue when invoking windows command prompt
9.   set default build configuration as debug, not debugoptimized, to simplify debugging (edited)


jmoguillansky-gpsw:
Instead of invoking windows executables from a command prompt, from a bash shell, we can just invoke windows executables directly from bash shell.  This simplifies the code a lot.

Makefile
```diff
@@ -31,9 +31,6 @@ endif
 
 PYTHON_MAJOR = 3
 
-#
```
jmoguillansky-gpsw, line 31: The user-configurable params are in many other places as well, so I removed this comment

.github/workflows/ci_msvc.yml
```diff
@@ -25,25 +25,25 @@ jobs:
 
       - uses: actions/checkout@v2
 
+      - name: Restore vcpkg binary cache archives
+        id: restore_vcpkg_binary_cache
+        uses: actions/cache@v2
+        with:
+          path: C:\Users\runneradmin\AppData\Local\vcpkg\archives
+          # if you change the vcpkg install packages, please change this key using a version number
+          key: vcpkg-binary-cache-v0.0.1
```
jmoguillansky-gpsw: TODO: compute key based on the triplet, package list, and vcpkg commit ID

jmoguillansky-gpsw (commented):
Todo: Reword commit, don't pull pynodegl-utils  requirements

PR #209, opened 2021-02-19
ci: (linux) misc improvements

jmoguillansky-gpsw:
Use matrix to consolidate common code across jobs

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -10,6 +10,13 @@ jobs:
   linux:
 
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - backend: "all"
+            run_tests_command: "DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests"
+          - backend: "opengles 2"
+            run_tests_command: "MESA_GLES_VERSION_OVERRIDE=2.0 DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests TESTS_SUITE=opengles"
 
     steps:
     - uses: actions/checkout@v1
@@ -31,26 +38,7 @@ jobs:
     - name: Build
       run: |
         make -j$(($(nproc)+1))
-    - name: Run tests
+    - name: Run tests ${{matrix.backend}}
       run: |
-        DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests
+        ${{matrix.run_tests_command}}
 
-  gles2:
-
-    runs-on: ubuntu-20.04
-
-    steps:
-    - uses: actions/checkout@v1
-
-    - name: Install dependencies
-      run: |
-        sudo apt -y update
-        sudo apt -y install libsdl2-dev python3-venv
-        sudo apt -y install ffmpeg libavcodec-dev libavutil-dev libavformat-dev libavdevice-dev libavfilter-dev libswscale-dev libswresample-dev libpostproc-dev
-
-    - name: Build
-      run: |
-        make -j$(($(nproc)+1))
-    - name: Run tests with OpenGLES2
-      run: |
-        MESA_GLES_VERSION_OVERRIDE=2.0 DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests TESTS_SUITE=opengles


cboesch-gpsw (changes requested):
One change at a time please: the matrix change is completely unrelated to the addition of renderdoc

jmoguillansky-gpsw:
> One change at a time please: the matrix change is completely unrelated to the addition of renderdoc

updated, I will add renderdoc and ninja in the other PR when it's actually used

.github/workflows/ci_linux.yml
```diff
@@ -10,11 +10,18 @@ jobs:
...
+          - backend: "all"
+            run_tests_command: "DEBUG=yes DEBUG_GL=yes make tests"
+          - backend: "opengles"
+            run_tests_command: "MESA_GLES_VERSION_OVERRIDE=2.0 DEBUG=yes DEBUG_GL=yes make tests TESTS_SUITE=opengles"
 
     steps:
     - uses: actions/checkout@v1
 
-    - name: Install dependencies
+    - name: Install Dependencies
```
ubitux, line 24: This looks unrelated

.github/workflows/ci_linux.yml
```diff
@@ -26,31 +33,17 @@ jobs:
         #         apt-get install python3-venv
         #
         sudo apt -y install libsdl2-dev python3-venv
-        sudo apt -y install ffmpeg libavcodec-dev libavutil-dev libavformat-dev libavdevice-dev libavfilter-dev libswscale-dev libswresample-dev libpostproc-dev
+        sudo apt -y install ffmpeg libavcodec-dev libavutil-dev libavformat-dev libavdevice-dev \
+          libavfilter-dev libswscale-dev libswresample-dev libpostproc-dev
```
ubitux, line 37: This looks unrelated

.github/workflows/ci_linux.yml
```diff
@@ -26,31 +33,17 @@ jobs:
...
 
     - name: Build
       run: |
-        make -j$(($(nproc)+1))
-    - name: Run tests
+        export PATH=/home/runner/.local/bin:$PATH
+        export PKG_CONFIG_PATH=$PWD/venv/lib/pkgconfig
```
ubitux, line 42: This looks unrelated

.github/workflows/ci_linux.yml
```diff
@@ -26,31 +33,17 @@ jobs:
...
+        make
+    - name: Run Tests ${{matrix.backend}}
       run: |
-        DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests
+        export PATH=/home/runner/.local/bin:$PATH
+        export PKG_CONFIG_PATH=$PWD/venv/lib/pkgconfig
```
ubitux, line 47: This looks unrelated

ubitux (changes requested):

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: updated, I dropped this change for now

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: updated, I dropped this change for now

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: updated, I dropped this change for now

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: updated, I dropped this change for now

.github/workflows/ci_linux.yml
```diff
@@ -10,6 +10,13 @@ jobs:
   linux:
 
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - backend: "all"
+            run_tests_command: "DEBUG=yes DEBUG_GL=yes make tests"
+          - backend: "opengles"
```
ubitux, line 18: can you explicit the `2`?

.github/workflows/ci_linux.yml
```diff
@@ -10,6 +10,13 @@ jobs:
   linux:
 
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - backend: "all"
+            run_tests_command: "DEBUG=yes DEBUG_GL=yes make tests"
```
ubitux, line 17: you removed `-k -j$(($(nproc)+1))`

.github/workflows/ci_linux.yml
```diff
@@ -10,6 +10,13 @@ jobs:
...
+          - backend: "opengles"
+            run_tests_command: "MESA_GLES_VERSION_OVERRIDE=2.0 DEBUG=yes DEBUG_GL=yes make tests TESTS_SUITE=opengles"
```
ubitux, line 19: you removed `-k -j$(($(nproc)+1))`

ubitux (changes requested):
Please change the commit message to "ci/linux: simplify code using matrix"; we never use this "( ... )" form.

Also, again, please stop trying to slip in unrelated changes all the time, it delays the review process unnecessarily. Every unrelated change is a potential never ending separated discussion: leave them out.

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: sorry I don't understand the question

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: I will remove this change and we can discuss it separately.  

The motivation for these changes:
The reason for removing -j is because the log output gets scrambled
We can still build targets using multiple threads, but if we build multiple targets in parallel, it scrambles the log output
As far as removing the -k, I think we should catch the errors, not keep going?

.github/workflows/ci_linux.yml
ubitux: This is specifically testing gles**2**, not just gles

.github/workflows/ci_linux.yml
ubitux: > The reason for removing -j is because the log output gets scrambled

Yeah but:
1. it's unrelated
2. jobbed `make` is useful for parallel downloading of external and various other build exec
3. why would you do it only for linux and not all the other
4. it needs a discussion and explanation when doing such a change

> As far as removing the -k, I think we should catch the errors, not keep going?

Same as previously (unrelated, not exclusive to linux, separated discussion,)

`-k` does catch the error, but it still tries all code paths: the reason we have this is because the tests were previously running through make, and you wanted to run them all before reporting any error. This might not be relevant anymore so it might indeed be relevant to remove them... separately, with a dedicated explanation such as the one I just provided.

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: updated, I removed this change, we can discuss separately

.github/workflows/ci_linux.yml
```diff
@@ -31,26 +38,7 @@ jobs:
     - name: Build
       run: |
         make -j$(($(nproc)+1))
-    - name: Run tests
+    - name: Run tests ${{matrix.backend}}
       run: |
-        DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests
+        ${{ matrix.run_tests_command }}
```
ubitux, line 43: Can you be consistent with your use of `${{ X }}` vs `${{X}}`? Also please don't forget my comment about the commit message.

jmoguillansky-gpsw: thanks, updated

cboesch-gpsw (approved):

mbouron-gpsw (approved):

PR #213, opened 2021-02-22
Jbelaston/fix makefile implicit external download dependency

jbelaston-gpsw:
Quick fix of possible makefile error when running `make` with `-jx`
Be sure that `external-download` is well processed before `PREFIX` target.
More info in the commits description.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -214,13 +214,13 @@ else
 	($(ACTIVATE) && $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir/libnodegl)
 endif
 
-pkg-config-install: external-download $(PREFIX)
+pkg-config-install: $(PREFIX)
 ifeq ($(TARGET_OS),Windows)
 	($(ACTIVATE) \&\& $(MESON_SETUP) -Dtests=false external\\pkgconf builddir\\pkgconf \&\& $(MESON_COMPILE) -C builddir\\pkgconf \&\& $(MESON_INSTALL) -C builddir\\pkgconf)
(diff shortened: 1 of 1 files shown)


ubitux:
Isn't this only needed if you want vswhere to be downloaded by externals? Unless I'm mistaken, in the current state the venv doesn't require the dependencies to be downloaded.

jbelaston-gpsw:
> Isn't this only needed if you want vswhere to be downloaded by externals? Unless I'm mistaken, in the current state the venv doesn't require the dependencies to be downloaded.

This was indeed the first intent but after consideration i think it can be good to globally consider external-dowload step as a prerequisite for entering in the env rather than specify it twice (for sxplayer and pkg-congif). What do you think ? 
Or maybe your point is to make clear the real dependencies and not add implicit logic, and therefore, don't change the current behaviour ?

jbelaston-gpsw:
> > Isn't this only needed if you want vswhere to be downloaded by externals? Unless I'm mistaken, in the current state the venv doesn't require the dependencies to be downloaded.
> 
> This was indeed the first intent but after consideration i think it can be good to globally consider external-dowload step as a prerequisite for entering in the env rather than specify it twice (for sxplayer and pkg-congif). What do you think ?
> Or maybe your point is to make clear the real dependencies and not add implicit logic, and therefore, don't change the current behaviour ?

(seen together with @cboesch-gpsw)
Finally, there is no need to do this change since there is no direct dependency between PREFIX and external-download targets.
`sxplayer` and `pkg-config` need to be download to be used so the they both have to explicit their dependency to `external-download` as it is done today.
I close this PR.

PR #217, opened 2021-02-24
tools/ipc: use closesocket() instead of close() on Windows

mbouron-gpsw:
On Windows/MSVC, close() cannot be used on sockets.

This commits prevents ngl-ipc from exiting prematurely (with an error
code different from 0) and allows ngl-ipc to be used by the
pynodegl-utils hooks.

### DIFF
--- a/ngl-tools/ngl-ipc.c
+++ b/ngl-tools/ngl-ipc.c
@@ -382,6 +382,10 @@ end:
     ipc_pkt_freep(&s.send_pkt);
     ipc_pkt_freep(&s.recv_pkt);
     if (fd != -1)
+#ifdef _WIN32
+        closesocket(fd);
+#else
         close(fd);
+#endif
     return ret;
 }


jmoguillansky-gpsw (approved):

ngl-tools/ngl-ipc.c
```diff
@@ -382,6 +382,10 @@ int main(int argc, char *argv[])
     ipc_pkt_freep(&s.send_pkt);
     ipc_pkt_freep(&s.recv_pkt);
     if (fd != -1)
+#ifdef _WIN32
```
jmoguillansky-gpsw, line 385: should this be msvc (is this compatible with mingw)?

cboesch-gpsw (commented):
Patch looks fine, but how is it possible to have the program exiting with an error code? Is `close()` implementation actually calling `exit(1)`? (wtf?)

ngl-tools/ngl-ipc.c
mbouron-gpsw: It is compatible with mingw, we use the winsock API for both mingw and msvc (and ngl-desktop already use closesocket() instead of close()).

mbouron-gpsw:
> Patch looks fine, but how is it possible to have the program exiting with an error code? Is `close()` implementation actually calling `exit(1)`? (wtf?)

I'm not sure about what's happening under the hood but the winsock API does not provide an Unix file descriptor but a handle to a kernel object and close() only works on file descriptor.

cboesch-gpsw (approved):

PR #218, opened 2021-03-01
rtt: restore the proper current rendertarget after the subgraph has been rendered

mbouron-gpsw:


jmoguillansky-gpsw:
Please test this with all the backends

cboesch-gpsw (approved):

jmoguillansky-gpsw:
Even with render to texture, you shouldn't have to use LOAD_OP_LOAD.  
You can first render to texture, then render the texture to the screen.

jmoguillansky-gpsw:
"- the RTT 1 node detects that it needs to allocate two rendertargets,
  one with LOAD_OP=CLEAR and one with LOAD_OP=LOAD because there are
  two separate render passes (one for each render)"
  
  This terminology is confusing.  Let's use terminology consistent with Vulkan.
  A rendertarget doesn't have load ops and store ops, a renderpass does.
  A rendertarget is just the render destination (e.g. texture or framebuffer)?
  
  Also, we don't need multiple render passes because we have multiple render nodes, they can still share the same render pass.
  We need multiple render passes if they have different configurations, e.g. drawing to a different destination format, etc.

jmoguillansky-gpsw:
From my perspective I would prefer if we first optimize the renderpass logic before merging to master.

jmoguillansky-gpsw:
"add loach_attachment_nested test"
typo (load attachment)

mbouron-gpsw:
> From my perspective I would prefer if we first optimize the renderpass logic before merging to master.

This commit fixes a regression and needs to go to master. Optimizing the renderpass logic is another topic and won't be impacted by this commit.

mbouron-gpsw:
> "- the RTT 1 node detects that it needs to allocate two rendertargets,
> one with LOAD_OP=CLEAR and one with LOAD_OP=LOAD because there are
> two separate render passes (one for each render)"
> 
> This terminology is confusing. Let's use terminology consistent with Vulkan.

I used this terminology to reflect what the code do and our current abstraction use a rendertarget object (which mixes the notion of framebufer+renderpass). This abstraction is not nice but I plan to split it into framebuffer + renderpass (I delayed this split to avoid causing too much conflicts with the upcoming backends branches).

> A rendertarget doesn't have load ops and store ops, a renderpass does.
> A rendertarget is just the render destination (e.g. texture or framebuffer)?

At the low-level graphics level (Vulkan, Metal, DX) yes. The commit description is talking about the current abstraction (the rendertarget struct). I will reword the commit description to make it more explicit.

> 
> Also, we don't need multiple render passes because we have multiple render nodes, they can still share the same render pass.
> We need multiple render passes if they have different configurations, e.g. drawing to a different destination format, etc.

mbouron-gpsw:
> Also, we don't need multiple render passes because we have multiple render nodes, they can still share the same render pass.
> We need multiple render passes if they have different configurations, e.g. drawing to a different destination format, etc.

In this case we have multiple render pass because they have different load/store operations.

mbouron-gpsw:
> "add loach_attachment_nested test"
> typo (load attachment)

Fixed.

mbouron-gpsw:
The commit description has been updated with the following sentence added:
```
    The terminology "rendertarget" used in the following description refers
    the ngli rendertarget structure (which aggregates a framebuffer and a
    renderpass and their respective configurations).
```

jmoguillansky-gpsw (approved):

mbouron-gpsw:
> Even with render to texture, you shouldn't have to use LOAD_OP_LOAD.
> You can first render to texture, then render the texture to the screen.

In theory yes but unfortunately the main user of nodegl is using this pattern. We can probably reorder internally the the graphics operations to avoid this scenario but that will require some work. IMHO, the right solution is to enforce this at the API level (while this will require probably less work on our side, it will require a lot of work in the codebase of our main user).

PR #219, opened 2021-03-02
utils: remove the use of a dedicated process to handle scripts live changes

mbouron-gpsw:


pynodegl-utils/pynodegl_utils/scriptsmgr.py
```diff
@@ -63,18 +69,32 @@ def resume(self):
...
+
     @QtCore.Slot()
     def reload(self):
         self.pause()
+        self._set_reloading()
         odict = query_subproc(query='list', pkg=self._module_pkgname)
         if 'error' in odict:
             self.update_filelist(odict['filelist'])
             self.resume()
+            self._set_reloaded()
```
cboesch-gpsw, line 97: These 2 are swapped compared to below. Is this on purpose?

pynodegl-utils/pynodegl_utils/scriptsmgr.py
```diff
@@ -50,6 +50,12 @@ def __init__(self, module_pkgname):
         self._observer.start()
         self._timer = None
 
+        self._lock = QtCore.QMutex()
+        self._query_count = 0
+        self._query_cond = QtCore.QWaitCondition()
+        self._reloading = False
+        self._reloaded_cond = QtCore.QWaitCondition()
```
cboesch-gpsw, line 59: Would it make sense to rename `_query_cond` and `_reloaded_cond` to respectively `_query_count_cond` and `_reloading_cond` so that we know what they are associated with precisely?

pynodegl-utils/pynodegl_utils/filetracker.py
```diff
@@ -56,10 +57,13 @@ def _get_trackable_files(self):
         return files
 
     def start_hooking(self):
+        self._start_modules = set(sys.modules.keys())
```
cboesch-gpsw, line 60: not a fan of the implicit convert from `dict_keys` to `set`; can you explicit the `set()` here as well?

pynodegl-utils/pynodegl_utils/filetracker.py
```diff
@@ -30,6 +30,7 @@ class FileTracker:
 
     def __init__(self):
         self.filelist = set()
+        self.modulelist = set()
```
cboesch-gpsw, line 33: It's really confusing to have a module list managed by the *file* tracker. Can you rename the module to something more appropriate? `ResourceTracker`, `StateTracker`, or simply `Tracker`. Or anything else really, but not "File".

pynodegl-utils/pynodegl_utils/scriptsmgr.py
```diff
@@ -125,6 +132,12 @@ def set_filelist(self, filelist):
         self._files_to_watch = set(filelist)
         self._update_dirs_to_watch()
 
+    def update_modulelist(self, modulelist):
+        self._modules.update(modulelist)
+
+    def set_modulelist(self, modulelist):
+        self._modules = set(modulelist)
```
cboesch-gpsw, line 139: it's already a `set()`, right?

pynodegl-utils/pynodegl_utils/ui/main_window.py
```diff
@@ -144,20 +144,23 @@ def _get_scene(self, **cfg_overrides):
...
             self._scripts_mgr.update_filelist(ret['filelist'])
+            self._scripts_mgr.update_modulelist(ret['modulelist'])
             self._scripts_mgr.resume()
             self._scripts_mgr.dec_query_count()
             self.error.emit(ret['error'])
             return None
 
         self.error.emit(None)
-        self._scripts_mgr.set_filelist(ret['filelist'])
+        self._scripts_mgr.update_filelist(ret['filelist'])
```
cboesch-gpsw, line 162: Why this change? Can it be extracted earlier?

pynodegl-utils/pynodegl_utils/hooks.py
```diff
@@ -197,18 +198,32 @@ def run(self):
...
+                with tempfile.NamedTemporaryFile('w', prefix='ngl_scene_', suffix='.ngl', delete=True) as f:
+                    f.write(serialized_scene)
+                    f.flush()
+                    self.sendingScene.emit(session_id, self._scene_id)
+                    try:
+                        self._hooks_caller.scene_change(session_id, f.name, cfg)
+                    except Exception as e:
+                        self.error.emit(session_id, 'Error while sending scene: %s' % str(e))
+                        return
+                    self.done.emit(session_id, self._scene_id, time.time() - start_time)
```
cboesch-gpsw, line 226: Can you move the following in a dedicated method that take `fname` as argument?
```python
                    self.sendingScene.emit(session_id, self._scene_id)
                    try:
                        self._hooks_caller.scene_change(session_id, fname, cfg)
                    except Exception as e:
                        self.error.emit(session_id, 'Error while sending scene: %s' % str(e))
                        return
                    self.done.emit(session_id, self._scene_id, time.time() - start_time)
```

This will prevent desync between the two blocks, typically you forgot to include `self.sendingScene.emit(...)`  in the Windows path (or is it on purpose?)

Makefile
```diff
@@ -164,7 +164,7 @@ endif
 # version), and it would fail anyway because pynodegl is currently not
 # available on PyPi.
 #
-# We do not pull the requirements on Windows because of various issues:
+# We do not pull the requirements on Windows/MinGW because of various issues:
```
cboesch-gpsw, line 167: "Windows (for both MSVC and MinGW)"

cboesch-gpsw (changes requested):

pynodegl-utils/pynodegl_utils/filetracker.py
```diff
@@ -56,10 +57,13 @@ def _get_trackable_files(self):
         return files
 
     def start_hooking(self):
+        self._start_modules = set(sys.modules.keys())
         self._start_files = self._get_trackable_files()
         builtins.open = self._open_hook
 
     def end_hooking(self):
+        new_modules = sys.modules.keys() - self._start_modules
```
cboesch-gpsw, line 65: not a fan of the implicit convert from `dict_keys` to `set`; can you explicit the `set()` here as well?

pynodegl-utils/pynodegl_utils/scriptsmgr.py
mbouron-gpsw: Ok for `_query_count_cond`, regarding `_reloaded_cond` it makes more sense to me to call it this way. You wait for the reloaded condition.

pynodegl-utils/pynodegl_utils/scriptsmgr.py
mbouron-gpsw: I followed what was done in set_filelist(). It supposed to be `set()` but it does not prevent someone to call it with a list.

pynodegl-utils/pynodegl_utils/ui/main_window.py
mbouron-gpsw: This code is executed in parralel and the query does not rely on a subprocess anymore meaning that if the underlying script has already been loaded, its file list will be empty.

Makefile
mbouron-gpsw: The issues described bellow are only on MinGW (except the one described for ngl-control).

cboesch-gpsw: Ah. Maybe explicit MinGW only then?

pynodegl-utils/pynodegl_utils/com.py
```diff
@@ -32,6 +32,7 @@
 import pynodegl as ngl
 from pynodegl_utils.filetracker import FileTracker
 
+from PySide2 import QtCore
```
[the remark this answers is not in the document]
mbouron: Fixed.

pynodegl-utils/pynodegl_utils/hooks.py
mbouron: Done.

pynodegl-utils/pynodegl_utils/filetracker.py
mbouron: Fixed.

Makefile
mbouron: Ok. Updated.

pynodegl-utils/pynodegl_utils/scriptsmgr.py
mbouron-gpsw: Fixed.

pynodegl-utils/pynodegl_utils/scriptsmgr.py
mbouron-gpsw: Actually renamed both. I don't really care (even though it makes more sense to me to use `_reloaded_cond`).

pynodegl-utils/pynodegl_utils/filetracker.py
mbouron-gpsw: Done in a separate commit.

cboesch-gpsw (approved):

PR #222, opened 2021-03-04
vk: add gpu capture support via renderdoc

jmoguillansky-gpsw:
we may need to merge the windows build support first...
we may want to consolidate common code for initializing the gpu capture ctx


mbouron-gpsw:
The commit has been merged to the Vulkan branch (after the branch has been rebased on latest master)

PR #224, opened 2021-03-09
Improve ngl-control responsiveness by not waiting the hooks from the gui thread

mbouron-gpsw:


mbouron-gpsw:
> From my perspective: the UI should always be responsive, never frozen, but we should wait until the hook finished processing before pushing another event to the queue. So for example, when a user selects a different scene, the UI can show a progress bar or something, but not let the user select a different scene again until the current event is complete (or canceled)

The issue affects both scene changes and scene parameter changes. Scene parameters can be controlled by a slider and the user might want to use it to quickly increases a value. IMHO, in this particular scenario we don't want to disable the entire UI (with a progress bar or not) while waiting for the hooks to have finished execution.
Moreover, I think we also don't want to wait for all the hooks to be finished because sending a scene with its medias to an Android or iOS device can take quite some time.

This PR is a direct improvement over the current behavior (where the UI thread is frozen because it waits for the previous commands to be finished causing sometimes deadlocks, and crashes on Wayland). It makes sense to me to merge it (if the code is OK) as it fixes issues and makes ngl-control a bit more usable. 

An improvement over the behavior this PR introduces would be to only process the last received command (and not all the accumulated commands until that point).

jmoguillansky-gpsw:
> Moreover, I think we also don't want to wait for all the hooks to be finished because sending a scene with its medias to an Android or iOS device can take quite some time.

Will this cause any stability issues?  How does the high-level logic work?  Do you push scene change events to a queue and the ngl-desktop app pulls events from a queue?  

> This PR is a direct improvement over the current behavior (where the UI thread is frozen because it waits for the previous commands to be finished causing sometimes deadlocks, and crashes on Wayland). 

Why is there deadlock: because the hook never finishes, and the UI is stuck waiting?  

In general we never want a UI that's frozen but we also don't want a UI that's responsive but stuck in a bad state.  
If the user selects a scene change, we should probably either wait for the current scene change event to finish, or cancel the current scene change event.  If you want to push events to a queue, we should probably put a size limit on the queue and block until there's space available on the queue when it's full.

pynodegl-utils/pynodegl_utils/hooks.py
```diff
@@ -234,28 +239,34 @@ def __init__(self, get_scene_func, hooks_view, hooks_caller):
...
         self._threads = []
 
     def process(self, module_name, scene_name):
-        self._wait_threads()  # Wait previous batch
         data = self._hooks_view.get_data_from_model()
-        for session_id, data_row in data.items():
+        for i, (session_id, data_row) in enumerate(data.items()):
+            if i >= len(self._threads):
+                thread = QtCore.QThread()
+                thread.start()
```
jmoguillansky-gpsw, line 255: is there a cap on the number of threads you create?
Can we improve performance using a threadpool?

pynodegl-utils/pynodegl_utils/hooks.py
```diff
@@ -234,28 +239,34 @@ def __init__(self, get_scene_func, hooks_view, hooks_caller):
...
-                                       session_id, data_row['backend'], data_row['system'],
-                                       module_name, scene_name)
-            self._threads.append(hook_thread)
-            hook_thread.uploadingFile.connect(self._hooks_uploading)
-            hook_thread.buildingScene.connect(self._hooks_building_scene)
-            hook_thread.sendingScene.connect(self._hooks_sending_scene)
-            hook_thread.done.connect(self._hooks_done)
-            hook_thread.error.connect(self._hooks_error)
-            hook_thread.start()
+            worker = _HooksWorker(self._get_scene_func, self._hooks_caller,
```
jmoguillansky-gpsw, line 259: is there a limit on the number of workers you create?

pynodegl-utils/pynodegl_utils/hooks.py
```diff
@@ -234,28 +239,34 @@ def __init__(self, get_scene_func, hooks_view, hooks_caller):
...
+                                  session_id, data_row['backend'], data_row['system'],
+                                  module_name, scene_name)
+            worker.uploadingFile.connect(self._hooks_uploading)
+            worker.buildingScene.connect(self._hooks_building_scene)
+            worker.sendingScene.connect(self._hooks_sending_scene)
+            worker.done.connect(self._hooks_done)
+            worker.error.connect(self._hooks_error)
+            self._workers.append(worker)
+            worker.moveToThread(self._threads[i])
```
jmoguillansky-gpsw, line 268: each worker gets its own thread.  Will the events be processed in the same order the user selects?

pynodegl-utils/pynodegl_utils/hooks.py
mbouron-gpsw: The number of threads used for the hooks is equal to the number of hooks registered. One thread is a dedicated to a particular hook. There is no need for a threadpool here, the threads have the same lifetime as the application.

pynodegl-utils/pynodegl_utils/hooks.py
mbouron-gpsw: There is currently no limit. The current number worker depends on how many changes has made by the user.

pynodegl-utils/pynodegl_utils/hooks.py
mbouron-gpsw: Each hook has his own dedicated thread and for each threads we dispatch a new worker if the hook is not disabled. The events are processed in the same order the user selects.
For example, if the user has registered two ngl-desktop instances and has enabled the android hooks with 2 devices attached, 4 threads will be allocated when the HookController is instantiated. Then, each time the user changes a parameter or the scene, a new worker is dispatched to each threads (with the user changes taken into account).

cboesch-gpsw (approved):

pynodegl-utils/pynodegl_utils/hooks.py
jmoguillansky-gpsw: ok, so the number of threads won't grow over time, if the event queue grows over time?  
That's good.  My main overall concern is since we're switching from synchronous to async, and single-threaded to multi-threaded, are there any stability issues

pynodegl-utils/pynodegl_utils/hooks.py
jmoguillansky-gpsw: it might be helpful to set a cap, otherwise are there long term stability issues, where the queue might keep growing over time?

jmoguillansky-gpsw:
so previously, when the user clicks a scene change event, the UI would freeze, and wait until the client finishes.  Now the user can click 20 different scene change events, and the UI will respond right away, but it might be 20 minutes until the client actually processes the scene change (the ui and client are totally out of sync), and a lot of unnecessary intermediate events are triggered.  IMO better for the UI to show a dialog, wait for the event to complete, then let the user trigger another scene change, but looks like you guys went ahead and merged this code anyways

PR #225, opened 2021-03-09
build: misc improvements part 2

jmoguillansky-gpsw:


jmoguillansky-gpsw:
In the existing code you call meson compile without -j, does that use mulithreading?

wmacdonald-gpsw (approved):
LGTM

Makefile
```diff
@@ -87,9 +87,9 @@ endif
 # syntax) for the issue described here:
 # https://github.com/ninja-build/ninja/issues/1139#issuecomment-724061270
 ifeq ($(TARGET_OS),Windows)
-MESON_COMPILE = meson compile
+MESON_COMPILE = meson compile -j8
 else
-MESON_COMPILE = MAKEFLAGS= meson compile
+MESON_COMPILE = MAKEFLAGS= meson compile -j8
```
mbouron-gpsw, line 92: -j should not be hardcoded to an arbitrary value.

.github/workflows/ci_linux.yml
```diff
@@ -14,9 +14,9 @@ jobs:
       matrix:
         include:
           - backend: "all"
-            run_tests_command: "DEBUG=yes DEBUG_GL=yes make -k -j$(($(nproc)+1)) tests"
+            run_tests_command: "DEBUG=yes DEBUG_GL=yes make tests"
```
mbouron-gpsw, line 17: Please remove the period at the end of the commit message and reword the commit description to something like:
```The -k parameter was used when the tests suite was written in Make. It allowed all the tests to be ran even if an error occurred. It is not useful anymore and can be removed since we are now relying on Meson (which does not stop tests execution on error).
The -j parameter can also be removed as Meson (using the ninja backend) runs the tests in parallel.

File: .github/workflows/ci_linux.yml [hunk 3]
```diff
@@ -37,7 +37,7 @@ jobs:
 
     - name: Build
       run: |
-        make -j$(($(nproc)+1)) DEBUG_GPU_CAPTURE=yes
+        make DEBUG_GPU_CAPTURE=yes
```
mbouron-gpsw, line 40: I'm sure that we want to drop the -j parameter for two reasons:
- it allows parallel download of our dependency
- it tests that there is no dependency error in the Makefile

Makefile
jmoguillansky-gpsw: I believe we can omit this param as meson compile will use multithreading by default?

.github/workflows/ci_linux.yml
jmoguillansky-gpsw: That's a good point, but it can also cause issues.
Let me give an example.
Suppose nodegl-install depends on sxplayer-install and shaderc-install.
Then sxplayer and shaderc will be built in parallel, and the log output will be scrambled.

mbouron-gpsw (approved):

PR #229, opened 2021-03-10
build: misc part 3

jmoguillansky-gpsw:
Add logic to indicate when prefix target actually completes.
This improves error handling.  Currently make thinks prefix target
is completed when the folder is created.
However, the prefix target may fail after the folder is created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,7 @@
 #
 
 PREFIX          ?= venv
+PREFIX_DONE     = .venv-done
 ifeq ($(TARGET_OS),Windows)
 PREFIX_FULLPATH = $(shell wslpath -wa .)\$(PREFIX)
 else
@@ -201,7 +202,7 @@ else
 	($(ACTIVATE) && PKG_CONFIG_PATH=$(PREFIX_FULLPATH)/lib/pkgconfig LDFLAGS=$(RPATH_LDFLAGS) pip -v install -e ./pynodegl)
 endif
 
-pynodegl-deps-install: $(PREFIX) nodegl-install
+pynodegl-deps-install: $(PREFIX_DONE) nodegl-install
 ifeq ($(TARGET_OS),Windows)
 	($(ACTIVATE) \&\& pip install -r pynodegl\\requirements.txt)
 else
@@ -229,7 +230,7 @@ else
 	($(ACTIVATE) && $(MESON_SETUP) $(NODEGL_DEBUG_OPTS) libnodegl builddir/libnodegl)
 endif
 
-pkg-config-install: external-download $(PREFIX)
+pkg-config-install: external-download $(PREFIX_DONE)
 ifeq ($(TARGET_OS),Windows)
 	($(ACTIVATE) \&\& $(MESON_SETUP) -Dtests=false external\\pkgconf builddir\\pkgconf \&\& $(MESON_COMPILE) -C builddir\\pkgconf \&\& $(MESON_INSTALL) -C builddir\\pkgconf)
 	($(CMD) copy "$(PREFIX_FULLPATH)\\Scripts\\pkgconf.exe" "$(PREFIX_FULLPATH)\\Scripts\\pkg-config.exe")
@@ -256,16 +257,19 @@ external-download:
 # We do not pull meson from pip on MinGW for the same reasons we don't pull
 # Pillow and PySide2. We require the users to have it on their system.
 #
-$(PREFIX):
+$(PREFIX_DONE):
 ifeq ($(TARGET_OS),Windows)
-	($(CMD) $(PYTHON) -m venv "$@")
+	($(CMD) $(PYTHON) -m venv $(PREFIX))
 	($(ACTIVATE) \&\& pip install meson ninja)
 else ifeq ($(TARGET_OS),MinGW-w64)
-	$(PYTHON) -m venv --system-site-packages $@
+	$(PYTHON) -m venv --system-site-packages $(PREFIX)
 else
-	$(PYTHON) -m venv $@
+	$(PYTHON) -m venv $(PREFIX)
 	($(ACTIVATE) && pip install meson ninja)
 endif
+	touch $@
+
+$(PREFIX): $(PREFIX_DONE)
 
 tests: nodegl-tests tests-setup
 ifeq ($(TARGET_OS),Windows)
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 builddir
 venv/
+.venv-done


wmacdonald-gpsw (approved):
Looks Good

mbouron-gpsw:
Can you mention that .DELETE_ON_ERROR cannot be used here because it does not work with directories and reference http://savannah.gnu.org/bugs/?func=detailitem&item_id=16372 ?

jmoguillansky-gpsw:
> Can you mention that .DELETE_ON_ERROR cannot be used here because it does not work with directories and reference http://savannah.gnu.org/bugs/?func=detailitem&item_id=16372 ?

updated

Makefile
```diff
@@ -256,16 +257,19 @@ external-download:
...
 	($(ACTIVATE) \&\& pip install meson ninja)
 else ifeq ($(TARGET_OS),MinGW-w64)
-	$(PYTHON) -m venv --system-site-packages $@
+	$(PYTHON) -m venv --system-site-packages $(PREFIX)
 else
-	$(PYTHON) -m venv $@
+	$(PYTHON) -m venv $(PREFIX)
 	($(ACTIVATE) && pip install meson ninja)
 endif
+	touch $(PREFIX_DONE)
```
cboesch-gpsw, line 270: `touch $@`

cboesch-gpsw (commented):
This should be fine but I'm somehow concerned about the added complexity. I'm not necessarily going to oppose to it but I'd I have 2 questions first:
- What kind of errors were you expecting that this commit worked around?
- Are you sure a partial state for the venv is actually going to work properly at the next run? Like, let's say the venv creation succeed but the `pip install meson` fails for whatever reason, leading to a `venv` ending up in a shitty state: at the next run run, do you think `pip` is going to be able to handle the current broken state? Alternatively, wouldn't it make sense to instead add a `fclean`/`fullclean`/`distclean` rule the user can use to trash the venv to start again from scratch if something failed?

Again, no strong opinion from me yet, just trying to understand what we're exactly trying to solve here.

jmoguillansky-gpsw:
> This should be fine but I'm somehow concerned about the added complexity. I'm not necessarily going to oppose to it but I have 2 questions first:
> 
> * What kind of errors were you expecting that this commit worked around?
> * Are you sure a partial state for the venv is actually going to work properly at the next run? Like, let's say the `venv` creation succeeds but the `pip install meson` fails for whatever reason, leading to a `venv` ending up in a shitty state: at the next run run, do you think `pip` is going to be able to handle the current broken state? Alternatively, wouldn't it make sense to instead add a `fclean`/`fullclean`/`distclean` rule the user can use to trash the venv to start again from scratch if something failed?
> 
> Again, no strong opinion from me yet, just trying to understand what we're exactly trying to solve here.

So the issue I was getting is the prefix target would fail, but on subsequent call to make , it would think prefix target already completed (because the folder was already created), and continue onto the next target, when prefix target hasn't finished yet.  It can fail for a variety of reasons: the remote server is done when downloading a package from URL, etc.  
So this makes the build more resilient.  
With this new logic: I check if .venv-done file exists, and this file is created only after prefix target completed.  I've been using this new logic for months now, no issues...

cboesch-gpsw:
> So the issue I was getting is the prefix target would fail, but on subsequent call to make , it would think prefix target already completed (because the folder was already created), and continue onto the next target, when prefix target hasn't finished yet. It can fail for a variety of reasons: the remote server is done when downloading a package from URL, etc.
> So this makes the build more resilient.
> With this new logic: I check if .venv-done file exists, and this file is created only after prefix target completed. I've been using this new logic for months now, no issues...

Yeah I understood that part, the commit description explains it.

> With regards to partial state working properly at next run: if pip install meson fails, then the prefix target will fail and .venv-done file won't be created, so this fix is more resilient? Previously, if pip install meson fails, make would continue on next run, thinking venv target passed?

Well, my point is that if `pip` fails and leave the `venv` in a broken state (like if `pip` tracebacks during an install because you `C-c` at the wrong time or whatever), the `venv` will contain partial/altered/broken state, which the next run will re-use and may not succeed at (just like today). So my question was whether the issues you had were into that realm or unrelated, because if they were, you wouldn't solve anything. AFAIU it's for download errors where the the venv wasn't yet altered, so for these cases (probably common?) I agree it can make sense. If you could mention in the commit description that specific scenario (failure at downloading dependencies) if that's indeed what you're trying to work around, it should be fine with me.

jmoguillansky-gpsw:
> > So the issue I was getting is the prefix target would fail, but on subsequent call to make , it would think prefix target already completed (because the folder was already created), and continue onto the next target, when prefix target hasn't finished yet. It can fail for a variety of reasons: the remote server is done when downloading a package from URL, etc.
> > So this makes the build more resilient.
> > With this new logic: I check if .venv-done file exists, and this file is created only after prefix target completed. I've been using this new logic for months now, no issues...
> 
> Yeah I understood that part, the commit description explains it.
> 
> > With regards to partial state working properly at next run: if pip install meson fails, then the prefix target will fail and .venv-done file won't be created, so this fix is more resilient? Previously, if pip install meson fails, make would continue on next run, thinking venv target passed?
> 
> Well, my point is that if `pip` fails and leave the `venv` in a broken state (like if `pip` tracebacks during an install because you `C-c` at the wrong time or whatever), the `venv` will contain partial/altered/broken state, which the next run will re-use and may not succeed at (just like today). So my question was whether the issues you had were into that realm or unrelated, because if they were, you wouldn't solve anything. AFAIU it's for download errors where the the venv wasn't yet altered, so for these cases (probably common?) I agree it can make sense. If you could mention in the commit description that specific scenario (failure at downloading dependencies) if that's indeed what you're trying to work around, it should be fine with me.

Originally I had git clean -fxd in the rule as well (at the beginning of the rule), so it handled that scenario as well.  
It may not be necessary to nuke the venv folder and start over, but there's more guarantee.  I will update the commit message...

jmoguillansky-gpsw:
> > > So the issue I was getting is the prefix target would fail, but on subsequent call to make , it would think prefix target already completed (because the folder was already created), and continue onto the next target, when prefix target hasn't finished yet. It can fail for a variety of reasons: the remote server is done when downloading a package from URL, etc.
> > > So this makes the build more resilient.
> > > With this new logic: I check if .venv-done file exists, and this file is created only after prefix target completed. I've been using this new logic for months now, no issues...
> > 
> > 
> > Yeah I understood that part, the commit description explains it.
> > > With regards to partial state working properly at next run: if pip install meson fails, then the prefix target will fail and .venv-done file won't be created, so this fix is more resilient? Previously, if pip install meson fails, make would continue on next run, thinking venv target passed?
> > 
> > 
> > Well, my point is that if `pip` fails and leave the `venv` in a broken state (like if `pip` tracebacks during an install because you `C-c` at the wrong time or whatever), the `venv` will contain partial/altered/broken state, which the next run will re-use and may not succeed at (just like today). So my question was whether the issues you had were into that realm or unrelated, because if they were, you wouldn't solve anything. AFAIU it's for download errors where the the venv wasn't yet altered, so for these cases (probably common?) I agree it can make sense. If you could mention in the commit description that specific scenario (failure at downloading dependencies) if that's indeed what you're trying to work around, it should be fine with me.
> 
> Originally I had git clean -fxd in the rule as well (at the beginning of the rule), so it handled that scenario as well.
> It may not be necessary to nuke the venv folder and start over, but there's more guarantee. I will update the commit message...

I updated the commit message, mentioned the specific scenario as an example.  Let me know if its ok now (or please reword the commit message however you prefer)

Makefile
jmoguillansky-gpsw: updated

mbouron-gpsw:
I've updated the commit with small adjustments made to the commit description (re-indent, Add -> Adds, Reference -> See).

mbouron-gpsw (approved):

jmoguillansky-gpsw:
> I've updated the commit with small adjustments made to the commit description (re-indent, Add -> Adds, Reference -> See).

Thanks so much, appreciate it!

PR #236, opened 2021-03-12
ci_mingw: build vulkan backend

mbouron-gpsw:


.github/workflows/ci_mingw.yml
```diff
@@ -19,13 +19,22 @@ jobs:
           msystem: MINGW64 # Start a 64 bit Mingw environment
           update: true
 
+
```
cboesch-gpsw, line 22: unrelated

doc/howto/installation.md
```diff
@@ -21,6 +21,7 @@ building and running the complete `node.gl` stack.
 
 - Install [MSYS2](https://www.msys2.org/) (which also brings **MinGW64**)
 - Install dependencies via pacman using **MinGW64** shell (*Not* MSYS2,
+- Install [Vulkan SDK](https://vulkan.lunarg.com/sdk/home#windows)
```
cboesch-gpsw, line 24: "Optionally"?

.github/workflows/ci_mingw.yml
```diff
@@ -19,13 +19,22 @@ jobs:
...
           pacman -S --noconfirm --needed mingw-w64-x86_64-{toolchain,ffmpeg,python}
           pacman -S --noconfirm --needed mingw-w64-x86_64-python3-{pillow,pip}
           pacman -S --noconfirm --needed mingw-w64-x86_64-meson
+          pacman -S --noconfirm --needed mingw-w64-x86_64-shaderc
+
+      - name: Install Vulkan SDK
+        shell: powershell
+        run: |
+          $env:CHERE_INVOKING = 'yes'  # Preserve the current working directory
+          C:\msys64\usr\bin\bash -lc "wget https://sdk.lunarg.com/sdk/download/1.2.162.1/windows/VulkanSDK-1.2.162.1-Installer.exe"
```
cboesch-gpsw, line 36: Could we at least download it in `external/Makefile` (not a fan of scattered specific version of dependencies)?

.github/workflows/ci_mingw.yml
```diff
@@ -20,13 +20,14 @@ jobs:
...
-          C:\msys64\usr\bin\bash -lc "pacman -S --noconfirm --needed mingw-w64-x86_64-meson"
+          pacman -S --noconfirm --needed git make
+          pacman -S --noconfirm --needed mingw-w64-x86_64-{toolchain,ffmpeg,python}
+          pacman -S --noconfirm --needed mingw-w64-x86_64-python3-{pillow,pip}
+          pacman -S --noconfirm --needed mingw-w64-x86_64-meson
 
       - name: Build
+        shell: msys2 {0}
         run: |
-          $env:CHERE_INVOKING = 'yes'  # Preserve the current working directory
```
cboesch-gpsw, line 31: Don't need that anymore? Interesting.

This whole simplification is nice BTW, thanks.

.github/workflows/ci_mingw.yml
mbouron-gpsw: Fixed.

doc/howto/installation.md
mbouron-gpsw: Done by adding (optional) at this end of the sentence (to make it consitence with MSVC).

.github/workflows/ci_mingw.yml
mbouron-gpsw: We already do that for MSVC. IMHO, the Vulkan SDK should be treated like the other dependencies that are not managed by vcpkg on Windows (Python, graphviz), having it *only* downloaded by the Makefile does not bring much value to me.

.github/workflows/ci_mingw.yml
mbouron-gpsw: It's not needed anymore, the msys2 shell is executed in the right directory.

.github/workflows/ci_mingw.yml
cboesch-gpsw: Ah I just realized this is for the Vulkan branch... We'll need to address that before merging to master.

I see value in having it in the external:
- less clumsy manual user labor (go to a random website, look for <something>, download it, double click install, follow instruction, ...)
- we would control the version in a single place: by having it in the CI I'm pretty sure we'll never bump it: together with the other dependencies reduce the risk when we decide to make a bump

It's special because it's system wide though, but at least pre-downloading it is a huge improvement.

cboesch-gpsw (approved):

PR #241, opened 2021-03-16
doc/project: rework and complete developer guidelines

cboesch-gpsw:
Best consulted here: https://github.com/gopro/gopro-lib-node.gl/blob/cboesch/doc-dev-guidelines/doc/project/developers.md

### DIFF
--- a/doc/project/developers.md
+++ b/doc/project/developers.md
@@ -1,26 +1,112 @@
 Developer guidelines
 ====================
 
+## Pull-request checklist
+
+### Style/cosmetics
+
+Cosmetics inconsistencies and cruft are the first things we notice, adding
+noise and distraction during the review, and often causing unnecessary review
+iterations. Make sure it's neat and clean before we can tackle more important
+topics:
+
+- Is your **coding style** following exactly what you observe within the files
+  you're modifying?
+- Do you have any **trailing whitespaces** or **missing end of file line
+  break**? (this is usually spotted by Git or a well configured editor)
+- Make sure there is **no stray change** unrelated to your commits
+- Whitespace/indent/typo/style/etc issue fixes belong in **dedicated commits**.
+
+### Functionality
+
+Some simple checks to greatly improve the basic functionality of your
+submission:
+
+- Make sure the build and **tests pass for every commits**: `git rebase master
+  --exec "make clean && make test"`
+- Can the same changes be **applied elsewhere** in the code base?
+- Avoid at all cost **glob-like code** such as `import *` or `*.ext` file
+  references: this kind of code leads to dangling references because they break
+  the obvious links between components
+- Double check **anything that can reference** the part you're making
+  modifications to (including the documentation); `git grep -i` on the file
+  names, symbols and keywords is your friend
+- Check the **imports**, **includes**, **dependencies**: are they still needed
+  after your changes?
+- Do you check and raise all possible **errors**?
+- Have you checked for potential **memory leaks**, in particular in error code
+  paths?
+- Do you **assert in impossible code paths** to help the compilers, tools and
+  developers make assumptions about the code?
+- Are the files you changed now in a **better state** than they were before?
+- Make sure you **understand deeply** and can **explain** every changes
+
+### Design
+
+Now is the time to take a step back:
+
+- Have you considered the impact of your changes from a **user perspective**?
+- Can your changes be made **simpler**? [Keep It Simple, Stupid][kiss]
+- Try to **avoid external dependencies**, especially for core features
+- If you are altering the **architecture and internal dependencies** between
+  components, is it clarifying / following the current structure or entangling
+  it?
+- What impact your changes have on the **maintainability**?
+
+### Exposure
+
 In order to keep the API maintainable in the long term, we try to keep it as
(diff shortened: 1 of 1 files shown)


doc/project/developers.md
```diff
@@ -1,26 +1,112 @@
...
+## Pull-request checklist
+
+### Style/cosmetics
+
+Cosmetics inconsistencies and cruft are the first things we notice, adding
+noise and distraction during the review, and often causing unnecessary review
+iterations. Make sure it's neat and clean before we can tackle more important
+topics:
+
+- Is your **coding style** following exactly what you observe within the files
```
jmoguillansky-gpsw, line 13: From my perspective, it's the project owner's responsibility to document the coding style.  Asking the contributor to infer the coding style from existing code is not sufficient IMO.  
As I've suggested before, I think it would benefit the project to adopt an existing coding style, such as GNU, and make additional changes on top if necessary, and more importantly adopt a automated style tool such as clang-format.  I've tested it in other projects and it works well.  It's also used in various open source projects such as RenderDoc.  It will speed up the PR and collaboration process.

doc/project/developers.md
```diff
@@ -1,26 +1,112 @@
...
+
+### Functionality
+
+Some simple checks to greatly improve the basic functionality of your
+submission:
+
+- Make sure the build and **tests pass for every commits**: `git rebase master
+  --exec "make clean && make test"`
+- Can the same changes be **applied elsewhere** in the code base?
+- Avoid at all cost **glob-like code** such as `import *` or `*.ext` file
```
jmoguillansky-gpsw, line 28: Not sure what's the issue with import *, it seems to be used commonly in production code?
If you're absolutely sure that you want to include all the modules in a python file, why not import *, it's much simpler?
You want to have more fine-grained control of which modules from the file are imported?

doc/project/developers.md
```diff
@@ -1,26 +1,112 @@
...
+- Do not insert a value in the middle of a public enum as it breaks ABI
+- Do not remove or change a public prototype as it breaks API and ABI
+
+### Integration
+
+This part is where we focus on the finalization of the work to reach production
+quality:
+
+- Are your commits **atomics**?
+- Check that your commit messages use a **standard prefix** (look at the Git
```
jmoguillansky-gpsw, line 73: Again: I think this should be documented.  The contributor should not be expected to infer this from the log.

cboesch-gpsw: Why? "When in Rome, do as the Romans do". It's part of the process of getting familiar with how a project works, you look around how things are done and copy it. The history is clean enough that a `git log` will give you the prefix pretty quickly.

We could maintain a list of all the prefixes we usually use, but tbh it's context specific and we'll have to change that list regularly.

doc/project/developers.md
cboesch-gpsw: It's not like the coding style is inconsistent in the codebase. It should be pretty straightforward to infer it.

I don't mind spending some time writing explicitly all the rules, but TBH just looking around the files should answer any question you have about the style instantly.

I have some experience writing configuration files for `uncrustify` (similar to `clang-format`) but it's actually pretty tricky to get them right. Do you want to block this PR until this is figured out?

jmoguillansky-gpsw: no, I'm not blocking anything, I'm just sharing suggestions.  Some of the rules are not so straightforward: like void space star instead of void star space, using two newlines after copyright header instead of one, using trailing commas, etc...
The benefit of tools is that it's not easy for the user to catch every syntax issue because their coding style becomes second nature, it's not something we consciously think about when writing code.  Tools can do a much better job with regards to enforcing coding style

cboesch-gpsw: > Some of the rules are not so straightforward:
> like void space star instead of void star space

But we do use the same pattern everywhere, don't we? (The reason is because `void *p, *q` is less error-prone (because otherwise you're tempted to write `void* p, q`

> using two newlines after copyright header instead of one

This is likely an inconsistency. We want to fix those and yeah a tool might help here.

> using trailing commas, etc...

The thing with this one is that typically it's context dependent and an auto formatting tool can't figure it out:

```c
enum {
    FOO_XXX,
    FOO_YYY,
    FOO_NB
};
```

In this case, we will never want someone to add an entry after `FOO_NB` because it's the number of entries and must always be the last one. So we definitely want to explicit that without a comma. But if we have:

```c
enum {
    FOO_XXX,
    FOO_YYY,
};
```

We want a trailing comma because we don't want to track modification on the last line when adding a new entry.

Of course it makes sense to document all that in style document, but a tool will hardly figure that out. In uncrustify what you have is a "soft" modification system, where basically it doesn't force a change unless it's violating the specified style (for example it will allow some manual vertical align with thresholds and that kind of stuff). There are a lot of settings to deal with these kind of issues, and it takes a very long time to figure them out properly. And that also means it will still not fix it for you if you add a comma after the `_NB`.

Basically my experience with automatic formatters is that trusting them blindly involve trade-offs, it's not all black and white.

jmoguillansky-gpsw (approved):

mbouron-gpsw (approved):

PR #243, opened 2021-03-16
tools/desktop: honor a few env variables to find the temp dir

cboesch-gpsw:
In the desktop Python hooks, we use tempfile.gettempdir() to find the temporary directory. This function will honor a few environment variables: TMPDIR, TEMP and TMP in priority. In ngl-desktop on the other hand, we hardcode /tmp (at least on Unix). Unfortunately, on macOS TMPDIR is set (to something different than /tmp), making ngl-control unable to find ngl-desktop instances.

Fixes regression since db87cad50639acecef120b7973a2aea11a37d434.

### DIFF
--- a/ngl-tools/ngl-desktop.c
+++ b/ngl-tools/ngl-desktop.c
@@ -572,15 +572,31 @@ static int makedirs(const char *path, int mode)
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


mbouron-gpsw (approved):

cboesch-gpsw:
Merged. Sorry I didn't wait for 24 hours but this fix was needed for today.

PR #245, opened 2021-03-18
CI: update CI tests, support vulkan backend

jmoguillansky-gpsw:


.github/workflows/ci_linux.yml
```diff
@@ -10,6 +10,8 @@ jobs:
   linux:
 
     runs-on: ubuntu-20.04
+    env:
+        VULKAN_SDK_VERSION: 1.2.170
```
mbouron-gpsw, line 14: Can't we just depend on mesa ? I guess ninja-build renderdoc is useful for master, can you do a separate PR with this two dependencies targetting master ?

jmoguillansky-gpsw: The goal is to confirm that vulkan backend builds ok in ci test.

mbouron-gpsw: I mean, can we rely on system packages ? like libvulkan-dev ?

mbouron-gpsw: Looks like shaderc is not packaged on Ubuntu so the commit is fine.

PR #248, opened 2021-03-19
build/windows: don't use vcpkg integrate install for pthreads support

jmoguillansky-gpsw:
vcpkg integrate install logic is broken.  The way it works is it adds
all libraries under installed/x64-windows/lib as linker args for all
MSVC projects.  This adds unnecessary additional libraries and slows
down the build.  See https://github.com/microsoft/vcpkg/issues/15452.


jmoguillansky-gpsw:
This solution patches the pthreads package and adds a pkg-config file. An alternate solution is to use "declare_dependency" and set the include paths, libs, and library paths directly for pthreads

jmoguillansky-gpsw:
corresponding sxplayer PR: 
https://github.com/Stupeflix/sxplayer/pull/24

mbouron-gpsw:
This has been addressed by removing the pthread dependency on Windows (https://github.com/gopro/gopro-lib-node.gl/pull/255)

PR #252, opened 2021-03-19
gpu_capture: pass gctx to gpu_capture_create()

mbouron-gpsw:


jmoguillansky-gpsw:
it might be better to keep the capture code more decoupled from the node.gl code, so it's more generic and reusable, and modular.
The capture code doesn't need the whole gctx.  It just needs a device handle

jmoguillansky-gpsw (dismissed):

jmoguillansky-gpsw (commented):
is there a way to tell if a gctx is a gl context, vulkan context, etc (e.g. if a gctx is a gctx_vk, etc)?

jmoguillansky-gpsw (dismissed):

libnodegl/backends/vk/gctx_vk.c
```diff
@@ -545,7 +545,7 @@ static int vk_init(struct gctx *s)
     const char *var = getenv("NGL_GPU_CAPTURE");
     s->gpu_capture = var && !strcmp(var, "yes");
     if (s->gpu_capture) {
-        s->gpu_capture_ctx = gpu_capture_ctx_create();
+        s->gpu_capture_ctx = gpu_capture_ctx_create(&s);
```
jmoguillansky-gpsw, line 548: this should be s, not &s.

libnodegl/backends/gl/gctx_gl.c
```diff
@@ -394,7 +394,7 @@ static int gl_init(struct gctx *s)
     const char *var = getenv("NGL_GPU_CAPTURE");
     s->gpu_capture = var && !strcmp(var, "yes");
     if (s->gpu_capture) {
-        s->gpu_capture_ctx = gpu_capture_ctx_create();
+        s->gpu_capture_ctx = gpu_capture_ctx_create(&s);
```
jmoguillansky-gpsw, line 397: this should be s, not &s

libnodegl/backends/vk/gctx_vk.c
mbouron-gpsw: fixed, thanks.

libnodegl/backends/gl/gctx_gl.c
mbouron-gpsw: fixed.

PR #255, opened 2021-03-24
internal: add pthread compatibility header

mbouron-gpsw:


.github/workflows/ci_msvc.yml
```diff
@@ -47,7 +47,6 @@ jobs:
           pip.exe install meson
           cd /mnt/c/vcpkg
           ./vcpkg.exe install --triplet ${{env.VCPKG_TRIPLET}} ${{env.VCPKG_INSTALL_PACKAGES}}
-          ./vcpkg.exe integrate install
```
ubitux, line 50: can we have an explanation on why this goes away?

mbouron-gpsw: Done.

cboesch-gpsw (approved):

jmoguillansky-gpsw (approved):

libnodegl/android_handlerthread.c
```diff
@@ -19,13 +19,13 @@
...
  */
 
-#include <pthread.h>
 #include <stdlib.h>
 
 #include "android_looper.h"
 #include "android_handler.h"
 #include "android_handlerthread.h"
 #include "memory.h"
+#include "pthread_compat.h"
```
jmoguillansky-gpsw, line 28: we can still use pthread.h for android-specific code?

mbouron-gpsw: IMHO, it eases maintenance as you only have to look for pthread_compat.h across the code base and you know that <pthread.h> is only included in pthread_compat.h.

PR #259, opened 2021-03-25
fix ngl desktop capture

jmoguillansky-gpsw:
RenderDoc GPU capture doesn't work when we create multiple graphics contexts, including dummy contexts.
When user selects a specific backend, no need to probe other backends.
Probe / Create the backend that the user specified

### DIFF
--- a/ngl-tools/opts.c
+++ b/ngl-tools/opts.c
@@ -93,14 +93,29 @@ static int get_backend(const char *id)
         return NGL_BACKEND_AUTO;
 
     int nb_backends;
-    struct ngl_backend *backends;
-    int ret = ngl_backends_probe(NULL, &nb_backends, &backends);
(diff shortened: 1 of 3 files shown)


ngl-tools/opts.c
```diff
@@ -93,14 +93,30 @@ static int get_backend(const char *id)
         return NGL_BACKEND_AUTO;
 
     int nb_backends;
-    struct ngl_backend *backends;
-    int ret = ngl_backends_probe(NULL, &nb_backends, &backends);
+    struct ngl_backend *backends = NULL;
```
jmoguillansky-gpsw, line 96: please merge ngl-tools: fix memory leak PR first, then rebase on master

jmoguillansky-gpsw, line 96: please merge ngl-tools: fix memory leak PR first, then rebase on master

PR #263, opened 2021-03-29
media: add auto_hwaccel param

mbouron-gpsw:
This parameter is useful to disable hardware decoding.


jmoguillansky-gpsw:
suggest to rename to "hwaccel" = [ true | false ]. 
If hwaccel is true and not supported, it will fall back to sw decoding

jmoguillansky-gpsw:
Also, why do you want to disable hardware decoding?

mbouron-gpsw:
> Also, why do you want to disable hardware decoding?

It is useful for testing purposes. Hardware decoders can fail to decode certain bitstream properly (the output can be garbled, ...) and you want to enforce the use of a software decoder to compare the output.
It can also be useful for running tests on different platforms using the same decoder implementation (FFmpeg here).

mbouron-gpsw:
> suggest to rename to "hwaccel" = [ true | false ].
> If hwaccel is true and not supported, it will fall back to sw decoding

I used this name to match the one in sxplayer. I am not against renaming it to hwaccel as you suggest. @cboesch-gpsw  what do you think ?

cboesch-gpsw:
> > suggest to rename to "hwaccel" = [ true | false ].
> > If hwaccel is true and not supported, it will fall back to sw decoding
> 
> I used this name to match the one in sxplayer. I am not against renaming it to hwaccel as you suggest. @cboesch-gpsw what do you think ?

I initially thought a bool param would have been appropriate but I'd make a `hwaccel` select param with `auto` and `disabled` as choices instead because "hwaccel=true" would otherwise lead to no acceleration sometimes, which is kinda counter intuitive.

mbouron-gpsw:
> > > suggest to rename to "hwaccel" = [ true | false ].
> > > If hwaccel is true and not supported, it will fall back to sw decoding
> > 
> > 
> > I used this name to match the one in sxplayer. I am not against renaming it to hwaccel as you suggest. @cboesch-gpsw what do you think ?
> 
> I initially thought a bool param would have been appropriate but I'd make a `hwaccel` select param with `auto` and `disabled` as choices instead because "hwaccel=true" would otherwise lead to no acceleration sometimes, which is kinda counter intuitive.

Updated.

cboesch-gpsw (approved):

PR #264, opened 2021-03-31
ngl-env

cboesch-gpsw:
~(Superset of https://github.com/gopro/gopro-lib-node.gl/pull/268)~


jmoguillansky-gpsw:
Thanks for the contribution.

A few comments:

First in terms of the design:
1) I think there's room to simplify / improve.  I suggest to move the core functionality to a module, e.g. ngl-build or whatever you want to call it.
Have a top level build.py python file which imports ngl-build and calls high-level APIs.
Make the APIs more generic and reusable, and don't reference global variables, etc.
Have a second build.py python file under external that deals with downloading / building the external deps.
2) One of the main things missing, compared to the gnu make approach is the ability to express targets and dependencies.
I suggest to explore the option of the Python module (e.g. ngl-build) emitting a ninja.build file.
This allows us to express targets and dependencies, and to retain very similar control flow / logic to the makefile based approach.  Also, the ninja.build syntax is very similar to gnu make.  

In terms of functionality:
1) can we add ability to list all targets (equivalent to ninja -t targets all)
2) can we add "dry-run": don't actually invoke the commands, just output the commands (equivalent to make -n)
3) please add a parameter to select the meson backend (e.g. ninja, xcode, etc)
4) please add a parameter to select the build directory.  This is useful if we're building with different configurations, backends, etc.

.github/workflows/ci_msvc.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     defaults:
       run:
-        shell: wsl-bash {0}
+        shell: powershell
 
     steps:
       - uses: Vampire/setup-wsl@v1
```
jmoguillansky-gpsw, line 24: please remove wsl

doc/expl/ngl-env.md
```diff
@@ -0,0 +1,38 @@
...
+
+The main command is `build` (see `./ngl-env build -h`). This command creates a
+Python Virtual environment in the `venv` directory (unless customized to
+something else with the `-p` option). In this virtual environment, `pip` is
+installed, along with `meson` (needed for building some of our components) and
+various other module requirements. A few external dependencies such as
+`sxplayer` are then pull, compiled and installed within this environment,
+followed by all the node.gl components.
+
+The `build` command is re-entrant, so a developer can modify the code and
```
jmoguillansky-gpsw, line 22: can you clarify this portion?
Are we supposed to modify the code to do incremental rebuilding?

scripts/msvc-env.ps1
```diff
@@ -9,4 +13,11 @@ $vsInstallationInfo = Get-VSSetupInstance -All -Prerelease |
     Select-Object -First 1
 $vcvars64 = "$($vsInstallationInfo.InstallationPath)\VC\Auxiliary\Build\vcvars64.bat"
 
-Write-Host $vcvars64
+cmd /c "call `"$vcvars64`" && set" |
```
jmoguillansky-gpsw, line 16: This is interesting.
Can you clarify what it does?
Does it call vcvars64 in a child process (e.g. cmd) and then extract the environment variables, and set them in the parent shell?
Is there room to simplify?

jmoguillansky-gpsw:
it may also be useful to dump the current directory and the environment variables

scripts/msvc-env.ps1
jmoguillansky-gpsw: I see that there's a "Developer PowerShell for VS 2019" option in the Start Menu.

Maybe it's possible to use that?

If you click on the shortcut, it does this:

C:\Windows\SysWOW64\WindowsPowerShell\v1.0\powershell.exe -noe -c "&{Import-Module """C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"""; Enter-VsDevShell 2f8cc3ee}"

jmoguillansky-gpsw: https://intellitect.com/enter-vsdevshell-powershell/

mbouron-gpsw: > https://intellitect.com/enter-vsdevshell-powershell/

I'll try to use this method instead, its simpler than the current solution, thanks.

.github/workflows/ci_msvc.yml
mbouron-gpsw: The get-vcpkg-cache-key step must be ported to powershell first. The plan was to address this topic as a separate commit / pr as it's not strictly required by this PR.

scripts/msvc-env.ps1
mbouron-gpsw: > This is interesting.
> Can you clarify what it does?
> Does it call vcvars64 in a child process (e.g. cmd) and then extract the environment variables, and set them in the parent shell?

It does exactly what you said.

> Is there room to simplify?

Another simpler solution is call to powershell from the command prompt (while it's simpler, l prefer not to use nested shells).

.github/workflows/ci_msvc.yml
jmoguillansky-gpsw: you can probably just use "bash" (from git bash) instead of wsl-bash as a quick solution.
Since this is only used on CI server, it should be ok?

cboesch-gpsw:
>     1. I think there's room to simplify / improve.  I suggest to move the core functionality to a module, e.g. ngl-build or whatever you want to call it.
>        Have a top level build.py python file which imports ngl-build and calls high-level APIs.

We can't have modules because we don't want to install the script. This is why I made a standalone script and did not split it. Having imports without a module is pretty wonky, you typically can't do relative imports. Splitting without a module would make it fragile because it would make difficult to track down where all the different files are located.

>        Make the APIs more generic and reusable, and don't reference global variables, etc.

More generic and reusable: do we really want that? This script is not meant to be a generic solution to be re-used elsewhere, it's a helper for developers and casual users, we want to keep it as simple as possible.

What global variables are you referring to? The global we use are non-muted constants/specifications/

>        Have a second build.py python file under external that deals with downloading / building the external deps.

Again, the monolithic approach was chosen on purpose. I made sure the external deps are the first thing you see in the file to ease editing it.

>     2. One of the main things missing, compared to the gnu make approach is the ability to express targets and dependencies.

Yes I'm well aware...

>        I suggest to explore the option of the Python module (e.g. ngl-build) emitting a ninja.build file.
>        This allows us to express targets and dependencies, and to retain very similar control flow / logic to the makefile based approach.  Also, the ninja.build syntax is very similar to gnu make.

Ninja is installed within the virtual env, which means we would probably need to trigger it from the `ngl-env` script itself. We also don't rely on commands anymore: typically, we are using the `venv` Python API to build the environment instead of calling `python -m venv ...`.

Do we want to roll back that approach?

> In terms of functionality:
> 
>     1. can we add ability to list all targets (equivalent to ninja -t targets all)

`,/ngl-env build -h`

>     2. can we add "dry-run": don't actually invoke the commands, just output the commands (equivalent to make -n)

We don't just run commands anymore, so no.

>     3. please add a parameter to select the meson backend (e.g. ninja, xcode, etc)

The meson backend for which target? How do you want that exposed? This sounds like a new feature which we don't support currently in the `Makefile`.

>     4. please add a parameter to select the build directory.  This is useful if we're building with different configurations, backends, etc.

This is not supported by the current `Makefile`, so I'd delay such addition to a later PR.

doc/expl/ngl-env.md
cboesch-gpsw: It means modifying the code and re-running the `build` command works as expected. Modifying the code refers to the libraries and tools, not `ngl-env`.

cboesch-gpsw:
> it may also be useful to add an option to dump the current directory and the environment variables

What do you mean by that?

jmoguillansky-gpsw:
> > ```
> > 1. I think there's room to simplify / improve.  I suggest to move the core functionality to a module, e.g. ngl-build or whatever you want to call it.
> >    Have a top level build.py python file which imports ngl-build and calls high-level APIs.
> > ```
> 
> We can't have modules because we don't want to install the script. This is why I made a standalone script and did not split it. Having imports without a module is pretty wonky, you typically can't do relative imports. Splitting without a module would make it fragile because it would make difficult to track down where all the different files are located.
> 
> > ```
> >    Make the APIs more generic and reusable, and don't reference global variables, etc.
> > ```
> 
> More generic and reusable: do we really want that? This script is not meant to be a generic solution to be re-used elsewhere, it's a helper for developers and casual users, we want to keep it as simple as possible.
> 
> What global variables are you referring to? The global we use are non-muted constants/specifications/
> 
> > ```
> >    Have a second build.py python file under external that deals with downloading / building the external deps.
> > ```
> 
> Again, the monolithic approach was chosen on purpose. I made sure the external deps are the first thing you see in the file to ease editing it.
> 
> > ```
> > 2. One of the main things missing, compared to the gnu make approach is the ability to express targets and dependencies.
> > ```
> 
> Yes I'm well aware...
> 
> > ```
> >    I suggest to explore the option of the Python module (e.g. ngl-build) emitting a ninja.build file.
> >    This allows us to express targets and dependencies, and to retain very similar control flow / logic to the makefile based approach.  Also, the ninja.build syntax is very similar to gnu make.
> > ```
> 
> Ninja is installed within the virtual env, which means we would probably need to trigger it from the `ngl-env` script itself. We also don't rely on commands anymore: typically, we are using the `venv` Python API to build the environment instead of calling `python -m venv ...`.
> 
> Do we want to roll back that approach?
> 
> > In terms of functionality:
> > ```
> > 1. can we add ability to list all targets (equivalent to ninja -t targets all)
> > ```
> 
> `,/ngl-env build -h`
> 
> > ```
> > 2. can we add "dry-run": don't actually invoke the commands, just output the commands (equivalent to make -n)
> > ```
> 
> We don't just run commands anymore, so no.
> 
> > ```
> > 3. please add a parameter to select the meson backend (e.g. ninja, xcode, etc)
> > ```
> 
> The meson backend for which target? How do you want that exposed? This sounds like a new feature which we don't support currently in the `Makefile`.
> 
> > ```
> > 4. please add a parameter to select the build directory.  This is useful if we're building with different configurations, backends, etc.
> > ```
> 
> This is not supported by the current `Makefile`, so I'd delay such addition to a later PR.

You don't need to install the script (build.py), just the module.  For example, meson is a python module.  Meson.build is a build script (or build file).  We install meson.  Similarly we can install ngl-build python module and build.py can import it?

jmoguillansky-gpsw:
>> More generic and reusable: do we really want that? This script is not meant to be a generic solution to be re-used elsewhere, it's a helper for developers and casual users, we want to keep it as simple as possible.

It doesn't have to be used everywhere, it's ok if it's only used here.  This could be a simpler solution.  The user doesn't need to concern themselves with the internal implementation of many of the functions, just the public apis and how they're used.  This would make the build.py file simpler. This would be a more scalable solution.

cboesch-gpsw:
> You don't need to install the script (build.py), just the module. For example, meson is a python module. Meson.build is a build script (or build file). We install meson. Similarly we can install ngl-build python module and build.py can import it?

We need a virtual env to install any module because we don't want to alter the system or user directory, and the script is responsible for creating that environment.

cboesch-gpsw:
> The user doesn't need to concern themselves with the internal implementation of many of the functions, just the public apis and how they're used.

What user are you exactly referring to? A casual nodegl user will never have to care about the `ngl-env` implementation in the current state; what am i misunderstanding?

jmoguillansky-gpsw:
>> Ninja is installed within the virtual env, which means we would probably need to trigger it from the ngl-env script itself. We also don't rely on commands anymore: typically, we are using the venv Python API to build the environment instead of calling python -m venv ....

Do we want to roll back that approach?

You can call python commands inside the ninja.build file. You can call ngl-build Python APIs.  You can either call python -m venv or some function in ngl-build to setup the venv, etc.

jmoguillansky-gpsw:
This is not supported by the current Makefile, so I'd delay such addition to a later PR.

>> It's suplorted in our latest development version of the Makefile (in multibackend branch).  We should continue from our latest work.  If you want to delay this to a future PR , ok, but in general I think we should continue from our latest effort

cboesch-gpsw:
> > > Ninja is installed within the virtual env, which means we would probably need to trigger it from the ngl-env script itself. We also don't rely on commands anymore: typically, we are using the venv Python API to build the environment instead of calling python -m venv ....
> 
> Do we want to roll back that approach?
> 
> You can call python commands inside the ninja.build file. You can call ngl-build Python APIs. You can either call python -m venv or some function in ngl-build to setup the venv, etc.

Who calls what exactly in your design? Ninja requires the venv to be available to the user. Can you elaborate on the complete flow you see (what scripts creates what, what commands the user runs, etc)?

jmoguillansky-gpsw:
> > > > Ninja is installed within the virtual env, which means we would probably need to trigger it from the ngl-env script itself. We also don't rely on commands anymore: typically, we are using the venv Python API to build the environment instead of calling python -m venv ....
> > 
> > 
> > Do we want to roll back that approach?
> > You can call python commands inside the ninja.build file. You can call ngl-build Python APIs. You can either call python -m venv or some function in ngl-build to setup the venv, etc.
> 
> Who calls what exactly in your design? Ninja requires the venv to be available to the user. Can you elaborate on the complete flow you see (what scripts creates what, what commands the user runs, 

Ninja can still setup the venv, the same way that the makefile does, it just needs to be the first target that gets invoked

jmoguillansky-gpsw:
> > > > > Ninja is installed within the virtual env, which means we would probably need to trigger it from the ngl-env script itself. We also don't rely on commands anymore: typically, we are using the venv Python API to build the environment instead of calling python -m venv ....
> > > 
> > > 
> > > Do we want to roll back that approach?
> > > You can call python commands inside the ninja.build file. You can call ngl-build Python APIs. You can either call python -m venv or some function in ngl-build to setup the venv, etc.
> > 
> > 
> > Who calls what exactly in your design? Ninja requires the venv to be available to the user. Can you elaborate on the complete flow you see (what scripts creates what, what commands the user runs,
> 
> Ninja can still setup the venv, the same way that the makefile does, it just needs to be the first target that gets invoked

We don't need to install ngl-build in the venv, similar to how we don't install gnu make.  It justs needs to be in the pythonpath

cboesch-gpsw:
> Ninja can still setup the venv, the same way that the makefile does, it just needs to be the first target that gets invoked

But then you need the user to manually install ninja on the system in addition to Python.

cboesch-gpsw:
> We don't need to install ngl-build in the venv, similar to how we don't install gnu make. It justs needs to be in the pythonpath

Yeah but like I said if you don't install `ngl-env` or its dependency as a module, you can't do relative imports and all the other files end up being scattered in random places hard to track.

jmoguillansky-gpsw:
> > Ninja can still setup the venv, the same way that the makefile does, it just needs to be the first target that gets invoked
> 
> But then you need the user to manually install ninja on the system in addition to Python

Yes, similar to how gnu make is installed in the system

jmoguillansky-gpsw:
> > > > Ninja is installed within the virtual env, which means we would probably need to trigger it from the ngl-env script itself. We also don't rely on commands anymore: typically, we are using the venv Python API to build the environment instead of calling python -m venv ....
> > 
> > 
> > Do we want to roll back that approach?
> > You can call python commands inside the ninja.build file. You can call ngl-build Python APIs. You can either call python -m venv or some function in ngl-build to setup the venv, etc.
> 
> Who calls what exactly in your design? Ninja requires the venv to be available to the user. Can you elaborate on the complete flow you see (what scripts creates what, what commands the user runs, etc)?

Ninja runs inside a parent shell (similar to how gnu make runs inside a shell).  On windows, it can run inside powershell.  On linux, macOS: bash.  Ninja can invoke shell commands to download / extract the archives.  It can be a git repo, tarball, zip, 7z, etc.  It also invokes meson and cmake.  Python generates the platform specific ninja.build

doc/expl/ngl-env.md
jmoguillansky-gpsw: Oh I see

jmoguillansky-gpsw:
> > The user doesn't need to concern themselves with the internal implementation of many of the functions, just the public apis and how they're used.
> 
> What user are you exactly referring to? A casual nodegl user will never have to care about the `ngl-env` implementation in the current state; what am i misunderstanding?

The individuals modifying / maintaining the build script.

I'm not sure I understand why there's a requirement that the module has to be installed in order to be used?  When we invoke the python script, can't we just modify PYTHONPATH env variable : append the path containing the module?

jmoguillansky-gpsw:
> > it may also be useful to add an option to dump the current directory and the environment variables
> 
> What do you mean by that? You mean adding some debug logging?

Yes.  Similar to how make can dump directory changes

jbnoel-gpsw:
@jmoguillansky-gpsw @cboesch-gpsw
Concerning the use of make:
- Removal was requested initially by @jmoguillansky-gpsw
- `make` is problematic on Windows

The goal was to simplify the build system.
We should merge this PR, if all the following conditions are satisfied:
1. the current implementation is a strict improvement compared to current state
2. there is no blocker/regression
3. the code is clear enough for the next iteration / suggestion / PR

Please, do not spend too much time in over-engineering a perfect design.

@jmoguillansky-gpsw make sure to make good use `quote` markdown ;-)

jmoguillansky-gpsw:
> @jmoguillansky-gpsw @cboesch-gpsw
> Concerning the use of make:
> 
> * Removal was requested initially by @jmoguillansky-gpsw
> * `make` is problematic on Windows
> 
> The goal was to simplify the build system.
> We should merge this PR, if all the following conditions are satisfied:
> 
> 1. the current implementation is a strict improvement compared to current state
> 2. there is no blocker/regression
> 3. the code is clear enough for the next iteration / suggestion / PR
> 
> Please, do not spend too much time in over-engineering a perfect design.
> 
> @jmoguillansky-gpsw make sure to make good use `quote` markdown ;-)

@jbnoel-gpsw if you guys want to merge it feel free but this is my perspective:

1.  I didn't request removal of make, but I'm okay with removing it.
2.  there is a regression in functionality.  The new approach doesn't support targets and dependencies.  It may not support incremental development and rebuilding as well.
If you're on Linux or MacOS, forgetting about Windows support: do you prefer the new solution or the gnu make - based approach?  My hope is that we can can achieve a solution that's on par with the gnu make based approach.  
We've invested a considerable amount of time building on top of the gnu make, and the build system is something we use every day.

jmoguillansky-gpsw (approved):

cboesch-gpsw:
Alternative implementation: https://github.com/gopro/gopro-lib-node.gl/pull/276

I'll close this PR since the alternative is likely an improvement over this one.

PR #269, opened 2021-04-06
build/Makefile: update ngfx-install rule

jmoguillansky-gpsw:


.github/workflows/ci_mac.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        brew install python3 curl pkg-config sdl2 ffmpeg glm spirv-cross ninja
+        brew install python3 curl pkg-config sdl2 ffmpeg glm spirv-cross ninja bash coreutils
```
wmacdonald-gpsw, line 19: Is it really necessary to install bash on mac? what is it used for?

wmacdonald-gpsw (commented):
General feedback: It is helpful for a PR to have some description to indicate the motivation for the changes.

jmoguillansky-gpsw:
> General feedback: It is helpful for a PR to have some description to indicate the motivation for the changes.

I updated the ngfx build system: added some build scripts to ngfx to install the dependencies, and added some CI tests to ngfx, currently they just verify the build

.github/workflows/ci_mac.yml
jmoguillansky-gpsw: I have some bash scripts over here:
https://github.com/gopro/ngfx/tree/main/build_scripts

I think the next step for ngfx build is to:
1) switch from gnu make to a ninja-based approach
2) we can switch from bash scripts to something like python, or we can use bash scripts on mac_os/linux and powershell scripts on windows?

The ngfx changes are an intermediate step forward

jmoguillansky-gpsw: also, the version of bash that's preinstalled on mac os is old (I think version 3), while brew installs version 4, that's why I install the latest bash via brew

wmacdonald-gpsw: Ok, thanks. I understand why we are currently using bash, and support looking at alternatives. Your second comment was the answer I was interested in.

wmacdonald-gpsw (approved):

PR #276, opened 2021-04-16
ngl env (v2: configure.py+[n]make)

cboesch-gpsw:


jmoguillansky-gpsw:
Thanks for the improvement.  Some comments below

jmoguillansky-gpsw:
"Last but not least, the
added branches in the Makefile are becoming a maintainance nightmare."

>> Please remove this portion from the commit message (we already removed this issue in the makefile-based approach)

"Large part of the Windows support done by Matthieu Bouron."

>> Please reword this commit:
"Large part of the new design for Windows support done by Matthieu Bouron."
or something similar

jmoguillansky-gpsw:
"Pulling a whole Linux distribution
just for one tool (Make) is completely overkill."

>>Please remove this portion from the commit message.  That wasn't the only reason to use WSL.  It also provides pkg-config tool, bash script support, and a similar environment as Linux, with tools such as untar, etc.

jmoguillansky-gpsw:
I will integrate / test in multi-backend branch and get back to you shortly

configure.py
```diff
@@ -0,0 +1,543 @@
...
+        '@echo "    Build completed."',
+        '@echo',
+        '@echo "    You can now enter the venv with:"',
+        f'@echo "        {activate}"',
+    ]
+
+
+@_block('tests-setup', [_ngl_tools_install, _pynodegl_utils_install])
+def _tests_setup(cfg):
+    return ['$(MESON_SETUP) ' + _cmd_join('tests', op.join('builddir', 'tests'))]
```
jmoguillansky-gpsw, line 280: this won't work when backend is visual studio.  meson test only support ninja backend

.github/workflows/ci_msvc.yml
```diff
@@ -42,17 +47,18 @@ jobs:
...
       - name: Build
         run: |
-          cd /mnt/c
-          # Workaround for meson issue #8329
-          # Currently meson requires the project to be on the same drive as dependencies
-          cp -Rf /mnt/d/a/gopro-lib-node.gl/gopro-lib-node.gl .
-          cd gopro-lib-node.gl
-          make -j$(($(nproc)+1)) TARGET_OS=Windows DEBUG_GPU_CAPTURE=yes \
-          VCPKG_DIR='C:\\vcpkg'
+          git clean -fdx  # trash the cached builds
```
jmoguillansky-gpsw, line 55: why is this necessary on CI server?

.github/workflows/ci_msvc.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
     defaults:
       run:
-        shell: wsl-bash {0}
+        shell: powershell
 
     steps:
       - uses: Vampire/setup-wsl@v1
```
jmoguillansky-gpsw, line 24: you can remove wsl

.github/workflows/ci_msvc.yml
```diff
@@ -42,17 +47,18 @@ jobs:
...
-          # Workaround for meson issue #8329
-          # Currently meson requires the project to be on the same drive as dependencies
-          cp -Rf /mnt/d/a/gopro-lib-node.gl/gopro-lib-node.gl .
-          cd gopro-lib-node.gl
-          make -j$(($(nproc)+1)) TARGET_OS=Windows DEBUG_GPU_CAPTURE=yes \
-          VCPKG_DIR='C:\\vcpkg'
+          git clean -fdx  # trash the cached builds
+          .\scripts\msvc-env.ps1
+          python.exe .\configure.py --debug gpu_capture
+          nmake
```
jmoguillansky-gpsw, line 58: I think one thing missing from nmake is "dry run"?  Is there a way to enable?

configure.py
```diff
@@ -0,0 +1,543 @@
...
+                logging.info('extracting %s', dst_file)
+                zip_.extractall(dst_base)
+    else:
+        assert False
+
+    # Link
+    target = op.join(dst_base, name)
+    if _SYSTEM in ('Windows', 'MinGW'):
+        logging.info(f'copy {target} -> {extract_dir}')
+        shutil.copytree(extract_dir, target, dirs_exist_ok=True)
```
jmoguillansky-gpsw, line 108: now we're not using wsl anymore.  Does os.symlink still not work on Windows?

configure.py
```diff
@@ -0,0 +1,543 @@
...
+
+
+@_block('nodegl-install', [_nodegl_setup])
+def _nodegl_install(cfg):
+    return _meson_compile_install_cmd('libnodegl')
+
+
+@_block('pynodegl-deps-install', [_nodegl_install])
+def _pynodegl_deps_install(cfg):
+    return ['$(PIP) ' + _cmd_join('install', '-r', op.join('.', 'pynodegl', 'requirements.txt'))]
```
jmoguillansky-gpsw, line 193: does this use system PIP or PIP in virtual environment?

configure.py
```diff
@@ -0,0 +1,543 @@
...
+        try:
+            distrib_id = check_output(['lsb_release', '-si']).strip().decode()
+        except Exception:
+            pass
+        else:
+            if distrib_id in ('Ubuntu', 'Debian'):
+                meson_setup += ['--libdir=lib']
+
+    return dict(
+        PIP=_cmd_join(python, '-m', 'pip'),
```
jmoguillansky-gpsw, line 383: does this invoke system python or venv python?

configure.py
```diff
@@ -0,0 +1,543 @@
...
+        MESON=meson,
```
jmoguillansky-gpsw, line 384: does this invoke system meson or venv meson?

configure.py
```diff
@@ -0,0 +1,543 @@
...
+
+        if _SYSTEM == 'Windows':
+            _sxplayer_setup.prerequisites.append(_pkgconf_install)
+            if 'gpu_capture' in args.debug:
+                _nodegl_setup.prerequisites.append(_renderdoc_install)
+
+            vcpkg_bin = op.join(args.vcpkg_dir, 'installed', 'x64-windows', 'bin')
+            for f in glob.glob(op.join(vcpkg_bin, '*.dll')):
+                logging.info('copy %s to venv/Scripts', f)
+                shutil.copy2(f, op.join('venv', 'Scripts'))
```
jmoguillansky-gpsw, line 491: I don't think this logic is in existing Makefile?
https://github.com/gopro/gopro-lib-node.gl/blob/master/Makefile

This is a bit dangerous: it copies all dlls from vcpkg folder?

configure.py
cboesch-gpsw: Yes it is necessary if you don't have the developer mode activated apparently.

.github/workflows/ci_msvc.yml
cboesch-gpsw: Switched to `bash` as you suggested in the other PR.

.github/workflows/ci_msvc.yml
cboesch-gpsw: I thought it was because it fixed something during one of the iteration. I looked at the log and nothing is getting cleaned up so I probably got things mixed up here. Chunk removed.

configure.py
cboesch-gpsw: Fixed. Do you have a meson issue reference I can use here? There is no explanation on `master`.

configure.py
cboesch-gpsw: venv Python: `python = op.join(cfg.bin_path, 'python')`. I could rely on the `PATH` being set, but I wanted to explicit the path for this one in case `python` or `python3` is not in the venv and would fallback on the one in the system or that sort of stuff.

configure.py
cboesch-gpsw: venv meson unless not available (typically on MinGW), using the `PATH` being set.

configure.py
cboesch-gpsw: > I don't think this logic is in existing Makefile?
> https://github.com/gopro/gopro-lib-node.gl/blob/master/Makefile

It's not indeed, but we need to do that to be able to import properly the module until we have something better using add/set dll directory mechanism.

> This is a bit dangerous: it copies all dlls from vcpkg folder?

How dangerous exactly?

cboesch-gpsw:
> > "Pulling a whole Linux distribution
> > just for one tool (Make) is completely overkill."
>
> Can we reword this portion?  That wasn't the only reason to use WSL.  It also provides pkg-config tool, bash script support, and a similar environment as Linux, with tools such as untar, etc.  Maybe something like "Pulling a whole Linux distribution to support the build is completely overkill"

Sure, changed. Note that we're not relying on pkg-config nor bash script support currently afaict, but it's true for the external download thing.

> > "Last but not least, the
> > added branches in the Makefile are becoming a maintainance nightmare."
> 
> Can we remove this portion from the commit message (we already removed this issue in the makefile-based approach)?

Sorry I didn't catch that: I'm referring to the `ifeq`/`else`/`endif` in the `Makefile` on master, which go away with this PR.

configure.py
jmoguillansky-gpsw: it's not a meson bug, it's a meson design feature.  It will report an error if user chooses vs backend.

configure.py
jmoguillansky-gpsw: can we check if it's supported and if so use symlink?
Python should return an "OSError" if not supported?  
We can store a flag somewhere if system supports symlinks?

https://docs.python.org/3/library/os.html
Note On newer versions of Windows 10, unprivileged accounts can create symlinks if Developer Mode is enabled. When Developer Mode is not available/enabled, the SeCreateSymbolicLinkPrivilege privilege is required, or the process must be run as an administrator.
OSError is raised when the function is called by an unprivileged user.

configure.py
jmoguillansky-gpsw: we just don't know what packages the user has, and what dlls are being copied, etc.  
It would be nice if we merge this commit also to avoid the DLL duplication, but we can do that later?
https://github.com/gopro/gopro-lib-node.gl/commit/e305182148705c5a8986ec60f673a53af5ec3de6

configure.py
jmoguillansky-gpsw: > > > "Pulling a whole Linux distribution
> > > just for one tool (Make) is completely overkill."
> > 
> > 
> > Can we reword this portion?  That wasn't the only reason to use WSL.  It also provides pkg-config tool, bash script support, and a similar environment as Linux, with tools such as untar, etc.  Maybe something like "Pulling a whole Linux distribution to support the build is completely overkill"
> 
> Sure, changed. Note that we're not relying on pkg-config nor bash script support currently afaict, but it's true for the external download thing.
> 
> > > "Last but not least, the
> > > added branches in the Makefile are becoming a maintainance nightmare."
> > 
> > 
> > Can we remove this portion from the commit message (we already removed this issue in the makefile-based approach)?
> 
> Sorry I didn't catch that: I'm referring to the `ifeq`/`else`/`endif` in the `Makefile` on master, which go away with this PR.

We have a fix already but it's not merged into master.  Anyhow it's up to you.  Maybe something like this would be better?
Last but not least, this removes the extra branches in the Makefile which are a maintanance nightmare

jmoguillansky-gpsw:
when I run configure from powershell, it spawns a new console window, and then the window disappears?
Is there a way to pipe the output to the powershell terminal directly?

configure.py
```diff
@@ -0,0 +1,553 @@
...
+
+
+@_block('nodegl-setup', [_sxplayer_install])
+def _nodegl_setup(cfg):
+    nodegl_debug_opts = []
+    if cfg.args.debug:
+        debug_opts = ','.join(cfg.args.debug)
+        nodegl_debug_opts += [f'-Ddebug_opts={debug_opts}']
+
+    if 'gpu_capture' in cfg.args.debug:
```
jmoguillansky-gpsw, line 179: should this be inside if cfg.args.debug?  What is cfg.args.debug is None?

configure.py
```diff
@@ -0,0 +1,553 @@
...
+
+        # On MinGW we need the path translated from C:\ to /c/, because when
+        # part of the PATH, the ':' separator will break
+        if _SYSTEM == 'MinGW':
+            self.prefix = check_output(['cygpath', '-u', self.prefix]).strip().decode()
+
+        self.bin_name = 'Scripts' if _SYSTEM == 'Windows' else 'bin'
+        self.bin_path = op.join(self.prefix, self.bin_name)
+        if _SYSTEM == 'Windows':
+            vcpkg_pc_path = op.join(args.vcpkg_dir, 'installed', 'x64-windows', 'lib', 'pkgconfig')
```
jmoguillansky-gpsw, line 487: when using the argument parser, does it know to store this param to vcpkg_dir automatically?

 parser.add_argument('--vcpkg-dir'
 
 or do we need to add dest='vcpkg_dir'?

configure.py
```diff
@@ -0,0 +1,553 @@
...
+    return _meson_compile_install_cmd('sxplayer')
+
+
+@_block('renderdoc-install')
+def _renderdoc_install(cfg):
+    renderdoc_dll = op.join(cfg.externals['renderdoc'], 'renderdoc.dll')
+    return [f'copy {renderdoc_dll} {cfg.bin_path}']
+
+
+@_block('nodegl-setup', [_sxplayer_install])
```
jmoguillansky-gpsw, line 172: how can I set the prerequisites for a target conditionally based on the command line args?

configure.py
cboesch-gpsw: I don't think that's possible to have `cfg.args.debug` set to `None` since the default  is explicitly set to `[]`

configure.py
cboesch-gpsw: Yeah but this PR is not ready; Matthieu has a prototype with a wrapper around `pynodegl` that does that, but it's not that trivial so it probably belongs in another PR.

configure.py
cboesch-gpsw: Adjusted, thanks

configure.py
jmoguillansky-gpsw: I haven't updated the DLL PR, the link is to a specific commit.  I added an env variable NGL_DLL_DIRS and call add_dll_directory in the python module.  We can address it later.

configure.py
```diff
@@ -0,0 +1,559 @@
...
+def _nodegl_setup(cfg):
+    nodegl_debug_opts = []
+    if cfg.args.debug:
+        debug_opts = ','.join(cfg.args.debug)
+        nodegl_debug_opts += [f'-Ddebug_opts={debug_opts}']
+
+    if 'gpu_capture' in cfg.args.debug:
+        renderdoc_dir = cfg.externals['renderdoc']
+        nodegl_debug_opts += [f'-Drenderdoc_dir={renderdoc_dir}']
```
jmoguillansky-gpsw, line 181: general suggestion: it might be better to make renderdoc_dir a makefile variable so that the generated makefile is a little cleaner, but it's not a requirement.  Similarly, for any other global variables in the makefile

configure.py
```diff
@@ -0,0 +1,559 @@
...
+        if not op.exists(target):
+            logging.info(f'symlink {target} -> {extract_dir}')
+            os.symlink(extract_dir, target)
+    return name, target
+
+
+def _fetch_externals(args):
+    dependencies = _EXTERNAL_DEPS.copy()
+    if _SYSTEM not in {'Windows', 'MinGW'}:
+        del dependencies['pkgconf']
```
jmoguillansky-gpsw, line 122: maybe the code is a little simpler if we add the dependencies conditionally instead of removing them?

configure.py
```diff
@@ -0,0 +1,559 @@
...
+    # mechanism.
+    #
+    meson = 'MAKEFLAGS= meson' if _SYSTEM != 'Windows' else 'meson'
+
+    buildtype = 'debugoptimized' if cfg.args.coverage or cfg.args.buildtype == 'debug' else 'release'
+    meson_setup = [
+        'setup',
+        '--prefix', cfg.prefix,
+        '-Drpath=true',
+        '--pkg-config-path', cfg.pkg_config_path,
```
jmoguillansky-gpsw, line 368: we can simplify generated makefile:
'--pkg-config-path', '$(PKG_CONFIG_PATH')

jmoguillansky-gpsw:
It seems like nmake invokes shell commands using command prompt, not using powershell?

jmoguillansky-gpsw:
"WSL (or MSYS2 when not using the MSVC toolchain) to make it work. "
>> You can use msys2 with msvc toolchain.
Please reword:
On Windows this means we need an overlay such as WSL or MSYS2

jmoguillansky-gpsw:
"Last but not least, this
removes the extra branches in the Makefile which are becoming a
maintainance nightmare."

>> Please remove this point.  We already submitted a PR 2 months ago which resolved this issue, so it's no longer relevant.

configure.py
cboesch-gpsw: This is a bit annoying to do because we need a special escaping logic; `shlex.join` typically quotes `$(...)` while it shouldn't in the context of the `Makefile`. I could add heuristics in the quote mechanism to prevent that, but I'd rather have an ugly generated `Makefile` than a complex `configure.py`.

configure.py
cboesch-gpsw: > maybe the code is a little simpler if we add the dependencies conditionally instead of removing them

Yeah that wasn't great. I changed that to make it insert-based (and fixed a few things):
```python
def _get_external_deps(args):
    deps = ['sxplayer']
    if _SYSTEM == 'Windows':
        deps.append('pkgconf')
    if 'gpu_capture' in args.debug:
        if _SYSTEM not in {'Windows', 'Linux'}:
            raise Exception(f'Renderdoc is not supported on {_SYSTEM}')
        deps.append('renderdoc')
    return {dep: _EXTERNAL_DEPS[dep] for dep in deps}
```

> or when we define _EXTERNAL_DEPS, maybe we can have a field to specify which OS's?

A field is not enough because it's platform *and* options dependent; we'd need some lambda or dedicated functions to get availability.

configure.py
cboesch-gpsw: Same issue as with renderdoc (quoting)

cboesch-gpsw:
> It seems like nmake invokes shell commands using command prompt, not using powershell?
> FYI: https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/nmake-fatal-error-u1056?view=msvc-160

I guess; is it an issue?

cboesch-gpsw:
> > "WSL (or MSYS2 when not using the MSVC toolchain) to make it work. "
>
> You can use msys2 with msvc toolchain.

Can we? I don't think this was ever supported in our build.

Please try to quote properly in markdown: `>>` is for double level quote, not for your answer.

configure.py
jmoguillansky-gpsw: I see

jmoguillansky-gpsw:
> > It seems like nmake invokes shell commands using command prompt, not using powershell?
> > FYI: https://docs.microsoft.com/en-us/cpp/error-messages/tool-errors/nmake-fatal-error-u1056?view=msvc-160
> 
> I guess; is it an issue? How do you trigger that?

It's not an issue.  I just noticed when updating the configure script.

configure.py
cboesch-gpsw: > can we check if it's supported and if so use symlink?
> Python should return an "OSError" if not supported?
> We can store a flag somewhere if system supports symlinks?

Done:
```python
    # Remove previous link if needed
    target = op.join(dst_base, name)
    rel_extract_dir = op.basename(extract_dir)
    if op.islink(target) and os.readlink(target) != rel_extract_dir:
        logging.info('unlink %s target', target)
        os.unlink(target)
    elif op.exists(target) and not op.islink(target):
        logging.info('remove previous %s copy', target)
        shutil.rmtree(target)

    # Link (or copy)
    if not op.exists(target):
        logging.info('symlink %s -> %s', target, rel_extract_dir)
        try:
            os.symlink(rel_extract_dir, target)
        except OSError:
            # This typically happens on Windows when Developer Mode is not
            # available/enabled
            logging.info('unable to symlink, fallback on copy (%s -> %s)', target, extract_dir)
            shutil.copytree(extract_dir, target)
```

.github/workflows/ci_msvc.yml
```diff
@@ -18,19 +18,20 @@ jobs:
 
     defaults:
       run:
-        shell: wsl-bash {0}
+        shell: powershell
```
jmoguillansky-gpsw, line 21: fyi: if you do shell: pwsh it will use newer version

cboesch-gpsw: It appears to work, adjusted, thanks.

(For reference: https://stackoverflow.com/questions/60124810/what-is-the-difference-between-pwsh-and-powershell-integrated-console-on-vs)

cboesch-gpsw:
@jmoguillansky-gpsw did I address all your comments or is there something I missed?

jmoguillansky-gpsw:
> @jmoguillansky-gpsw did I address all your comments or is there something I missed?

Yes, I'm just finishing up integrating this into multi-backend branch, as soon as it's working i'll approve (should be today).
It would be nice if you can make some changes to the commit message but it's up to you:

On Windows this means we need an overlay such as WSL -> WSL or MSYS2
Last but not least, this removes the extra branches ... -> remove

.github/workflows/ci_mac.yml
```diff
@@ -20,5 +20,11 @@ jobs:
 
     - name: Build
       run: |
-        make -j$(($(sysctl -n hw.logicalcpu)+1))
+        ./configure.py
```
jmoguillansky-gpsw, line 23: --buildtype debug --debug gl gpu_capture

.github/workflows/ci_linux.yml
```diff
@@ -37,8 +37,15 @@ jobs:
 
     - name: Build
       run: |
-        make -j$(($(nproc)+1)) DEBUG_GPU_CAPTURE=yes
-    - name: Run tests ${{matrix.backend}}
+        ./configure.py --buildtype debug --debug gl gpu_capture
```
jmoguillansky-gpsw, line 40: nit:
it might be better to pass debug args as comma-separated values, but up to you
parser.add_argument('--debug', type=str)
debug_list = parser.parse_args().debug.split(",")

configure.py
```diff
@@ -0,0 +1,564 @@
...
+    url = dep['url'].replace('@VERSION@', version)
+    dst_file = dep.get('dst_file', op.basename(url)).replace('@VERSION@', version)
+    dst_base = op.join(_ROOTDIR, 'external')
+    dst_path = op.join(dst_base, dst_file)
+    os.makedirs(dst_base, exist_ok=True)
+
+    # Download
+    if not op.exists(dst_path):
+        logging.info('downloading %s to %s', url, dst_file)
+        urllib.request.urlretrieve(url, dst_path)
```
jmoguillansky-gpsw, line 93: I started getting an SSL certificate error on the CI server.  Investigating...

I saw this API is deprecated, you may want to update it.  
Also may want to add some additional error handling

Traceback (most recent call last):
  File ".\configure.py", line 682, in <module>
    _run()
  File ".\configure.py", line 669, in _run
    cfg = _Config(args)
  File ".\configure.py", line 599, in __init__
    self.externals = _fetch_externals(args)
  File ".\configure.py", line 151, in _fetch_externals
    return dict(p.map(_download_extract, dependencies.items()))
  File "C:\hostedtoolcache\windows\Python\3.8.9\x64\lib\multiprocessing\pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "C:\hostedtoolcache\windows\Python\3.8.9\x64\lib\multiprocessing\pool.py", line 771, in get
    raise self._value
urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1125)>

jmoguillansky-gpsw: seems like there's an issue with the pkgconf URL?
https://distfiles.dereferenced.org/pkgconf/pkgconf-@VERSION@.tar.xz

If you try to access in your browser: NET::ERR_CERT_DATE_INVALID

jmoguillansky-gpsw (approved):

configure.py
```diff
@@ -0,0 +1,564 @@
...
+
+_RENDERDOC_FILE = 'RenderDoc_@VERSION@_64.zip' if _SYSTEM == 'Windows' else 'renderdoc_@VERSION@.tar.gz'
+_EXTERNAL_DEPS = dict(
+    sxplayer=dict(
+        version='9.8.1',
+        url='https://github.com/Stupeflix/sxplayer/archive/v@VERSION@.tar.gz',
+        dst_file='sxplayer-@VERSION@.tar.gz',
+    ),
+    pkgconf=dict(
+        version='1.7.3',
```
jmoguillansky-gpsw, line 52: please use 1.7.4

.github/workflows/ci_linux.yml
cboesch-gpsw: Renamed to `--debug-opts`. Not sure about the split, the arg parser has that feature builtin.

configure.py
cboesch-gpsw: It appears it's been fixed on their side.

configure.py
cboesch-gpsw: Bumped in an earlier commit.

Also bumped RenderDoc in a separate commit.

cboesch-gpsw:
> Last but not least, this removes the extra branches ... -> remove

Huh?

Anyway, I adjusted a few other bits here and there, notably:
- simplified ubuntu/debian checking (actually align it with the logic used in meson)
- added an external dependency checksum in case of mis-download (or if the configure is aborted during a download)

mbouron-gpsw:
Typo in the commit description: jungling -> juggling.

configure.py
```diff
@@ -0,0 +1,581 @@
...
+    return ['$(MESON_SETUP) ' + _cmd_join('ngl-tools', op.join('builddir', 'ngl-tools'))]
+
+
+@_block('ngl-tools-install', [_ngl_tools_setup])
+def _ngl_tools_install(cfg):
+    ret = _meson_compile_install_cmd('ngl-tools')
+    if _SYSTEM == 'Windows':
+        bindir = cfg.bin_path
+        ngl_tools_dlls = op.join('builddir', 'ngl-tools', '*.dll')
+        ret += [f'xcopy /Y {ngl_tools_dlls} {bindir}']
```
mbouron-gpsw, line 276: This seems unneeded on my machine, there is no dll anymore in this folder.

mbouron-gpsw:
I tested on my machine, linux, mingw, msvc and it works fine. Other than my remarks (and the other remaining ones), LGTM.

mbouron-gpsw (approved):

scripts/msvc-env.ps1
```diff
@@ -9,4 +13,11 @@ $vsInstallationInfo = Get-VSSetupInstance -All -Prerelease |
...
 
-Write-Host $vcvars64
+cmd /c "call `"$vcvars64`" && set" |
+foreach {
+    if ($_ -match "=") {
+        $v = $_.split("="); Set-Item -force -Path "env:\$($v[0])" -Value "$($v[1])"
+    }
+}
+
+Set-Item -force -Path env:VCVARS64 -Value "$vcvars64"
```
mbouron-gpsw, line 23: This line can be simplified to $env:VCVARS64 = "$vcvars64" but keeping the use of the Set-Item form makes the code consistent with what we do above. You can choose whatever you prefer.

scripts/msvc-env.ps1
```diff
@@ -9,4 +13,11 @@ $vsInstallationInfo = Get-VSSetupInstance -All -Prerelease |
     Select-Object -First 1
 $vcvars64 = "$($vsInstallationInfo.InstallationPath)\VC\Auxiliary\Build\vcvars64.bat"
 
-Write-Host $vcvars64
+cmd /c "call `"$vcvars64`" && set" |
+foreach {
+    if ($_ -match "=") {
+        $v = $_.split("="); Set-Item -force -Path "env:\$($v[0])" -Value "$($v[1])"
```
mbouron-gpsw, line 19: You can split this into two lines and remove `;`.

mbouron-gpsw: `Set-Item -force -Path "env:\$($v[0])" -Value "$($v[1])"` can be simplified to `Set-Item -Force "env:$($v[0])" "$($v[1])"`
Note that the '\' before $() does not seem to do anything.

scripts/msvc-env.ps1
mbouron-gpsw: We can also use `Set-Item env:VCVARS64 "$vcvars64"` (to keep it consistent with the code above).

configure.py
cboesch-gpsw: > I started getting an SSL certificate error on the CI server. Investigating...
> 
> I saw this API is deprecated, you may want to update it:
> 
> ```
> import requests
> ...
> response = requests.get(url)
> with open(dst_file, 'wb') as file:
> file.write(response.content)
> ```

Forgot to reply to that part:
- the API is not deprecated, it is just legacy ("They might become deprecated at some point in the future.")
- `requests` is an external dependency, we won't want to use that
- i looked at `urlretrieve` implementation and they're doing quite a bit of work which I'm not exactly willing to duplicate

I'll keep `urlretrieve` for now.

configure.py
cboesch-gpsw: removed

scripts/msvc-env.ps1
cboesch-gpsw: applied your diff

scripts/msvc-env.ps1
cboesch-gpsw: applied your diff

PR #278, opened 2021-04-27
tests: add tolerance in rtt_clear_attachment_with_timeranges

mrobertseidowsky-gpsw:
### Environment
Ubuntu 20.04.2
Kernel driver in use: nvidia
NVIDIA Corporation GP107M [**GeForce GTX 1050** Ti Mobile] 
nvidia kernel drivers versions: NVIDIA UNIX x86_64 Kernel Module  **450.119.03**  

### Description

2 tests are failing:

- `BACKEND=opengl MALLOC_PERTURB_=102  rtt_clear_attachment_with_timeranges  `
- `BACKEND=opengl MALLOC_PERTURB_=98  rtt_clear_attachment_with_timeranges  `

### Solution

A tolerance is added in `rtt_clear_attachment_with_timeranges` test


jmoguillansky-gpsw:
Can you explain why adding a tolerance is necessary?

mrobertseidowsky-gpsw:
> Can you explain why adding a tolerance is necessary? Lets dump the output and manually compare. Also please try to capture using renderdoc, and we can further investigate. Could be a precision issue or something

Sorry for the lack of context, I already did the investigation with @cboesch-gpsw, the root cause is suspected to be during the image downscaling algorithm made by the library PIL (pillow): the hashs between the downscaled image computed on my computer and the downscalled reference image are different.
Therefore, it induces a bit flip into the hash diff.
These are explanations from @cboesch-gpsw  which will surely explain the problem better ;-)

cboesch-gpsw:
Can you add the following description to the commit?

```
Some frames have a few off-by-one pixels in the green component which
actually cascade down to the timber sized (9x9) picture of the finger print.
The reason for that is probably because these off-by-one are happening on 0x7F
vs 0x80 values (the middle of the 8-bit range). These offsets on the border of
the shape cause a slightly discontinuity in the location of the sign
differences in the final fingerprint.

The visual difference is unoticeable and perfectly acceptable, so a minimal 3%
tolerance is chosen to address the problem.

Reproducible on GeForce GTX 1050 Ti Mobile (with official drivers 450.119.03).
```

cboesch-gpsw:
> Can you explain why adding a tolerance is necessary? Lets dump the output and manually compare. Also please try to capture using renderdoc, and we can further investigate. Could be a precision issue or something

It's perfectly normal, we investigated, see the suggested commit description in my previous answer. tldr it's a few rounding off-by-ones that cascade down to the fingerprint following an unfortunate combination of circumstances.

jmoguillansky-gpsw:
I still would have liked further investigation before you merged it.  You can capture with renderdoc, and see which portion is different.  In the past I noticed some issues related to precision: but you can see it in the gpu capture.

Get Outlook for Android<https://aka.ms/AAb9ysg>

jmoguillansky-gpsw:
> > Can you explain why adding a tolerance is necessary? Lets dump the output and manually compare. Also please try to capture using renderdoc, and we can further investigate. Could be a precision issue or something
> 
> It's perfectly normal, we investigated, see the suggested commit description in my previous answer. tldr it's a few rounding off-by-ones that cascade down to the fingerprint following an unfortunate combination of circumstances.

You're just commenting on the output but you didn't identify the root cause

cboesch-gpsw:
> > > Can you explain why adding a tolerance is necessary? Lets dump the output and manually compare. Also please try to capture using renderdoc, and we can further investigate. Could be a precision issue or something
> > 
> > 
> > It's perfectly normal, we investigated, see the suggested commit description in my previous answer. tldr it's a few rounding off-by-ones that cascade down to the fingerprint following an unfortunate combination of circumstances.
> 
> You're just commenting on the output but you didn't identify the root cause

The exact root cause of the differences doesn't really matter since both outputs are correct and almost identical; we already accept that drivers and graphic cards generate similar but not exactly identical output, it is pretty common and the root causes are multiple and irrelevant. What was special here was the cascading of such off-by-one down to the fingerprint, which has been analyzed.

PR #281, opened 2021-05-05
pynodegl: add and use NGL_DLL_DIRS to set additional paths to the DLL search path

mbouron-gpsw:
This will allow the pynodegl module to find more easily the DLLs it
depends on.


pynodegl/pynodegl.py
```diff
@@ -0,0 +1,10 @@
+import os
+import platform
+import sysconfig
+
+if platform.system() == 'Windows':
+    ngl_home = os.getenv('NGL_HOME')
+    if ngl_home:
+        os.add_dll_directory(os.path.join(ngl_home, 'venv', 'Scripts')
```
mbouron-gpsw, line 8: This needs to be adjusted, venv/Scripts might not useful outside the ngl env itself and will not work on a proper installation.

mbouron-gpsw: Fixed.

pynodegl/pynodegl.py
```diff
@@ -0,0 +1,10 @@
+import os
+import platform
+import sysconfig
```
cboesch-gpsw, line 3: looks unused

cboesch-gpsw:
Sounds like a good idea; I'm looking forward the integration with the build ecosystem and activate.

pynodegl/pynodegl.py
jmoguillansky-gpsw: what's wrong with my previous solution?

if platform.system() == 'Windows':
    def add_dll_directories():
        dll_dirs = os.getenv("NGL_DLL_DIRS")
        if not dll_dirs:
            return
        dll_dirs = dll_dirs.split(';')
        for p in dll_dirs:
            if os.path.exists(p):
                os.add_dll_directory(p)

jmoguillansky-gpsw: https://github.com/gopro/gopro-lib-node.gl/pull/272/commits/99cf4d0beee5322daf5929a9150273c088e61c70

cboesch-gpsw: The issue wasn't in the code (which was reviewed), it was that it was in the wrong place: we needed that for pre-pynodegl, and for all pynodegl users. I believe we explained that it was the issue a few months ago.

jmoguillansky-gpsw: ok

pynodegl/pynodegl.py
mbouron-gpsw: Fixed.

pynodegl/pynodegl/__init__.py
```diff
@@ -19,6 +19,18 @@
...
 #
 
+import os
+import platform
+
+
+if platform.system() == 'Windows':
+    ngl_dll_dirs = os.getenv('NGL_DLL_DIRS')
+    if ngl_dll_dirs:
+        dll_dirs = ngl_dll_dirs.split(';')
```
jbnoel-gpsw, line 29: use `os.pathsep` (https://docs.python.org/3/library/os.html#os.pathsep) instead `';'`

cboesch-gpsw: Note that this is only for Windows. We also got surprises with Python constants in the past, typically under MSYS2.

mbouron-gpsw: `os.pathsep` should be fine, it returns `;` on windows native and mingw.

cboesch-gpsw (approved):

PR #285, opened 2021-05-11
rendertarget: add readable param

mbouron-gpsw:


mbouron-gpsw:
> Why do we need "readable" param?

We do for modern backends. This flags is used by the vulkan backend to allocate a staging buffer at initialization. This staging buffer is used later on to download the content of the render target.

jmoguillansky-gpsw:
> > Why do we need "readable" param?
> 
> We do for modern backends. This flags is used by the vulkan backend to allocate a staging buffer at initialization. This staging buffer is used later on to download the content of the render target.

For long term I suggest some architectural changes:
Consider moving rendertarget functionality to framebuffer, remove readpixels, and in texture class have upload / download / map / unmap functions or something similar.  For short term lgtm

cboesch-gpsw (approved):

PR #287, opened 2021-05-14
Path (part 1: animation)

cboesch-gpsw:
Path drawing is a work in progress that will land later. This part merely addresses the animation. It's already useful from a user point of view, and a corner stone for the remaining work.


libnodegl/path.c
```diff
@@ -0,0 +1,408 @@
...
+    return s;
+}
+
+/*
+ * Convert from original bézier quadratic form:
+ *   B₂(t) = (1-t)² p0 + 2(1-t)t p1 + t² p2
+ * To polynomial form:
+ *   B₂(t) = at² + bt + c
+ */
+static void poly_from_bezier2(float *dst, float p0, float p1, float p2)
```
mbouron-gpsw, line 181: Is it possible to move this function closer to its use in ngli_path_bezier2_to() ? (Same applies to poly_from_bezier3).

libnodegl/path.c
```diff
@@ -0,0 +1,408 @@
...
+static void poly_from_bezier3(float *dst, float p0, float p1, float p2, float p3)
+{
+    const float a =     -p0 + 3.f*p1 - 3.f*p2 + p3;
+    const float b =  3.f*p0 - 6.f*p1 + 3.f*p2;
+    const float c = -3.f*p0 + 3.f*p1;
+    const float d =      p0;
+    dst[0] = a;
+    dst[1] = b;
+    dst[2] = c;
+    dst[3] = d;
```
mbouron-gpsw, line 207: I guess the split between the computation and the assignment is intended ?

libnodegl/path.c
```diff
@@ -0,0 +1,408 @@
...
+    struct darray steps;        /* array of struct path_step */
+    struct darray steps_dist;   /* array of floats */
+    float cursor[3];            /* temporary cursor used during path construction */
+    uint32_t segment_flags;     /* temporary segment flags used during path construction */
+};
+
+/*
+ * Interpolate a 3D point using the polynomials
+ */
+static void poly_eval(float t, float *dst, const struct path_segment *segment)
```
mbouron-gpsw, line 64: It looks like the inputs and outputs arguments are mixed. Is it possible to put float *dst at the beginning or the end of the argument list ?

mbouron-gpsw: Moreover, is it possible to move the static function closer to their use + moving ngli_path_evaluate() before ngli_path_freep() (to reflect the order introduced in the header) ?

libnodegl/path.c
```diff
@@ -0,0 +1,408 @@
...
+        /*
+         * We check if we are at the very last step of the last segment, or in
+         * the situation where a move order occurred between the current
+         * segment and the next one. These 2 scenarios imply that we compute
+         * the last point coordinate (t=1) of the current segment because there
+         * won't be an overlap with the next segment (if any).
+         */
+        if (i == nb_segments - 1 || (segments[i + 1].flags & SEGMENT_FLAG_NEW_ORIGIN)) {
+            struct path_step step = {.segment_id=i, .flags = STEP_FLAG_DISCONTINUITY};
+            poly_eval(1.f, step.position, &segments[i]);
```
mbouron-gpsw, line 337: `poly_eval(1.f, step.position, segment);` ?

libnodegl/node_path.c
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2020 GoPro Inc.
```
mbouron-gpsw, line 2: 2021 (or 2020-2021)

libnodegl/node_path.c
```diff
@@ -0,0 +1,104 @@
...
+
+#include <stddef.h>
+#include <string.h>
+
+#include "nodegl.h"
+#include "nodes.h"
+#include "path.h"
+
+struct path_priv {
+    struct path *path;
```
mbouron-gpsw, line 30: Put it after the parameters ?

libnodegl/node_pathkey.c
```diff
@@ -0,0 +1,121 @@
...
+};
+
+#define OFFSET_BEZIER3(x) offsetof(struct pathkey_bezier3_priv, x)
+static const struct node_param pathkey_bezier3_params[] = {
+    {"control1", NGLI_PARAM_TYPE_VEC3, OFFSET_BEZIER3(control1),
+                 .desc=NGLI_DOCSTRING("first control point")},
+    {"control2", NGLI_PARAM_TYPE_VEC3, OFFSET_BEZIER3(control2),
+                 .desc=NGLI_DOCSTRING("second control point")},
+    {"to",       NGLI_PARAM_TYPE_VEC3, OFFSET_BEZIER3(to),
+                 .desc=NGLI_DOCSTRING("end position of the curve, new cursor position")},
```
mbouron-gpsw, line 58: Can you make the description consistent with the bezier2 one ? (`end position` vs `end point`) ?

libnodegl/node_pathkey.c
```diff
@@ -0,0 +1,121 @@
...
+        ngli_bstr_printf(b, "line to:%g,%g,%g", NGLI_ARG_VEC3(s->to));
+    } else if (node->cls->id == NGL_NODE_PATHKEYBEZIER2) {
+        struct pathkey_bezier2_priv *s = node->priv_data;
+        ngli_bstr_printf(b, "bezier2 ctl:%g,%g,%g to:%g,%g,%g",
+                         NGLI_ARG_VEC3(s->control), NGLI_ARG_VEC3(s->to));
+    } else if (node->cls->id == NGL_NODE_PATHKEYBEZIER3) {
+        struct pathkey_bezier3_priv *s = node->priv_data;
+        ngli_bstr_printf(b, "bezier3 ctl1:%g,%g,%g ctl2:%g,%g,%g to:%g,%g,%g",
+                         NGLI_ARG_VEC3(s->control1), NGLI_ARG_VEC3(s->control2), NGLI_ARG_VEC3(s->to));
+    }
```
mbouron-gpsw, line 81: else { ngli_assert(0) ? }

libnodegl/node_smoothpath.c
```diff
@@ -0,0 +1,130 @@
...
+#include <stddef.h>
+#include <string.h>
+
+#include "log.h"
+#include "nodegl.h"
+#include "nodes.h"
+#include "path.h"
+
+struct smoothpath_priv {
+    struct path *path;
```
mbouron-gpsw, line 31: Declare path after parameters ?

libnodegl/path.c
cboesch-gpsw: It helps understanding the function comment mentioning `a`, `b`, `c` and `d`; the alternative would be a comment on the same line, for example:
```c
    dst[0] =     -p0 + 3.f*p1 - 3.f*p2 + p3;  // a
    dst[1] =  3.f*p0 - 6.f*p1 + 3.f*p2;       // b
    dst[2] = -3.f*p0 + 3.f*p1;                // c
    dst[3] =      p0;                         // d
```

Or:
```c
    dst[0 /* a */] =     -p0 + 3.f*p1 - 3.f*p2 + p3;
    dst[1 /* b */] =  3.f*p0 - 6.f*p1 + 3.f*p2;
    dst[2 /* c */] = -3.f*p0 + 3.f*p1;
    dst[3 /* d */] =      p0;
```

Any preference? The latter is a form we used in colorconv typically.

libnodegl/path.c
cboesch-gpsw: The current order is:

```c
static void poly_from_bezier2(float *dst, float p0, float p1, float p2);
static void poly_from_bezier3(float *dst, float p0, float p1, float p2, float p3);

int ngli_path_move_to(struct path *s, const float *to);
static int add_segment_and_move(struct path *s, const struct path_segment *segment, const float *to);
int ngli_path_line_to(struct path *s, const float *to);
int ngli_path_bezier2_to(struct path *s, const float *ctl, const float *to);
int ngli_path_bezier3_to(struct path *s, const float *ctl0, const float *ctl1, const float *to);
```

I'm assuming you would like:

```c
int ngli_path_move_to(struct path *s, const float *to);

static int add_segment_and_move(struct path *s, const struct path_segment *segment, const float *to);
static void poly_from_bezier2(float *dst, float p0, float p1, float p2);
static void poly_from_bezier3(float *dst, float p0, float p1, float p2, float p3);

int ngli_path_line_to(struct path *s, const float *to);
int ngli_path_bezier2_to(struct path *s, const float *ctl, const float *to);
int ngli_path_bezier3_to(struct path *s, const float *ctl0, const float *ctl1, const float *to);
```

?

The rationale was that I didn't want to make `ngli_path_{move,line,bezier2,bezier3}_to` away from each other too much. I was kind of forced with `add_segment_and_move()` since there was initially an inter-dependency during development (not present anymore). I'm planning to go toward:

```c
static void poly_from_bezier2(float *dst, float p0, float p1, float p2);
static void poly_from_bezier3(float *dst, float p0, float p1, float p2, float p3);
static int add_segment_and_move(struct path *s, const struct path_segment *segment, const float *to);

int ngli_path_move_to(struct path *s, const float *to);
int ngli_path_line_to(struct path *s, const float *to);
int ngli_path_bezier2_to(struct path *s, const float *ctl, const float *to);
int ngli_path_bezier3_to(struct path *s, const float *ctl0, const float *ctl1, const float *to);
```

Would that be fine? I like to have `ngli_path_{move,line,bezier2,bezier3}_to` packed together, and `add_segment_and_move` close to `ngli_path_move_to` because those a very similar functions.

libnodegl/node_path.c
cboesch-gpsw:
```c
/* We must have the struct path in 1st position for AnimatedPath */
NGLI_STATIC_ASSERT(path_1st_field, OFFSET(path) == 0);
```

:wink:

libnodegl/node_smoothpath.c
cboesch-gpsw: See my comment on `node_path.c`

libnodegl/path.c
cboesch-gpsw: Changed to last proposition

libnodegl/path.c
cboesch-gpsw: I picked the latter.

libnodegl/path.c
cboesch-gpsw: I shuffled all of this: params are now `segment, dst, t` (context, destination, param)

It is also moved evaluate down to match the header, but I had to keep poly_eval relatively up because it's used in other functions.

libnodegl/path.c
cboesch-gpsw: Yup done

libnodegl/node_path.c
cboesch-gpsw: Fixed

libnodegl/node_pathkey.c
cboesch-gpsw: Fixed

libnodegl/node_pathkey.c
cboesch-gpsw: Added

libnodegl/path.c
mbouron-gpsw: > I shuffled all of this: params are now `segment, dst, t` (context, destination, param)
> 

From my perspective, it feels more natural to have something like poly_eval(dst, segment, t). dst is obviously the output of this function and segment and t are the inputs. I might be missing something but i don't get why segment is considered a "context" here.
We follow the pattern I described in math_utils (for example ngli_vec2_scale(dst, v, s) where v and s are the function inputs, v is the first input argument onto which we applied a scale, the second argument).

tests/transform.py
```diff
@@ -241,3 +240,71 @@ def transform_rotate_quat_animated(cfg, quat0=(0, 0, -0.474, 0.880), quat1=(0, 0
...
+        (-0.8, -0.1,  0.1),
+    )
+    controls = (
+        ( 0.2,  0.3, -0.2),
+        (-0.2, -0.8, -0.4),
+    )
+
+    keyframes = (
+        ngl.PathKeyMove(to=points[0]),
+        ngl.PathKeyBezier3(control1=controls[0], control2=controls[1], to=points[1]),
```
mbouron, line 263: Is it possible to add the other PathKey types to increase coverage ? (Line and Bezier2)

tests/transform.py
```diff
@@ -241,3 +240,71 @@ def transform_rotate_quat_animated(cfg, quat0=(0, 0, -0.474, 0.880), quat1=(0, 0
...
+    )
+    path = ngl.Path(keyframes);
+
+    # We use back_in_out easing to force an overflow on both sides
+    anim_kf = [
+        ngl.AnimKeyFrameFloat(0, 0),
+        ngl.AnimKeyFrameFloat(cfg.duration, 1, 'back_in_out'),
```
mbouron, line 270: Is it possible to reduce the duration of the animation (compared to the scene duration) to cover the path animation copy function ?

tests/transform.py
```diff
@@ -241,3 +240,71 @@ def transform_rotate_quat_animated(cfg, quat0=(0, 0, -0.474, 0.880), quat1=(0, 0
...
+        ( 0.04, -0.27, 0.0),
+        ( 0.36,  0.28, 0.0),
+        ( 0.65, -0.04, 0.0),
+    )
+    controls = (
+        (-0.84, 0.07, 0.0),
+        ( 0.84, 0.04, 0.0),
+    )
+
+    flat_points = (elt for iterable in points for elt in iterable)
```
mbouron, line 295: flat_points = (elt for point in points for elt in point) ?

tests/transform.py
cboesch-gpsw: Added. It shouldn't change the coverage much though, since lines and bezier2 are tested internally. It may add the coverage of the info strings in PathKey though.

tests/transform.py
cboesch-gpsw: Done.

tests/transform.py
cboesch-gpsw: Done

tests/transform.py
mbouron: It should also increase the coverage of Path.

libnodegl/path.c
cboesch-gpsw: Done

mbouron-gpsw (approved):
Good job, LGTM.

PR #290, opened 2021-06-07
Add YUV support

mbouron-gpsw:
Depends on https://github.com/Stupeflix/sxplayer/pull/31.


libnodegl/hwupload_common.c
```diff
@@ -28,21 +28,81 @@
...
+    int layout;
+    int nb_planes;
+    int log2_chroma_width;
+    int log2_chroma_height;
+    int formats[4];
+} format_descs[] = {
+    [SXPLAYER_PIXFMT_RGBA] = {
+        .layout = NGLI_IMAGE_LAYOUT_DEFAULT,
+        .nb_planes = 1,
+        .formats[0] = NGLI_FORMAT_B8G8R8A8_UNORM,
```
jmoguillansky-gpsw, line 52: why are these inverted?

libnodegl/hwupload_common.c
```diff
@@ -28,21 +28,81 @@
...
+    },
+    [SXPLAYER_PIXFMT_BGRA] = {
+        .layout = NGLI_IMAGE_LAYOUT_DEFAULT,
+        .nb_planes = 1,
+        .formats[0] = NGLI_FORMAT_R8G8B8A8_UNORM,
```
jmoguillansky-gpsw, line 57: inverted

libnodegl/hwupload.c
```diff
@@ -52,6 +52,10 @@ static const struct hwmap_class *hwupload_gl_class_map[] = {
 #elif defined(HAVE_VAAPI)
     [SXPLAYER_PIXFMT_VAAPI]       = &ngli_hwmap_vaapi_gl_class,
 #endif
+    [SXPLAYER_PIXFMT_NV12]        = &ngli_hwmap_common_class,
+    [SXPLAYER_PIXFMT_YUV420P]     = &ngli_hwmap_common_class,
```
jmoguillansky-gpsw, line 56: consider adding support for a "flexible format" (e.g. planar / interleaved, etc)

see https://android.googlesource.com/platform/system/core/+/e49b696/include/system/graphics.h
as reference

libnodegl/hwupload_common.c
mbouron-gpsw: Typo, fixed.

libnodegl/hwupload_common.c
mbouron-gpsw: Typo, fixed.

libnodegl/hwupload.c
mbouron-gpsw: If I understand correctly the android "flexible format" is an abstraction describing how to handle multiple formats (yuv*,rgb*). It looks like supporting a "flexible format" needs to be done at the sxplayer level and means another change of API (like adding the relevant fields describing subsampling, bit depth, etc).

My plan is to gradually adds support for new formats after this PR (yuv 10-bit formats for example).  I'll consider unifying them once their support is done.

libnodegl/hwupload_common.c
```diff
@@ -74,32 +84,45 @@ static int common_init(struct ngl_node *node, struct sxplayer_frame *frame)
...
     }
 
     const int linesize = frame->linesize >> 2;
-    return ngli_texture_upload(s->texture, frame->data, linesize);
+    return ngli_texture_upload(common->planes[0], frame->data, linesize);
 }
 
 const struct hwmap_class ngli_hwmap_common_class = {
     .name      = "default",
+    .priv_size = sizeof(struct hwupload_common),
```
cboesch-gpsw, line 197: After this commit  `hwmap_class->priv_size` is always defined, we may be able to make a simplification to `hwupload.c`.

I'm also seeing that `hwupload.c` is currently allocating an `s->texture` (but I'm probably confused what's going on here). One thing though, it appears that no other `hwupload` implementation alter `s->texture` anymore: can we have make either some simplifications and/or more non-mutability assumptions about the `hwupload_*` now?

ngl-tools/ngl-player.c
```diff
@@ -54,6 +55,7 @@ static const struct opt options[] = {
     {"-m", "--samples",          OPT_TYPE_INT,      .offset=OFFSET(cfg.samples)},
     {"-u", "--disable-ui",       OPT_TYPE_TOGGLE,   .offset=OFFSET(player_ui)},
     {"-r", "--framerate",        OPT_TYPE_RATIONAL, .offset=OFFSET(framerate)},
+    {NULL, "--disable-hwaccel",  OPT_TYPE_TOGGLE,   .offset=OFFSET(hwaccel)},
```
cboesch-gpsw, line 59: `"--hwaccel"` so that we can extend it more easily?

ngl-tools/ngl-player.c
```diff
@@ -56,6 +57,7 @@ static const struct opt options[] = {
     {"-u", "--disable-ui",       OPT_TYPE_TOGGLE,   .offset=OFFSET(player_ui)},
     {"-r", "--framerate",        OPT_TYPE_RATIONAL, .offset=OFFSET(framerate)},
     {NULL, "--disable-hwaccel",  OPT_TYPE_TOGGLE,   .offset=OFFSET(hwaccel)},
+    {NULL, "--enable-mipmap",    OPT_TYPE_TOGGLE,   .offset=OFFSET(mipmap)},
```
cboesch-gpsw, line 60: Same: `"--mipmap"` is preferred here in case we want to extend that

libnodegl/hwupload_common.c
```diff
@@ -59,34 +113,45 @@ static int common_init(struct ngl_node *node, struct sxplayer_frame *frame)
...
 
-    hwupload->require_hwconv = 0;
+    int direct_rendering = 1;
+    if (desc->layout != NGLI_IMAGE_LAYOUT_DEFAULT) {
+        direct_rendering = (s->supported_image_layouts & (1 << desc->layout));
+        if (s->params.mipmap_filter)
+            direct_rendering = 0;
+    }
+
+    hwupload->require_hwconv = !direct_rendering;
```
cboesch-gpsw, line 154: Do you mind following the pattern of the other hwupload implementation? That is, using a `support_direct_rendering()` function?

libnodegl/hwupload_common.c
```diff
@@ -115,8 +180,16 @@ static int common_map_frame(struct ngl_node *node, struct sxplayer_frame *frame)
             return ret;
     }
 
-    const int linesize = frame->linesize >> 2;
-    return ngli_texture_upload(common->planes[0], frame->data, linesize);
+    for (int i = 0; i < common->nb_planes; i++) {
+        struct texture *plane = common->planes[i];
+        struct texture_params *params = &plane->params;
+        const int linesize = frame->linesizep[i] / ngli_format_get_bytes_per_pixel(params->format);
```
cboesch-gpsw, line 186: Not blocking but I think we could replace `format_comp_sizes.size` with `format_comp_sizes.nbits` (and have `ngli_format_get_bytes_per_pixel()` return `1<<nbits`). The benefit I see is that we won't risk to end up with a non-exact division (if one format size can not be expressed with a bit shift we have an issue). And bonus, we will obviously have the shift op back.

ngl-tools/ngl-player.c
mbouron-gpsw: Done.

ngl-tools/ngl-player.c
mbouron-gpsw: Done.

libnodegl/hwupload_common.c
mbouron-gpsw: Done.

libnodegl/hwupload_common.c
mbouron-gpsw: > After this commit `hwmap_class->priv_size` is always defined, we may be able to make a simplification to `hwupload.c`.
> 

Done as a separate commit.

> I'm also seeing that `hwupload.c` is currently allocating an `s->texture` (but I'm probably confused what's going on here). One thing though, it appears that no other `hwupload` implementation alter `s->texture` anymore: can we have make either some simplifications and/or more non-mutability assumptions about the `hwupload_*` now?

`s->texture` is allocated by the hwconv part of hwupload. Moreover, even though we do not use s->texture directly in  the hwupload_ implementation we are still using s->image, so we can't make `struct ngl_node *node` const. The whole hwupload API needs to be updated to avoid using the `s->{texture,image,...}` and have proper inputs/outputs defined. I can work on that later on.

cboesch-gpsw (approved):

PR #293, opened 2021-06-24
pgcraft: add NGLI_PGCRAFT_SHADER_TEX_TYPE_VIDEO

mbouron-gpsw:


libnodegl/pgcraft.c
```diff
@@ -743,8 +743,7 @@ static int samplers_preproc(struct pgcraft *s, struct bstr *b)
     while ((p = strstr(p, "ngl"))) {
         struct token token = {.pos = p - base_str};
         p = read_token_id(p, token.id, sizeof(token.id));
-        if (strcmp(token.id, "ngl_texvideo") &&
-            strcmp(token.id, "ngli_texvideo"))
+        if (strcmp(token.id, "ngl_texvideo"))
```
cboesch-gpsw, line 767: technically now you should be able to `strstr(p, "ngl_texvideo")` in the `while` loop above

libnodegl/pgcraft.c
```diff
@@ -653,7 +653,7 @@ struct token {
 
 #define ARG_FMT(x) (int)x##_len, x##_start
 
-static int handle_token(struct pgcraft *s, const struct token *token, const char *p, struct bstr *dst)
+static int handle_token(struct pgcraft *s, const struct pgcraft_params *params, const struct token *token, const char *p, struct bstr *dst)
```
cboesch-gpsw, line 656: do you mind line breaking after `params`?

libnodegl/pgcraft.c
```diff
@@ -668,6 +668,15 @@ static int handle_token(struct pgcraft *s, const struct token *token, const char
...
     ptrdiff_t arg0_len = p - arg0_start;
 
+    enum pgcraft_shader_tex_type texture_type = NGLI_PGCRAFT_SHADER_TEX_TYPE_NONE;
+    for (int i = 0; i < params->nb_textures; i++) {
+        const struct pgcraft_texture *pgcraft_texture = &params->textures[i];
+        if (!strncmp(arg0_start, pgcraft_texture->name, arg0_len)) {
+            texture_type = pgcraft_texture->type;
+            break;
+        }
+    }
```
cboesch-gpsw, line 678: i'm really not a fan of break when they can be avoided: can you put that in a function where you simply `return pgcraft_texture->type` or `NGLI_PGCRAFT_SHADER_TEX_TYPE_NONE` in the default case?

(and you can have a `texture_type` made `const` in this function as a result).

cboesch-gpsw (changes requested):
Only cosmetics remarks, thanks

libnodegl/pgcraft.c
mbouron-gpsw: Done.

libnodegl/pgcraft.c
mbouron-gpsw: Done.

cboesch-gpsw (approved):

PR #295, opened 2021-06-28
hwupload: only retrieve hwmap class on frame properties change

mbouron-gpsw:


cboesch-gpsw (commented):
LGTM functionally but I would have appreciated a few more elaborated descriptions.

For the first one the following info:
1. saving the `get_hwmap_class()` call because that might become expensive in the near future
2. the `hwmap_class != hwupload->hwmap_class` condition being unneeded as `hwmap_class` is deterministic and solely depending on the frame pixel format (and non-switchable backend). It's (close to being?) a pure function.

For the 2nd commit:
1. not limiting the handling of common pixel format to `BACKEND_GL` (needs context to figure out the blocks)
2. problem not visible currently as only GL is implemented

While being "obvious" it took me a while to actually figure out the details even though I was aware of the hidden motivation of these changes.

mbouron-gpsw:
> LGTM functionally but I would have appreciated a few more elaborated descriptions.
> 
> For the first one the following info:
> 
>     1. saving the `get_hwmap_class()` call because that might become expensive in the near future
> 
>     2. the `hwmap_class != hwupload->hwmap_class` condition being unneeded as `hwmap_class` is deterministic and solely depending on the frame pixel format (and non-switchable backend). It's (close to being?) a pure function.
> 
> 
> For the 2nd commit:
> 
>     1. not limiting the handling of common pixel format to `BACKEND_GL` (needs context to figure out the blocks)
> 
>     2. problem not visible currently as only GL is implemented
> 
> 
> While being "obvious" it took me a while to actually figure out the details even though I was aware of the hidden motivation of these changes.

The commit description have been updated accordingly.

cboesch-gpsw (approved):

PR #296, opened 2021-07-04
media: move the android texture allocation to hwupload_mediacodec_gl

mbouron-gpsw:


libnodegl/node_media.c
```diff
@@ -211,8 +190,7 @@ static int media_init(struct ngl_node *node)
             if (!handler)
                 return NGL_ERROR_EXTERNAL;
 
-            struct texture_gl *texture_gl = (struct texture_gl *)s->android_texture;
-            s->android_surface = ngli_android_surface_new(texture_gl->id, handler);
+            s->android_surface = ngli_android_surface_new(0, handler);
```
cboesch-gpsw, line 193: Can you link to a reference showing that `texName=0` implies that `SurfaceTexture` is created in detached mode? I might not be looking into the correct place, but the doc seems to mention that only `public SurfaceTexture (boolean singleBufferMode)` can be used for the detached mode. It's likely that they use `texName=0` internally in this one to do that, I'm just a bit uncomfortable without a double check because texture name 0 has a special meaning in GL.

Please also add a small sentence in the commit description mentioning that we are now in detached mode by default in that codepath, with the texture attached later through `ngli_android_surface_attach_to_gl_context()`/`attachToGLContext()`

libnodegl/node_media.c
```diff
@@ -344,7 +322,6 @@ static void media_uninit(struct ngl_node *node)
         ngli_android_surface_free(&s->android_surface);
         ngli_android_handlerthread_free(&s->android_handlerthread);
     }
-    ngli_texture_freep(&s->android_texture);
```
cboesch-gpsw, line 347: I'm not seeing a `ngli_texture_freep(&mc->texture)`; did I miss something?

mbouron-gpsw: No, I forgot to add it in the mediacodec hwmap module. Fixed.

libnodegl/node_media.c
mbouron-gpsw: The commit description has been updated to include the relevant explanation.

cboesch-gpsw (approved):

PR #297, opened 2021-07-05
Hwupload improvements - part 1

mbouron-gpsw:
The goal of this PR is to progressively change the API of the hwupload module to move away the API from depending on the internal nodes structures. It also includes a bunch of fixes for the videotoolbox hwupload modules.

### DIFF
--- a/libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
+++ b/libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
@@ -206,14 +206,15 @@ static int support_direct_rendering(struct ngl_node *node, struct sxplayer_frame
         break;
     case kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange:
         direct_rendering = s->supported_image_layouts & (1 << NGLI_IMAGE_LAYOUT_NV12);
-        if (direct_rendering && s->params.mipmap_filter) {
-            LOG(WARNING, "IOSurface NV12 buffers do not support mipmapping: "
-                "disabling direct rendering");
-            direct_rendering = 0;
-        }
         break;
     default:
         ngli_assert(0);
+    }
+
+    if (direct_rendering && s->params.mipmap_filter) {
+        LOG(WARNING, "Videotoolbox textures do not support mipmapping: "
+            "disabling direct rendering");
+        direct_rendering = 0;
     }
 
     return direct_rendering;
@@ -230,20 +231,15 @@ static int vt_ios_init(struct ngl_node *node, struct sxplayer_frame *frame)
     CVPixelBufferRef cvpixbuf = (CVPixelBufferRef)frame->data;
     vt->format = CVPixelBufferGetPixelFormatType(cvpixbuf);
 
-    struct texture_params plane_params = s->params;
-    if (plane_params.mipmap_filter) {
-        LOG(WARNING, "IOSurface RGBA/BGRA buffers do not support mipmapping: "
-            "disabling mipmapping");
-        plane_params.mipmap_filter = NGLI_MIPMAP_FILTER_NONE;
-    }
-
     struct format_desc format_desc = {0};
     int ret = vt_get_format_desc(vt->format, &format_desc);
     if (ret < 0)
         return ret;
 
     for (int i = 0; i < format_desc.nb_planes; i++) {
+        struct texture_params plane_params = s->params;
         plane_params.format = format_desc.planes[i].format;
+        plane_params.mipmap_filter = NGLI_MIPMAP_FILTER_NONE;
         plane_params.usage = NGLI_TEXTURE_USAGE_SAMPLED_BIT;
 
         vt->planes[i] = ngli_texture_create(gpu_ctx);
--- a/libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
+++ b/libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
@@ -237,10 +237,17 @@ static int vt_ios_init(struct ngl_node *node, struct sxplayer_frame *frame)
         return ret;
 
     for (int i = 0; i < format_desc.nb_planes; i++) {
-        struct texture_params plane_params = s->params;
-        plane_params.format = format_desc.planes[i].format;
-        plane_params.mipmap_filter = NGLI_MIPMAP_FILTER_NONE;
-        plane_params.usage = NGLI_TEXTURE_USAGE_SAMPLED_BIT;
+        const struct texture_params texture_params = &s->params;
+
+        const struct texture_params plane_params = {
+            .type             = NGLI_TEXTURE_TYPE_2D,
+            .format           = format_desc.planes[i].format;
+            .min_filter       = texture_params->min_filter,
+            .mag_filter       = texture_params->mag_filter,
+            .wrap_s           = texture_params->wrap_s,
+            .wrap_t           = texture_params->wrap_t,
+            .usage            = NGLI_TEXTURE_USAGE_SAMPLED_BIT,
+        };
 
         vt->planes[i] = ngli_texture_create(gpu_ctx);
         if (!vt->planes[i])
--- a/libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
+++ b/libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
@@ -198,13 +198,14 @@ static int support_direct_rendering(struct ngl_node *node, struct sxplayer_frame
 
     CVPixelBufferRef cvpixbuf = (CVPixelBufferRef)frame->data;
     OSType cvformat = CVPixelBufferGetPixelFormatType(cvpixbuf);
-    int direct_rendering = s->supported_image_layouts & (1 << NGLI_IMAGE_LAYOUT_NV12);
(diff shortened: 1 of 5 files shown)


libnodegl/node_texture.c
```diff
@@ -357,7 +357,9 @@ static void handle_buffer_frame(struct ngl_node *node)
     struct buffer_priv *buffer = s->data_src->priv_data;
     const uint8_t *data = buffer->data;
 
-    ngli_texture_upload(s->texture, data, 0);
+    int ret = ngli_texture_upload(s->texture, data, 0);
+    if (ret < 0)
+        LOG(ERROR, "could not upload texture buffer");
```
cboesch-gpsw, line 375: This is called by `texture_update()` which does return an error code: can we take the opportunity to add a commit that checks and raise the errors from `handle_buffer_frame()` and `handle_media_frame()`?

libnodegl/node_texture.c
```diff
@@ -346,7 +346,16 @@ static int texture_prefetch(struct ngl_node *node)
...
 static void handle_media_frame(struct ngl_node *node)
 {
-    int ret = ngli_hwupload_upload_frame(node);
+    struct texture_priv *s = node->priv_data;
+    struct media_priv *media = s->data_src->priv_data;
+    struct sxplayer_frame *frame = media->frame;
+    if (!frame)
+        return;
+
+    /* Transfer frame ownership to hwupload and prevent other textures from using it */
```
cboesch-gpsw, line 355: Do you mean other texture *nodes*? That shouldn't be possible, right?

libnodegl/hwupload.c
```diff
@@ -145,7 +145,7 @@ static int exec_hwconv(struct ngl_node *node)
     return 0;
 }
 
-int ngli_hwupload_upload_frame(struct ngl_node *node, struct sxplayer_frame *frame)
+int ngli_hwupload_upload_frame(struct ngl_node *node, struct sxplayer_frame *frame, struct image *image)
```
cboesch-gpsw: `image_dst`? (feel free to ignore)

libnodegl/node_texture.c
```diff
@@ -355,9 +355,13 @@ static void handle_media_frame(struct ngl_node *node)
...
     media->frame = NULL;
 
-    int ret = ngli_hwupload_upload_frame(node, frame);
+    /* Reset destination image */
+    ngli_image_reset(&s->image);
+
+    int ret = ngli_hwupload_upload_frame(node, frame, &s->image);
     if (ret < 0)
         LOG(ERROR, "could not map media frame");
+    s->image.ts = frame->ts;
```
cboesch-gpsw: This is dangerous: if `ngli_hwupload_upload_frame()` fails, it resets `frame`, so you don't want to access `frame->ts` in this situation. This might be fixed if we raise the error instead of tolerating it as mentioned in an earlier comment.

libnodegl/hwupload.c
```diff
@@ -191,12 +191,12 @@ int ngli_hwupload_upload_frame(struct ngl_node *node, struct sxplayer_frame *fra
...
             ret = init_hwconv(node);
             if (ret < 0)
-                return ret;
+                goto end;
             hwupload->hwconv_initialized = 1;
         }
         ret = exec_hwconv(node);
         if (ret < 0)
-            return ret;
+            goto end;
```
cboesch-gpsw, line 199: Not exactly related to this commit, but in this codepath `frame` will be released if and only if `HWMAP_FLAG_FRAME_OWNER` is set: don't we need that as well in the previous `sxplayer_release_frame(frame)` calls in that function?

libnodegl/hwupload.h
```diff
@@ -54,7 +54,7 @@ struct hwupload {
     int hwconv_initialized;
 };
 
-int ngli_hwupload_upload_frame(struct ngl_node *node, struct sxplayer_frame *frame, struct image *image);
+int ngli_hwupload_map_frame(struct ngl_node *node, struct sxplayer_frame *frame, struct image *image);
```
cboesch-gpsw: This is an interesting rename because I was just wondering about the confusion between `hwupload`, `hwmap` and `hwconv`, and I was assuming `hwmap`/`hwconv` were the close/identical concepts, not `hwupload`/`hwmap`. Can we make sure we agree on the semantic of the 3 entities?

libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
```diff
@@ -237,10 +237,17 @@ static int vt_ios_init(struct ngl_node *node, struct sxplayer_frame *frame)
...
+
+        const struct texture_params plane_params = {
+            .type             = NGLI_TEXTURE_TYPE_2D,
+            .format           = format_desc.planes[i].format;
+            .min_filter       = texture_params->min_filter,
+            .mag_filter       = texture_params->mag_filter,
+            .wrap_s           = texture_params->wrap_s,
+            .wrap_t           = texture_params->wrap_t,
+            .usage            = NGLI_TEXTURE_USAGE_SAMPLED_BIT,
+        };
```
cboesch-gpsw, line 250: what happened to `mipmap_filter`? Are you relying on the fact that it's `0`? If so, please mention it in the commit description.

libnodegl/node_texture.c
mbouron-gpsw: This needs investigations. Do we want to stop rendering of the entire frame if we fail to map a media frame (instead of using a dummy black texture) ?

libnodegl/hwupload.c
mbouron-gpsw: We don't need this check in the previous call to sxplayer_release_frame(frame) because these functions are called during the initialization of the underlying mapper (and no frame has been mapped/owned yet by a hwmap implementation).

libnodegl/backends/gl/hwupload_videotoolbox_ios_gl.c
mbouron-gpsw: We are relying in the fact that it's 0 (MIPMAP_FILTER_NONE). We use this pattern in a lot of place (where we do not need mipmapping) in the code base. I've updated the commit description.

libnodegl/hwupload.h
mbouron-gpsw: I've removed this commit from the PR as this rename does not feel enough to me, we probably want to also rename the hwupload module to hwmap and move the hwconv code path outside of it.

libnodegl/node_texture.c
mbouron-gpsw: I've removed this chunk and kept the image.ts assignment in hwupload.c.

libnodegl/node_texture.c
mbouron-gpsw: I've reworded this comment to make it more accurate as we just want to ensure that media->frame cannot be re-used later on by any module of our code (and in particular in node_media.c/media_release()).

libnodegl/hwupload.c
cboesch-gpsw: I see. But then what happens if the mapper flagged as a frame owner fails to initialize? Aren't we going to skip the frame release even though the mapper instance technically doesn't own it?

libnodegl/node_texture.c
cboesch-gpsw: Maybe not; in that case maybe we should print the message as a warning instead to show that's not intended to be a hard fail?

libnodegl/hwupload.c
mbouron-gpsw: If the mapper flagged as a frame owner fails to initialize we currently release the media frame and error out (hwmap_class.init() does not take ownership of the frame). On the other hand, hwmap_class.map_frame() takes ownership of the frame whatever happens, it is up the map_frame() implementation to release the underlying resource when needed.
What happens in practice the vaapi and videotoolbox (darwin) implementations, is the following pattern placed at the very beginning of the function:
```c
    sxplayer_release_frame(vt->frame);
    vt->frame = frame;
```
so even if an error happens, the mapper release the previous frame and keeps a reference on the new one.

libnodegl/node_texture.c
mbouron-gpsw: Even though we are not stopping rendering, I think we should keep the messages as errors for now as they are quite important, more than just a regular warning depending on the root cause. I'm not decided yet on whether we should stop rendering or not.

libnodegl/hwupload.c
cboesch-gpsw: OK makes sense now, thanks!

libnodegl/node_texture.c
cboesch-gpsw: OK well, I just want to make sure the next person reading this code doesn't think this is an overlooked issue. I felt like a warning was transmitting this idea well, but a comment would work as well (`/* Tolerate texture upload failure for media sources because they're a "likely" errors where we prefer to black-out part of the presentation instead of hard-failing */` or whatever)

mbouron-gpsw: I've updated the first commit message and added a new one. The first commit makes the code error out if there is an issue while uploading a regular buffer to the texture. The second one makes it more explicit why we ignore media frames mapping/upload failures.

cboesch-gpsw (approved):

PR #299, opened 2021-07-19
Hwupload improvements - part 2

mbouron-gpsw:
The goal of this PR is to improve and simplify the declaration of new hwmap classes for the upcoming graphics backends.


libnodegl/hwupload.c
```diff
@@ -38,41 +38,35 @@ extern const struct hwmap_class ngli_hwmap_vt_darwin_gl_class;
...
-    if (backend == NGL_BACKEND_OPENGL || backend == NGL_BACKEND_OPENGLES) {
-        if (frame->pix_fmt < 0 || frame->pix_fmt >= NGLI_ARRAY_NB(hwupload_gl_class_map))
-            return NULL;
+    const struct hwmap_class **hwmap_classes = NULL;
 
-        return hwupload_gl_class_map[frame->pix_fmt];
+    if (backend == NGL_BACKEND_OPENGL || backend == NGL_BACKEND_OPENGLES)
+        hwmap_classes = hwupload_gl_classes;
+
+    for (int i = 0; hwmap_classes && hwmap_classes[i]; i++) {
```
cboesch-gpsw, line 45: I'm not sure compilers are smart enough to avoid checking `hwmap_classes` at every loop; do you mind moving the check in a dedicated `if` above the `for` loop?

cboesch-gpsw (approved):
Aside from the nitpick, LGTM, thanks

libnodegl/hwupload.c
mbouron-gpsw: Updated. Let me know if it is OK for you (or if you prefer a `goto` variant, i'm not really satisfied with the current version though).

cboesch-gpsw: yup LGTM thanks; i'm fine with your current proposition or a `if` around the `for`

cboesch-gpsw: I assume you don't like it because the last return code is certain to error out later on in hwupload_common?

mbouron-gpsw: I'll go with the if surrounding the for loop. I don't like the current pattern because we return the fallback hwmap class at two different places. Thanks for the suggestion.

mbouron-gpsw: Updated.

PR #301, opened 2021-07-19
Ordered hmap

cboesch-gpsw:


libnodegl/test_hmap.c
```diff
@@ -41,20 +41,39 @@ static void free_func(void *arg, void *data)
...
+static const struct {
+    const char *key;
+    const char *val;
+} kvs[] = {
+    {"foo",     "bar"},
+    {"hello",   "world"},
+    {"lorem",   "ipsum"},
+    {"bazbaz",  ""},
+    {"abc",     "def"},
+    {"last",    "samurai"},
```
mbouron-gpsw, line 53: Can you also check that overriding a existing key does not change the order ?

cboesch-gpsw: done

libnodegl/hmap.c
```diff
@@ -68,6 +74,52 @@ int ngli_hmap_count(const struct hmap *hm)
...
+        .entry_id  = (int)(entry - hm->buckets[ref.bucket_id].entries),
+    };
+    return ref;
+}
+
+static struct hmap_entry *fixed_entry_from_ref(const struct hmap *hm, struct hmap_ref ref, struct hmap_ref removed)
+{
+    const int need_fix = ref.bucket_id == removed.bucket_id && ref.entry_id > removed.entry_id;
+    const int entry_id = ref.entry_id - need_fix;
+    return HAS_REF(ref) ? &hm->buckets[ref.bucket_id].entries[entry_id] : NULL;
```
mbouron-gpsw, line 97: Can we early return here (so we avoid computing need_fix and entry_id if there is no ref) ?

libnodegl/hmap.c
```diff
@@ -68,6 +74,52 @@ int ngli_hmap_count(const struct hmap *hm)
     return hm->count;
 }
 
+static struct hmap_entry *entry_from_ref(const struct hmap *hm, struct hmap_ref ref)
```
mbouron-gpsw, line 77: We usually don't pass the structures by copy in the rest of the codebase, I guess you made this choice because the size of the structure hmap_ref is the same as a pointer on 64-bit systems ?

libnodegl/hmap.c
```diff
@@ -68,6 +74,52 @@ int ngli_hmap_count(const struct hmap *hm)
...
+{
+    return HAS_REF(ref) ? &hm->buckets[ref.bucket_id].entries[ref.entry_id] : NULL;
+}
+
+static struct hmap_ref ref_from_entry(const struct hmap *hm, const struct hmap_entry *entry)
```
mbouron-gpsw, line 82: Is this pattern now allowed across the codebase ? (returning a struct by copy)

libnodegl/hmap.c
cboesch-gpsw: Good point, done

libnodegl/hmap.c
cboesch-gpsw: I assumed for a small struct with a non-exposed API, it was not an issue (the copy doesn't actually happen).

libnodegl/hmap.c
cboesch-gpsw: ¯\_(ツ)_/¯

Your call. I believe small struct + local scope is fine.

libnodegl/hmap.c
mbouron-gpsw: Ok.

libnodegl/hmap.c
mbouron-gpsw: I'm fine with this pattern, I wanted to use it in other part of the codebase some time ago.

mbouron-gpsw (approved):

PR #305, opened 2021-07-29
Misc hooks fixes

mbouron-gpsw:


pynodegl-utils/pynodegl_utils/hooks.py
```diff
@@ -52,14 +52,17 @@ def get_sessions(self):
...
-            ))
+            try:
+                # TODO: asynchronous calls? (individual session queries could take some time)
+                session_info = self._module.get_session_info(session_id)
+                sessions.append((
+                    session_id,
+                    session_desc,
+                    session_info.get('backend', ''),
+                    session_info.get('system', ''),
+                ))
```
ubitux, line 63: Can you keep this part out of the try catch? Same in the other commit.

LGTM otherwise, thanks

mbouron-gpsw: Done. I used a continue in the `except` block. Not sure if that's the preferred form.

cboesch-gpsw: Thanks; I would have picked that form as well to reduce the indent level for the valid codepath

cboesch-gpsw (approved):

pynodegl-utils/pynodegl_utils/hooks.py
mbouron-gpsw: Thanks !

PR #312, opened 2021-09-13
pass: fix buffer usage flags if the buffer is writable

mbouron-gpsw:


libnodegl/pass.c
```diff
@@ -209,8 +209,10 @@ static int register_block(struct pass *s, const char *name, struct ngl_node *blo
         const struct ngl_node *resprops_node = ngli_hmap_get(params->properties, name);
         if (resprops_node) {
             const struct resourceprops_priv *resprops = resprops_node->priv_data;
-            if (resprops->variadic || resprops->writable)
+            if (resprops->variadic || resprops->writable) {
                 type = NGLI_TYPE_STORAGE_BUFFER;
+                usage = NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;
+            }
```
cboesch-gpsw, line 215: Instead of 3 pairs of (type, usage) set to the same thing and risking of missing one (like how it happened here), why not do below:

```c
if (type == NGLI_TYPE_STORAGE_BUFFER)
    block_priv->usage |= NGLI_BUFFER_USAGE_STORAGE_BUFFER_BIT;
else if (type == NGLI_TYPE_UNIFORM_BUFFER)
    block_priv->usage |= NGLI_BUFFER_USAGE_UNIFORM_BUFFER_BIT;
```

tests/compute.py
```diff
@@ -291,4 +289,17 @@ def compute_animation(cfg):
     render = ngl.Render(geometry, program)
     render.update_frag_resources(color=ngl.UniformVec4(value=COLORS.sgreen))
 
-    return ngl.Group(children=(compute, render))
+    children = (compute, render) if animate_pre_render else (render, compute)
```
cboesch-gpsw, line 292: This is not going to cause reading uninitialized values at the first iteration?

mbouron-gpsw: Nope, output_block (as well as input_block) are already initialized with the same data at the beginning of this function.

libnodegl/pass.c
mbouron-gpsw: Done in a separate commit.

cboesch-gpsw: You could remove the `{` `}` added in previous commit but LGTM

tests/compute.py
cboesch-gpsw: OK

cboesch-gpsw (approved):

libnodegl/pass.c
mbouron-gpsw: Done.

cboesch-gpsw: Thanks

PR #318, opened 2021-10-04
Prepare callback fixes

mbouron-gpsw:


libnodegl/node_texture.c
```diff
@@ -470,6 +470,17 @@ static int texture2d_init(struct ngl_node *node)
     s->params.format = get_preferred_format(gpu_ctx, s->format);
     s->supported_image_layouts = s->direct_rendering ? -1 : (1 << NGLI_IMAGE_LAYOUT_DEFAULT);
 
+    /* Check that the underlying media node is not shared across multiple textures */
```
cboesch-gpsw, line 473: can  you keep the Android explanation about why we don't do this?

cboesch-gpsw (commented):
I think it makes sense to have the prepare-on-all-children done automatically at API level and keep the callbacks simple. OK as-is too but it's kind of a lot of self-imposed boilerplate, in particular since we make mandatory to nodes to do it all the time.

libnodegl/node_texture.c
mbouron-gpsw: Done.

mbouron-gpsw:
> I think it makes sense to have the prepare-on-all-children done automatically at API level and keep the callbacks simple. OK as-is too but it's kind of a lot of self-imposed boilerplate, in particular since we make mandatory to nodes to do it all the time.

Unfortunately we can't do that because we need a different state for each branch of the graph (see group_prepare() implementation).

cboesch-gpsw (approved):

PR #333, opened 2021-11-03
More efficient resource tracker

cboesch-gpsw:


pynodegl-utils/pynodegl_utils/resourcetracker.py
```diff
@@ -77,3 +77,18 @@ def end_hooking(self):
...
+
+
+if __name__ == '__main__':
+    import pkgutil
+
+    tracker = ResourceTracker()
+    tracker.start_hooking()
+
+    open(op.join(op.dirname(__file__), 'examples', 'shaders', 'texture.vert'))
+    pkgutil.get_data('pynodegl_utils.examples.shaders', 'texture.frag')
```
mbouron-gpsw, line 89: Would it be worth to use pkgutil.get_data in cfg.get_{frag,vert,comp} ? If so, would it make sense to split to addition of the __init__.py in a separate commit stating it allows the future use of pkgutil.get_data() ?

cboesch-gpsw: Done, see `utils/misc: use pkgutil.get_data() to load shaders`

mbouron-gpsw: Thanks.

mbouron-gpsw (approved):

PR #337, opened 2021-11-09
Fix random GUI scene option widgets reset

cboesch-gpsw:


pynodegl-utils/pynodegl_utils/resourcetracker.py
```diff
@@ -37,7 +39,7 @@ def __init__(self):
         self._pysysdir = op.realpath(sysconfig.get_paths()['stdlib'])
 
     def _register_file(self, file):
-        if not op.isfile(file):
+        if not op.isfile(file) or op.realpath(file) == op.realpath(Config.FILEPATH):
```
mbouron-gpsw, line 42: nit: store op.realpath(file) and reuse it for `self.filelist.update([op.realpath(file)])`

mbouron-gpsw:
In the commit description of `utils/resourcetracker: blacklist config file`, can you mention that the 1s timeout config slot can be concurrent with query_inplace  because query_inplace() is executed from multiple QThread (used by the HooksController) ?

mbouron-gpsw (approved):
Other than my 2 remarks, LGTM.

pynodegl-utils/pynodegl_utils/resourcetracker.py
cboesch-gpsw: Code reworked

cboesch-gpsw:
> In the commit description of `utils/resourcetracker: blacklist config file`, can you mention that the 1s timeout config slot can be concurrent with query_inplace because query_inplace() is executed from multiple QThread (used by the HooksController) ?

Commit description expanded.

mbouron-gpsw:
> > In the commit description of `utils/resourcetracker: blacklist config file`, can you mention that the 1s timeout config slot can be concurrent with query_inplace because query_inplace() is executed from multiple QThread (used by the HooksController) ?
> 
> Commit description expanded.

LGTM, thanks.

PR #340, opened 2021-11-16
Misc fixes

mbouron-gpsw:


libnodegl/backends/gl/glcontext_eagl.m
```diff
@@ -74,6 +74,11 @@ static int eagl_init(struct glcontext *ctx, uintptr_t display, uintptr_t window,
 {
     struct eagl_priv *eagl = ctx->priv_data;
 
+    if (ctx->backend != NGL_BACKEND_OPENGLES) {
+        LOG(ERROR, "unsupported backend: %d, only OpenGLES is supported by EAGL", ctx->backend);
+        return -1;
+    }
```
cboesch-gpsw, line 80: Not blocking because the whole file is full of `return -1`, but this one is obviously meant to be `NGL_ERROR_GRAPHICS_UNSUPPORTED`

libnodegl/backends/gl/hwmap_videotoolbox_darwin_gl.c
```diff
@@ -162,6 +162,11 @@ static int support_direct_rendering(struct hwmap *hwmap, struct sxplayer_frame *
             LOG(WARNING, "Videotoolbox textures do not support mipmapping: "
                 "disabling direct rendering");
             direct_rendering = 0;
+        } else if (params->texture_wrap_s != NGLI_WRAP_CLAMP_TO_EDGE ||
+                   params->texture_wrap_t != NGLI_WRAP_CLAMP_TO_EDGE) {
+            LOG(WARNING, "Videotoolbox textures only support clamp to edge wrapping: "
+                "disabling direct rendering");
```
cboesch-gpsw, line 168: Another open question / not blocking: do we really want to warn in case of legit internal fallback (the warning kind of imply that users are expected to explicitly force `direct_rendering` to 0 when targeting rendering for a macOS)? Maybe it could be in info or verbose? The same question applies for the one above, so we can obviously defer that to a future change.

cboesch-gpsw (approved):

libnodegl/backends/gl/glcontext_eagl.m
mbouron-gpsw: Sure, will do in a separate PR.

libnodegl/backends/gl/hwmap_videotoolbox_darwin_gl.c
mbouron-gpsw: No strong opinion, INFO seems like a good idea for all these kinds of messages (and it applies to all hwmap implementations not just this one). These warning are only there to warn that enabling these features has performance implication.

PR #344, opened 2021-11-22
Api commands cleanup

mbouron-gpsw:


pynodegl-utils/pynodegl_utils/com.py
```diff
@@ -24,23 +24,13 @@
 import inspect
 import os.path as op
 import pkgutil
-import sys
 import traceback
+from pynodegl_utils.module import load_script
```
mbouron-gpsw, line 28: Could also be `pynodegl_utils.scripts` so it is aligned with `pynodegl_utils.scriptsmgr`.

libnodegl/api.c
```diff
@@ -410,6 +408,8 @@ static int cmd_make_current(struct ngl_ctx *s, void *arg)
 #define DONE_CURRENT &(int[]){0}
 static int configure_ios(struct ngl_ctx *s, struct ngl_config *config)
 {
+    cmd_reset(s, &(int[]){KEEP_SCENE});
+
```
cboesch-gpsw: in the commit message, "call cmd_reset()", not "call cmd_stop()"

libnodegl/api.c
```diff
@@ -408,7 +408,7 @@ static int cmd_make_current(struct ngl_ctx *s, void *arg)
 #define DONE_CURRENT &(int[]){0}
 static int configure_ios(struct ngl_ctx *s, struct ngl_config *config)
 {
-    cmd_reset(s, &(int[]){KEEP_SCENE});
+    dispatch_cmd(s, cmd_reset, &(int[]){KEEP_SCENE});
```
cboesch-gpsw: In the commit description:

"responsible to free up" → "responsible for freeing up"

libnodegl/api.c
```diff
@@ -90,6 +90,11 @@ static int get_default_platform(void)
 #define KEEP_SCENE  0
 #define UNREF_SCENE 1
 
+static int cmd_stop(struct ngl_ctx *s, void *arg)
+{
+    return 0;
```
cboesch-gpsw, line 95: This function seems to remain a no-op after all your commits. What is its end purpose and how does it differ from `cmd_reset()`?

libnodegl/api.c
```diff
@@ -170,29 +172,34 @@ static int cmd_configure(struct ngl_ctx *s, void *arg)
...
+            keep_scene = UNREF_SCENE;
+            goto fail;
         }
     }
 
     if (config->hud) {
         s->hud = ngli_hud_create(s);
         if (!s->hud)
-            return NGL_ERROR_MEMORY;
+            goto fail;
```
cboesch-gpsw, line 191: missing a `ret` assignment

libnodegl/api.c
```diff
@@ -170,29 +172,34 @@ static int cmd_configure(struct ngl_ctx *s, void *arg)
...
+        goto fail;
+    }
 
     if (s->scene) {
         ret = ngli_node_attach_ctx(s->scene, s);
         if (ret < 0) {
-            ngli_node_detach_ctx(s->scene, s);
-            ngl_node_unrefp(&s->scene);
-            return ret;
+            keep_scene = UNREF_SCENE;
```
cboesch-gpsw, line 183: the error code paths below don't need this?

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -24,7 +24,7 @@
 import os.path as op
 import sys
 
-from pynodegl_utils.com import load_script
+from pynodegl_utils.module import load_script
```
cboesch-gpsw, line 27: I feel like this is not the first time we are removing the PyQt/PySide dependency. Could we add a simple test like this in our test suite?

```python
>>> any(k.startswith('Qt') for k in globals().keys())
False
>>> from PySide2 import QtCore
>>> any(k.startswith('Qt') for k in globals().keys())
True
```

cboesch-gpsw (changes requested):

libnodegl/api.c
mbouron-gpsw: cmd_stop() is only responsible to stop the rendering thread while cmd_reset() is in charge of resetting the current context. We want to explicitly call cmd_reset() outside of cmd_configure() so we can control the thread from which the function is executed.

libnodegl/api.c
mbouron-gpsw: Fixed.

libnodegl/api.c
mbouron-gpsw: Fixed.

libnodegl/api.c
mbouron-gpsw: Fixed.

libnodegl/api.c
mbouron-gpsw: Nope, we only detach the scene if we are not able to attach it to the new context (meaning the graph is not compatible with the new context).

pynodegl-utils/pynodegl_utils/tests/__init__.py
mbouron-gpsw: Added a check in `tests/__init__.py:run()`.

libnodegl/api.c
```diff
@@ -121,6 +121,8 @@ static int cmd_reset(struct ngl_ctx *s, void *arg)
 
 static int cmd_configure(struct ngl_ctx *s, void *arg)
 {
+    int ret = 0;
+    int keep_scene = KEEP_SCENE;
```
cboesch-gpsw, line 125: not a fan of this name: here we have `keep_scene = 0`, even though we are actually keeping the scene. I'd be more comfortable with `reset_arg = {KEEP_SCENE,UNREF_SCENE}` or a name in that spirit.

pynodegl-utils/pynodegl_utils/tests/__init__.py
```diff
@@ -125,6 +125,9 @@ def run():
     module = load_script(script_path)
     func = getattr(module, func_name)
 
+    # Ensure PySide/Qt is not imported
+    assert not any(k.startswith('PySide') or k.startswith('Qt') for k in globals().keys())
```
cboesch-gpsw, line 129: `k.startswith(('PySide', 'Qt'))`

cboesch-gpsw (changes requested):

pynodegl-utils/pynodegl_utils/tests/__init__.py
mbouron-gpsw: Done.

libnodegl/api.c
```diff
@@ -121,6 +121,8 @@ static int cmd_reset(struct ngl_ctx *s, void *arg)
 
 static int cmd_configure(struct ngl_ctx *s, void *arg)
 {
+    int ret = 0;
```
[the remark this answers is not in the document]
mbouron-gpsw: It is not needed but I wanted the ret variable to be initialized close to the keep_scene variable.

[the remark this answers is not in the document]
mbouron-gpsw: Anyway, I removed the change.

libnodegl/api.c
mbouron-gpsw: Renamed to `reset_param`.

cboesch-gpsw (approved):

PR #345, opened 2021-11-22
Improved dot

cboesch-gpsw:
## Before

dot-before
dot-before-2

## After

dot-after
dot-after-2


libnodegl/gen_doc.c
```diff
@@ -204,8 +204,8 @@ static int check_node_params(const struct node_class *cls)
         }
 
         if (par->flags & NGLI_PARAM_FLAG_DOT_DISPLAY_PACKED) {
-            if (par->type != NGLI_PARAM_TYPE_NODELIST) {
-                fprintf(stderr, "parameter %s.%s is not a node list, "
+            if (par->type != NGLI_PARAM_TYPE_NODELIST && par->type != NGLI_PARAM_TYPE_NODEDICT) {
+                fprintf(stderr, "parameter %s.%s is not a node list nor a node dict, "
```
mbouron-gpsw, line 208: Shouldn't this be part of a dedicated chunk ?

cboesch-gpsw: We could technically commit this chunk first, but the code in the dot can not be triggered unless this is enabled; it felt like those changes go together. What do you think?

libnodegl/dot.c
```diff
@@ -162,22 +162,36 @@ static void print_all_decls(struct bstr *b, const struct ngl_node *node, struct
...
 {
-    ngli_bstr_printf(b, "    %s_%p[label=<<b>%s</b> (x%d)", key, children, key, nb_children);
+    ngli_bstr_printf(b, "    %s_%p", key, children);
+    table_header(b, key, is_active, 2);
     for (int i = 0; i < nb_children; i++) {
         const struct ngl_node *node = children[i];
         char *info_str = node->cls->info_str ? node->cls->info_str(node) : NULL;
-        ngli_bstr_printf(b, LB "- %s", info_str ? info_str : "?");
+        ngli_bstr_printf(b, "<tr><td>#%d</td><td align=\"left\">%s</td></tr>",
+                         i, info_str ? info_str : "?");
```
mbouron-gpsw, line 191: Not related to this commit, but shouldn't we return an error if `node->info_str()` returns `NULL` ?

libnodegl/gen_doc.c
mbouron-gpsw: I prefer a separate commit but up to you.

mbouron-gpsw (approved):
LGTM, graph looks way nicer :)

libnodegl/dot.c
cboesch-gpsw: Yeah probably. It's a bit tricky because it can mean a memory error or a bug (not implemented).

libnodegl/gen_doc.c
cboesch-gpsw: Split out the chunk in "gen_doc: allow node dict to be displayed as packed"

PR #347, opened 2021-11-23
api: copy ngl_config.hud_export_filename internally

mbouron-gpsw:


tests/api.py
```diff
@@ -160,6 +162,19 @@ def api_hud(width=234, height=123):
...
+    with tempfile.TemporaryDirectory(prefix="ngl_") as td:
+        fname = op.join(td, 'hud_stats.csv')
+        ctx = ngl.Context()
+        assert ctx.configure(offscreen=1, width=width, height=height, backend=_backend, hud=1, hud_export_filename=fname) == 0
+        del fname
+        scene = _get_scene()
+        assert ctx.set_scene(scene) == 0
+        for i in range(60 * 3):
+            assert ctx.draw(i / 60.) == 0
+        del ctx
```
cboesch-gpsw, line 175: We already have `cmp_resources.py` which provides a HUD+CSV backed `test_resources` decorator, which we use in `tests/media.py`. Do we really need this test?

libnodegl/api.c
```diff
@@ -95,6 +95,12 @@ static int cmd_stop(struct ngl_ctx *s, void *arg)
     return 0;
 }
 
+static void config_reset(struct ngl_config *config)
+{
+    ngli_freep(&config->hud_export_filename);
+    memset(config, 0, sizeof(*config));
```
cboesch-gpsw, line 101: I would add the `memset` in a dedicated commit, it resets much more data than before.

cboesch-gpsw (changes requested):

tests/api.py
mbouron-gpsw: I didn't remember this test. Commit removed.

libnodegl/api.c
mbouron-gpsw: Done.

cboesch-gpsw (approved):

PR #348, opened 2021-11-24
gpu_ctx_gl: reset the color write mask before calling glClear()

mbouron-gpsw:


libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -821,7 +821,14 @@ static void gl_begin_render_pass(struct gpu_ctx *s, struct rendertarget *rt)
...
 
-    const int scissor_test = s_priv->glstate.scissor_test;
+    const struct glstate *glstate = &s_priv->glstate;
+    const GLboolean *color_write_mask = glstate->color_write_mask;
+    static const GLboolean default_color_write_mask[4] = {GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE};
+    const int reset_color_write_mask = memcmp(color_write_mask, default_color_write_mask, sizeof(default_color_write_mask));
+    if (reset_color_write_mask)
+        ngli_glColorMask(gl, GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE);
+
+    const int scissor_test = glstate->scissor_test;
```
mbouron-gpsw, line 831: NOTE: the reset of the OpenGL state will probably be moved to ngli_rendertarget_gl_clear() in a later PR and i'm still thinking about renaming ngli_rendertarget_gl_clear() to something else (maybe something around _load_attachments or _begin_pass)

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -821,7 +821,14 @@ static void gl_begin_render_pass(struct gpu_ctx *s, struct rendertarget *rt)
     struct rendertarget_gl *rt_gl = (struct rendertarget_gl *)rt;
     ngli_glBindFramebuffer(gl, GL_FRAMEBUFFER, rt_gl->id);
 
-    const int scissor_test = s_priv->glstate.scissor_test;
+    const struct glstate *glstate = &s_priv->glstate;
+    const GLboolean *color_write_mask = glstate->color_write_mask;
+    static const GLboolean default_color_write_mask[4] = {GL_TRUE, GL_TRUE, GL_TRUE, GL_TRUE};
+    const int reset_color_write_mask = memcmp(color_write_mask, default_color_write_mask, sizeof(default_color_write_mask));
```
cboesch-gpsw, line 827: could we use the `graphicstate` struct instead of the `glstate` struct to simplify this check? (In `graphicstate`, the mask is 32 bits instead of 128)

cboesch-gpsw (approved):
LGTM with or without my comment honored.

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: We can't because we don't store the current graphicstate in gpu_ctx nor gpu_ctx_gl.

PR #349, opened 2021-11-30
nodes: release nodes in the right order (top-bottom)

mbouron-gpsw:
While prefetching the nodes must be done bottom-up (resources must be
initialized first), releasing the nodes must be done top-bottom (we
don't want to release the resources first).


libnodegl/api.c
```diff
@@ -305,15 +305,21 @@ static int cmd_prepare_draw(struct ngl_ctx *s, void *arg)
 
     const int64_t start_time = s->hud ? ngli_gettime_relative() : 0;
 
-    ngli_darray_clear(&s->activitycheck_nodes);
+    ngli_darray_clear(&s->prefetch_activitycheck_nodes);
+    ngli_darray_clear(&s->release_activitycheck_nodes);
```
mbouron-gpsw, line 309: Theses names are far from ideal IMHO, could be `activitycheck_nodes_bu` and `activitycheck_nodes_tb` ? bu -> bottom-up, tb = top-bottom + adding the relevant comments in `internal.h`.

libnodegl/nodes.c
```diff
@@ -554,17 +554,22 @@ static int node_prefetch(struct ngl_node *node)
 int ngli_node_honor_release_prefetch(struct darray *nodes_array)
 {
     struct ngl_node **nodes = ngli_darray_data(nodes_array);
+
     for (int i = 0; i < ngli_darray_count(nodes_array); i++) {
```
cboesch-gpsw, line 558: can you add a short comment stating the order of the nodes (leaf-first) in one or two of these places:

1. in the declaration of `activitycheck_nodes` in the context
2. during the push in `ngli_node_visit()`
3. here above each loop (respectively "release (starting from parent)", "prefetch (starting from children)")

It's not obvious that `activitycheck_nodes` is leaf-first so these loops can be confusing.

cboesch-gpsw (changes requested):
Functionally it LGTM, but could you add a simple RTT test with timeranges?

Nitpick: you should use `.` instead of `'` for the graph "angles" in your schematic

libnodegl/nodes.c
mbouron-gpsw: Done.

cboesch-gpsw (approved):

PR #352, opened 2021-12-01
Eval expr

cboesch-gpsw:


libnodegl/node_eval.c
```diff
@@ -0,0 +1,240 @@
...
+                                          NGL_NODE_NOISEVEC4,       \
+                                          NGL_NODE_EVALFLOAT,       \
+                                          NGL_NODE_EVALVEC2,        \
+                                          NGL_NODE_EVALVEC3,        \
+                                          NGL_NODE_EVALVEC4,        \
+                                          NGL_NODE_UNIFORMBOOL,     \
+                                          NGL_NODE_UNIFORMFLOAT,    \
+                                          NGL_NODE_UNIFORMVEC2,     \
+                                          NGL_NODE_UNIFORMVEC3,     \
+                                          NGL_NODE_UNIFORMVEC4,     \
```
mbouron-gpsw, line 53: Maybe I'm missed something but are vector types really supported ? I tried to use them but I couldn't find a way to access their components (either by doing something like `v.x` or `|v[0]`. Only the first element seems accessible by using `v`.

Depending on whether we support vectors or not, please add a description on how to access their components or remove them from the allowed nodes.

Moreover, it would be nice to be able to use functions and operators directly on them (like we do in glsl) but I'm not sure it is compatible with the current design which uses a separate expression per output component. On the function side of things, providing some function like the quaternion slerp would be useful.

libnodegl/eval.c
```diff
@@ -0,0 +1,677 @@
...
+}
+
+#define PUSH(dst, token) do {               \
+    if (!ngli_darray_push(dst, token))      \
+        return NGL_ERROR_MEMORY;            \
+} while (0)
+
+/*
+ * Prototypes for the recursive descent parsing:
+ * - parse_subexpr: a sub-expression is basically anything the composes a value
```
mbouron-gpsw, line 259: typo: the -> that ?

libnodegl/eval.c
```diff
@@ -0,0 +1,677 @@
...
+        if (token->chr == '(') {
+            PUSH(operators, token);
+            continue;
+        }
+
+        if (token->chr == ')') {
+            for (;;) {
+                const struct token *o = ngli_darray_pop(operators);
+                if (!o) {
+                    LOG(ERROR, "expected opening '(' not found for closing ')' at position %d", token->pos);
```
mbouron-gpsw, line 559: Isn't this error message dedicated to the case where `o =! NULL && o->chr != '('` ?

libnodegl/eval.c
```diff
@@ -0,0 +1,677 @@
...
+            .chr        = *p,
+            .precedence = get_binary_operator_precedence(*p),
+            .func.f     = get_binary_operator_func(*p),
+            .nb_args    = 2,
+        };
+        PUSH(&s->tokens, &token);
+        return parse_subexpr(s, expr, p + 1);
+    }
+
+    /* Parse special end characters ')' and ',' */
```
mbouron-gpsw, line 396: Can you mention how `,` works in the documentation and add an example ?

libnodegl/node_eval.c
cboesch-gpsw: Yep, it's not supported, I forgot to remove those. Fixed in this new iteration, including an assert that the underlying data type is a float.

For vectorial data, we need to figure out a sane way of doing the swizzling, so that may requires some deeper thinking.

libnodegl/eval.c
cboesch-gpsw: Yep, fixed, thanks

libnodegl/eval.c
```diff
@@ -0,0 +1,677 @@
...
+}
+
+/*
+ * This function has multiple purposes:
+ *
+ * - check if the expression is valid by simulating a simplified evaluation of
+ *   the expression with extra checks (which would have been redundant if
+ *   called for every eval_run call).
+ * - make sure the temporary stack is pre-allocated with enough space so that
+ *   eval_run calls do not trigger any heap re-alloc
```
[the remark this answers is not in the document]
cboesch-gpsw: At the end yes, but note for intermediate states.

Example:
- input: `(1+2) * 3`
- rpn: `3 2 1 + *`

The stack will be:

| step | stack |
|---|---|
| 0 | 3 |
| 1 | 3 2 |
| 2 | 3 2 1 |
| 3 | 3 3 |
| 4 | 9 |

libnodegl/eval.c
```diff
@@ -0,0 +1,677 @@
...
+    ngli_darray_clear(stack);
+
+    const struct token *tokens = ngli_darray_data(&s->output);
+    for (int i = 0; i < ngli_darray_count(&s->output); i++) {
+        const struct token *token = &tokens[i];
+
+        if (token->type == TOKEN_CONSTANT || token->type == TOKEN_VARIABLE) {
+            PUSH(stack, token);
+        } else {
+            int ret = check_eval_func(stack, token);
```
mbouron-gpsw, line 505: nit: Would it be possible to move the logic that consumes the stack here, so all the stack logic is in the same scope ? (same applies to ngli_eval_run())

libnodegl/eval.c
```diff
@@ -0,0 +1,677 @@
...
+        } else if (token->type == TOKEN_CONSTANT) {
+            PUSH(stack, token);
+        } else {
+            const struct token r = {.type=TOKEN_CONSTANT, .value=eval_func(stack, token)};
+            PUSH(stack, &r);
+        }
+    }
+
+    const struct token *res = ngli_darray_pop(stack);
+    *dst = res ? res->value : 0.0;
```
mbouron-gpsw, line 662: Can res be NULL ? Shouldn't it be considered an error ?

cboesch-gpsw: It's basically to handle the case of an empty expression (no tokens).

libnodegl/eval.c
cboesch-gpsw: Inlining the function or part of it makes the code respectively clumsy (because it creates many nested if) or inefficient (because we would have to branch twice on the number of arguments).

I renamed the functions and added comments:
```diff
diff --git a/libnodegl/eval.c b/libnodegl/eval.c
index 84f7ef05..9b330f61 100644
--- a/libnodegl/eval.c
+++ b/libnodegl/eval.c
@@ -457,7 +457,11 @@ static int missing_argument(const struct token *token, int got)
     return NGL_ERROR_INVALID_DATA;
 }

-static int check_eval_func(struct darray *stack, const struct token *token)
+/*
+ * Check that the operator has the expected number of arguments by consuming
+ * the stack.
+ */
+static int check_operator_pop(struct darray *stack, const struct token *token)
 {
     const struct token *o1 = ngli_darray_pop(stack);
     if (!o1)
@@ -502,7 +506,7 @@ static int prepare_eval_run(struct eval *s)
         if (token->type == TOKEN_CONSTANT || token->type == TOKEN_VARIABLE) {
             PUSH(stack, token);
         } else {
-            int ret = check_eval_func(stack, token);
+            int ret = check_operator_pop(stack, token);
             if (ret < 0)
                 return ret;
             const struct token r = {.type=TOKEN_CONSTANT}; /* fake result */
@@ -621,7 +625,10 @@ int ngli_eval_init(struct eval *s, const char *expr, const struct hmap *vars)
     return 0;
 }

-static float eval_func(struct darray *stack, const struct token *token)
+/*
+ * Evaluate the operator by consuming the stack.
+ */
+static float eval_operator_pop(struct darray *stack, const struct token *token)
 {
     const struct token *o1 = ngli_darray_pop(stack);
     if (token->nb_args == 1)
@@ -653,7 +660,7 @@ int ngli_eval_run(struct eval *s, float *dst)
         } else if (token->type == TOKEN_CONSTANT) {
             PUSH(stack, token);
         } else {
-            const struct token r = {.type=TOKEN_CONSTANT, .value=eval_func(stack, token)};
+            const struct token r = {.type=TOKEN_CONSTANT, .value=eval_operator_pop(stack, token)};
             PUSH(stack, &r);
         }
     }
```

libnodegl/eval.c
mbouron-gpsw: Can you add such case (`""`)to the tests ? Should we also handle NULL or error out instead of crashing ?

libnodegl/eval.c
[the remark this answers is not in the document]
mbouron-gpsw: Thanks for the explanation, what confused me if that I did not notice that check_eval_func and eval_func consume the stack.

libnodegl/eval.c
mbouron-gpsw: Maybe an assert in the init function ?

cboesch-gpsw: Added the following:
```c
    if (!expr)
        return NGL_ERROR_INVALID_DATA;
```

mbouron-gpsw: Thanks.

libnodegl/eval.c
```diff
@@ -0,0 +1,687 @@
...
+
+static void *get_binary_operator_func(char c)
+{
+    switch (c) {
+    case '+': return f_add;
+    case '-': return f_sub;
+    case '*': return f_mul;
+    case '/': return f_div;
+    }
+    return NULL;
```
mbouron-gpsw, line 60: ngli_asser(0) ?

mbouron-gpsw (approved):
LGTM, thanks.

libnodegl/eval.c
cboesch-gpsw: done

PR #353, opened 2021-12-13
GL cleanups preparing the support of OpenGL wrapped contexts

mbouron-gpsw:


libnodegl/backends/gl/pipeline_gl.c
```diff
@@ -714,6 +714,7 @@ void ngli_pipeline_gl_draw(struct pipeline *s, int nb_vertices, int nb_instances
     ngli_glstate_update(gl, glstate, &graphics->state);
     int scissor[4];
     get_scissor(s, scissor);
+    ngli_glstate_update_viewport(gl, glstate, gpu_ctx_gl->viewport);
```
cboesch-gpsw, line 717: why don't we have a `get_viewport()` similarly to `get_scissor()` above?

libnodegl/backends/gl/glstate.c
```diff
@@ -102,44 +102,72 @@ static GLenum get_gl_cull_mode(int cull_mode)
     return gl_cull_mode_map[cull_mode];
 }
 
-void ngli_glstate_probe(const struct glcontext *gl, struct glstate *state)
+void ngli_glstate_reset(const struct glcontext *gl, struct glstate *glstate)
```
cboesch-gpsw, line 105: this commit needs a little description, explaining the "why"

libnodegl/backends/gl/glstate.h
```diff
@@ -56,8 +57,11 @@ struct glstate {
     GLenum cull_face_mode;
 
     GLboolean scissor_test;
+
+    /* Dynamic graphics state */
     int scissor[4];
 
+    /* Common state */
```
cboesch-gpsw: Aren't you deleting this comment in a later commit?

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -683,8 +683,6 @@ static int gl_end_draw(struct gpu_ctx *s, double t)
     struct glcontext *gl = s_priv->glcontext;
     struct ngl_config *config = &s->config;
 
-    ngli_gpu_ctx_end_render_pass(s);
```
cboesch-gpsw, line 689: did you mean "explicitly" instead of "implicitly" in the commit message? Can you add a sentence explaining why?

libnodegl/backends/gl/glstate.c
```diff
@@ -168,6 +168,10 @@ void ngli_glstate_reset(const struct glcontext *gl, struct glstate *glstate)
     /* Program */
     ngli_glUseProgram(gl, 0);
     glstate->program_id = 0;
+
+    /* VAO */
+    if (gl->features & NGLI_FEATURE_VERTEX_ARRAY_OBJECT)
+        ngli_glBindVertexArray(gl, 0);
```
cboesch-gpsw, line 174: Is `unbind_vertex_attribs()` still needed in the pipeline?

You are resetting the program and VAO; how about the FBO?

Also, should we have a current VAO id in the state (and honor it as needed in the pipeline instead of unconditionally like currently)? What's the direction overall here: are we going to move out `program_id`, or are we going to have `{program,fbo,vao}_id` in glstate?

Lastly: you are resetting the VAO state, what about the `glVertexAttribDivisor` and other related "state"?

cboesch-gpsw (changes requested):

libnodegl/backends/gl/glcontext.c
```diff
@@ -117,6 +123,21 @@ static int glcontext_probe_version(struct glcontext *glcontext)
...
+    const char *gl_version = (const char *)ngli_glGetString(glcontext, GL_VERSION);
+    if (!gl_version) {
+        LOG(ERROR, "could not get OpenGL version");
+        return NGL_ERROR_BUG;
+    }
+
+    const char *es_prefix = "OpenGL ES";
+    const int is_es = !strncmp(es_prefix, gl_version, strlen(es_prefix));
+    const int backend = is_es ? NGL_BACKEND_OPENGLES : NGL_BACKEND_OPENGL;
+    if (glcontext->backend != backend) {
```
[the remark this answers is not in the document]
mbouron-gpsw: We cannot have auto currently however I plan to support it in the future.

libnodegl/backends/gl/glcontext.c
```diff
@@ -92,6 +92,12 @@ static const int platform_to_glplatform[] = {
     [NGL_PLATFORM_WAYLAND] = GLPLATFORM_EGL,
 };
 
+static const char * backend_names[] = {
```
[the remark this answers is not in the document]
mbouron-gpsw: Done.

libnodegl/backends/gl/glstate.c
```diff
@@ -342,3 +342,11 @@ void ngli_glstate_update_scissor(const struct glcontext *gl, struct glstate *gls
     memcpy(glstate->scissor, scissor, sizeof(glstate->scissor));
     ngli_glScissor(gl, scissor[0], scissor[1], scissor[2], scissor[3]);
 }
+
+void ngli_glstate_update_viewport(const struct glcontext *gl, struct glstate *glstate, const int *viewport)
```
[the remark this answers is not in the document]
mbouron-gpsw: Reworded.

libnodegl/backends/gl/pipeline_gl.c
mbouron-gpsw: We don't need to invert the viewport.

libnodegl/backends/gl/glstate.h
mbouron-gpsw: Fixed. Deleting this comment was not intentional.

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: I meant implicitly. Added a sentence in the commit description.

libnodegl/backends/gl/glcontext.c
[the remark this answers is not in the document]
cboesch-gpsw: And you will need a string mapping in this situation?

libnodegl/backends/gl/glstate.c
mbouron-gpsw: > Is `unbind_vertex_attribs()` still needed in the pipeline?

It is still needed for contexts that do not support VAOs (GLES 2.0).

> 
> You are resetting the program and VAO; how about the FBO?

We don't need to reset the FBO state. node.gl binds the relevant FBO when required.

> 
> Also, should we have a current VAO id in the state (and honor it as needed in the pipeline instead of unconditionally like currently)? What's the direction overall here: are we going to move out `program_id`, or are we going to have `{program,fbo,vao}_id` in glstate?

We currently don't need to manage the VAO state on our side as we don't share VAOs between pipelines. The overall idea is to add states that are relevant depending on our use cases.

> 
> Lastly: you are resetting the VAO state, what about the `glVertexAttribDivisor` and other related "state"?

It should not be needed as the attrib divisor state is stored in the VAO state and for contexts not supporting VAOs (GLES 2.0) we reset the vertex attrib state after each pipeline draw.

libnodegl/backends/gl/pipeline_gl.c
mbouron-gpsw: It's not always inverted. I would rather keep get_scissor.

libnodegl/backends/gl/glcontext.c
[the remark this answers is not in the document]
mbouron-gpsw: Probably for error messages yes.

libnodegl/backends/gl/gpu_ctx_gl.c
cboesch-gpsw: OK. The use of "implicitly" is weird; I'm assuming you mean in the "unexpectedly" sense, because the call to end render pass is pretty explicit in gl end draw. Anyway, that's just nitpick.

mbouron-gpsw: I can reword, I meant "implicitly" from the point of view of the user manipulating the gpu_ctx API.

libnodegl/backends/gl/glstate.c
cboesch-gpsw: > > You are resetting the program and VAO; how about the FBO?
> 
> We don't need to reset the FBO state. node.gl binds the relevant FBO when required.
> 

But isn't it the same with the program id? The reason we reset it is to restore a clean state (for wrapped but also between re-configure maybe?).

mbouron-gpsw: For non-wrapped contexts there is no point in resetting the framebuffer to 0, if a reconfigure happens the context is entirely recreated.
For wrapped contexts, we want to let the wrapped framebuffer bound after ngl_draw() has exited so the user does not need to rebind it to draw to it (binding a framebuffer has potential performance implications depending on the platform).

cboesch-gpsw (approved):

PR #355, opened 2021-12-16
Pipeline consts

cboesch-gpsw:


mbouron-gpsw (approved):
Checked against the multibackend branch and it does not introduce any incompatibility so LGTM, thanks.

PR #356, opened 2021-12-17
Change default stencil read and write mask values

mbouron-gpsw:


cboesch-gpsw (changes requested):
you need to fix this as well: `const GLboolean stencil_write_mask = state->stencil_write_mask;`

mbouron-gpsw:
> * stencil_write_mask

Done in a separate commit. I also adjusted stencil_ref type.

cboesch-gpsw (approved):

PR #359, opened 2021-12-19
Render* and Filter* nodes

cboesch-gpsw:
~~Depends on https://github.com/gopro/gopro-lib-node.gl/pull/355 and https://github.com/gopro/gopro-lib-node.gl/pull/352~~


libnodegl/blending.c
```diff
@@ -0,0 +1,75 @@
...
+    [NGLI_BLENDING_DST_IN]   = {NGLI_BLEND_FACTOR_ZERO,                NGLI_BLEND_FACTOR_SRC_ALPHA},
+    [NGLI_BLENDING_SRC_ATOP] = {NGLI_BLEND_FACTOR_DST_ALPHA,           NGLI_BLEND_FACTOR_ONE_MINUS_SRC_ALPHA},
+    [NGLI_BLENDING_DST_ATOP] = {NGLI_BLEND_FACTOR_ONE_MINUS_DST_ALPHA, NGLI_BLEND_FACTOR_SRC_ALPHA},
+    [NGLI_BLENDING_XOR]      = {NGLI_BLEND_FACTOR_ONE_MINUS_DST_ALPHA, NGLI_BLEND_FACTOR_ONE_MINUS_SRC_ALPHA},
+};
+
+const struct param_choices ngli_blending_choices = {
+    .name = "blend_preset",
+    .consts = {
+        {"default",  NGLI_BLENDING_DEFAULT,  .desc=NGLI_DOCSTRING("unchanged current graphic state")},
```
mbouron-gpsw, line 47: graphics

libnodegl/node_render.c
```diff
@@ -149,6 +151,9 @@ static const struct node_param render_params[] = {
                  .desc=NGLI_DOCSTRING("per instance extra vertex attributes made accessible to the `program`")},
     {"nb_instances", NGLI_PARAM_TYPE_I32, OFFSET(nb_instances), {.i64 = 1},
                  .desc=NGLI_DOCSTRING("number of instances to draw")},
+    {"blending", NGLI_PARAM_TYPE_SELECT, OFFSET(blending),
+                 .choices=&ngli_blending_choices,
+                 .desc=NGLI_DOCSTRING("define how this node and the current frame buffer are blending together")},
```
mbouron-gpsw, line 156: are blended ?

libnodegl/glsl/helper_misc_utils.glsl
```diff
@@ -0,0 +1,5 @@
+#define sat(x) clamp(x, 0.0, 1.0)
+#define linear(a, b, x) (((x) - (a)) / ((b) - (a)))
+
+const vec3 luma_weights = vec3(.2126, .7152, .0722); // BT.709
+const float pi = 3.14159265358979323846;
```
mbouron-gpsw, line 5: Since these are internal helpers, shouldn't they prefixed by ngl_ ? For example I used the `luma_weights ` in some of my shaders, resulting in a clash with my variable name.

libnodegl/filterschain.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "internal.h"
+
+struct filterschain;
+
+/* We can not use the #include mechanism from glsl2c.py because a helper can
+ * appear in several filters */
+#define NGLI_FILTER_HELPER_MISC_UTILS  (1 << 0)
+#define NGLI_FILTER_HELPER_SRGB2LINEAR (1 << 1)
+#define NGLI_FILTER_HELPER_LINEAR2SRGB (1 << 2)
```
mbouron-gpsw, line 33: Why not defining directly both functions ?

libnodegl/filterschain.c
```diff
@@ -0,0 +1,146 @@
...
+    struct hmap *unique_filters;
+    struct bstr *str;
+    const char *source_name;
+    const char *source_code;
+    uint32_t helpers;
+};
+
+struct filterschain *ngli_filterschain_create(void)
+{
+    struct filterschain *s = ngli_calloc(1, sizeof(*s));
```
mbouron-gpsw, line 45: `return ngli_calloc(1, sizeof(*s));` ?

ubitux: You mean `return ngli_calloc(1, sizeof(struct filterschain))`? I don't think we use this pattern elsewhere.

libnodegl/node_renderother.c
```diff
@@ -0,0 +1,814 @@
...
+    struct pipeline_desc *descs = ngli_darray_data(&s->pipeline_descs);
+    struct pipeline_desc *desc = &descs[rnode->id];
+
+    struct graphicstate state = rnode->graphicstate;
+    int ret = ngli_blending_apply_preset(&state, s->blending);
+    if (ret < 0)
+        return ret;
+
+    struct pipeline_params pipeline_params = {
+        .type     = NGLI_PIPELINE_TYPE_GRAPHICS,
```
mbouron-gpsw, line 511: extra spaces

libnodegl/node_renderother.c
```diff
@@ -0,0 +1,814 @@
...
+    struct pipeline_desc *desc = &descs[ctx->rnode_pos->id];
+
+    const float *modelview_matrix  = ngli_darray_tail(&ctx->modelview_matrix_stack);
+    const float *projection_matrix = ngli_darray_tail(&ctx->projection_matrix_stack);
+
+    if (!ctx->render_pass_started) {
+        struct gpu_ctx *gpu_ctx = ctx->gpu_ctx;
+        ngli_gpu_ctx_begin_render_pass(gpu_ctx, ctx->current_rendertarget);
+        ctx->render_pass_started = 1;
+    }
```
mbouron-gpsw, line 741: Can you move this chunk next to the `s->draw(s, desc->pipeline);` or directly into the draw function ?

libnodegl/pass.c
```diff
@@ -687,6 +687,70 @@ int ngli_pass_update(struct pass *s, double t)
     return 0;
 }
 
+static void update_texture_uniforms(struct pipeline *pipeline, const struct pgcraft_texture_info *info)
```
mbouron-gpsw: It's not only about texture uniforms update, it also updates the texture. I would rather name it `update_texture` or `update_texture_resources`.

libnodegl/node_renderother.c
```diff
@@ -0,0 +1,814 @@
...
+            (ret = ngli_buffer_init(s->uvcoords, sizeof(default_uvcoords), VERTEX_USAGE_FLAGS)) < 0 ||
+            (ret = ngli_buffer_upload(s->vertices, default_vertices, sizeof(default_vertices), 0)) < 0 ||
+            (ret = ngli_buffer_upload(s->uvcoords, default_uvcoords, sizeof(default_uvcoords), 0)) < 0)
+            return ret;
+
+        s->position_attr.stride = 3 * 4;
+        s->position_attr.buffer = s->vertices;
+        s->uvcoord_attr.stride = 2 * 4;
+        s->uvcoord_attr.buffer = s->uvcoords;
+        s->nb_vertices = 4;
```
mbouron-gpsw, line 372: Can you add the same spacing that you did for the else block bellow ?

libnodegl/node_renderother.c
```diff
@@ -0,0 +1,814 @@
...
+        struct buffer_layout vertices_layout = geom_node->vertices_layout;
+        struct buffer_layout uvcoords_layout = geom_node->uvcoords_layout;
+
+        if (!uvcoords) {
+            LOG(ERROR, "the specified geometry is missing UV coordinates");
+            return NGL_ERROR_INVALID_USAGE;
+        }
+
+        if (vertices_layout.type != NGLI_TYPE_VEC3 ||
+            vertices_layout.format != NGLI_FORMAT_R32G32B32_SFLOAT) {
```
mbouron-gpsw, line 388: This is not supposed to happen right ? Unless something change the type of vertices in one of the allowed nodes ? (the geometry node also only allow vec3 as input for vertices)

libnodegl/blending.c
cboesch-gpsw: changed, but the structure is called `graphic(_)state` though :D

libnodegl/node_render.c
cboesch-gpsw: changed

libnodegl/glsl/helper_misc_utils.glsl
cboesch-gpsw: done

libnodegl/filterschain.h
cboesch-gpsw: Because initially I planed to have 2 separated filters for them. I actually implemented them, dropped them, and now just re-added them, this time in a dedicated commit: `nodes: add FilterLinear2sRGB and FilterSRGB2Linear` (see the dedicated test to understand the purpose of these filters). These 2 nodes need one of the helper only, so I think it makes sense to separated them.

Of course we could have all the helpers combined, but I'm afraid that with time the boilerplate will get too large, so I think it makes sense to start splitting them a bit and include them only when necessary.

Would you prefer a "gamma helper" or something?

libnodegl/node_renderother.c
cboesch-gpsw: It's because it's aligned with the following field. Removed anyway.

libnodegl/node_renderother.c
cboesch-gpsw: Moved close to the draw (don't want to have that chunk duplicated in each callback).

libnodegl/pass.c
cboesch-gpsw: Done in `pass: move texture uniforms update in a dedicated pipeline utils module` with: `pipeline_utils.[ch]:ngli_pipeline_utils_update_texture()`

libnodegl/pass.h
```diff
@@ -85,6 +85,7 @@ int ngli_pass_init(struct pass *s, struct ngl_ctx *ctx, const struct pass_params
 int ngli_pass_prepare(struct pass *s);
 void ngli_pass_uninit(struct pass *s);
 int ngli_pass_update(struct pass *s, double t);
+void ngli_pass_update_texture_uniforms(struct pipeline *pipeline, const struct pgcraft_texture_info *info);
```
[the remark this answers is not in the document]
cboesch-gpsw: Yup, see previous comment.

libnodegl/node_renderother.c
cboesch-gpsw: Done

libnodegl/node_renderother.c
cboesch-gpsw: That's right, but I'm afraid of the time when we will add more flexibility in the `Geometry` node (or add another dedicated geometry node) and forget about this: it may not be detected unless someone uses this particular combination.

I could return an `NGL_ERROR_BUG` but if the mentioned scenario happens, it will be not be a bug but indeed something unsupported... I can't even add a comment saying that it "can't happen" because again if the mentioned scenario happens the comment won't be true anymore (as we will easily forget to update it as well).

libnodegl/blending.c
mbouron-gpsw: This will be addressed soon :)

libnodegl/filterschain.c
mbouron-gpsw: Nevermind then.

libnodegl/filterschain.h
mbouron-gpsw: The current name is fine.

mbouron-gpsw:
LGTM, thanks.

mbouron-gpsw (approved):

PR #362, opened 2021-12-27
gpu_ctx_gl: reset depth/stencil write masks before calling glClear()

mbouron-gpsw:
The OpenGL documentation states:

> The pixel ownership test, the scissor test, dithering, and the buffer
writemasks affect the operation of glClear.

Without this patch, rendertargets can end up partially cleared if the
depth/stencil write masks are not reset to their default values.

Related to bdb5f389560d3e81a5a1019a6c0b74013f0fbc09.


tests/depth_stencil.py
```diff
@@ -0,0 +1,111 @@
+#!/usr/bin/env python
```
ubitux, line 1: unneeded shebang

tests/depth_stencil.py
```diff
@@ -0,0 +1,111 @@
...
+#
+# Copyright 2020 GoPro Inc.
```
ubitux, line 3: 2021 (or 2022)

tests/depth_stencil.py
```diff
@@ -0,0 +1,111 @@
...
+from pynodegl_utils.toolbox.colors import COLORS
+
+
+def _render_quad(cfg, corner=(-1, -1, 0), width=(2, 0, 0), height=(0, 2, 0), color=(1, 1, 1, 1)):
+    quad = ngl.Quad(corner, width, height)
+    program = ngl.Program(vertex=cfg.get_vert('color'), fragment=cfg.get_frag('color'))
+    program.update_vert_out_vars(var_tex0_coord=ngl.IOVec2(), var_uvcoord=ngl.IOVec2())
+    render = ngl.Render(quad, program)
+    render.update_frag_resources(color=ngl.UniformVec4(value=color))
+    return render
```
ubitux, line 35: note to self: this needs to be adjusted in my branches if upstreamed before my other PRs

tests/meson.build
```diff
@@ -174,6 +174,11 @@ foreach backend : backends
     endif
   endif
 
+  tests_depth_stencil = [
+    'depth',
+    'stencil',
+  ]
```
ubitux, line 180: Both tests are passing for me with your fix reverted. They're not supposed to be related?

ubitux (changes requested):

tests/depth_stencil.py
mbouron-gpsw: Removed.

tests/meson.build
mbouron-gpsw: They are related, I updated them to use 2 keyframes. The tests are now failing if the first commit is reverted.
The issue can also be reproduced with `ngl-python -u tests/depth_stencil.py depth_stencil_depth` (`-u` is mandatory to reproduce the issue).

tests/depth_stencil.py
mbouron-gpsw: Done.

tests/depth_stencil.py
```diff
@@ -0,0 +1,109 @@
+# Copyright 2022 GoPro Inc.
```
cboesch-gpsw, line 1: missing '#' at the beginning

cboesch-gpsw (changes requested):

cboesch-gpsw (approved):

PR #366, opened 2022-01-04
build: use the meson binary provided by the Python virtual environment for the tests

mbouron-gpsw:
Fixes `make tests` on systems where meson is not installed.


cboesch-gpsw (changes requested):
The change itself is fine, but as discussed the reason why it fixes `make tests` is not at all what we expect. I'm fine with the commit but with a different description stating:

1. the change is for consistency with other meson calls
2. fixes `make tests` on MacOS where there is a broken GNU/Make version packaged by a side effect of having a global variable specified before meson (`MAKEFLAGS=...` but it could be anything)

Bonus point for a deeper investigation, because if we remove `MAKEFLAGS` are are going to have a bad time.

Investigation or not, I think it would make sense to have `MESON = /path/to/venv/bin/meson` hardcoded, which would work even without that `MAKEFLAGS=` specified.

mbouron-gpsw:
> The change itself is fine, but as discussed the reason why it fixes `make tests` is not at all what we expect. I'm fine with the commit but with a different description stating:
> 
>     1. the change is for consistency with other meson calls
> 
>     2. fixes `make tests` on MacOS where there is a broken GNU/Make version packaged by a side effect of having a global variable specified before meson (`MAKEFLAGS=...` but it could be anything)
> 

The commit description has been reworded.

> 
> Bonus point for a deeper investigation, because if we remove `MAKEFLAGS` are are going to have a bad time.
> 
> Investigation or not, I think it would make sense to have `MESON = /path/to/venv/bin/meson` hardcoded, which would work even without that `MAKEFLAGS=` specified.

I agree but I keep this investigation for later.

cboesch-gpsw (approved):

PR #367, opened 2022-01-04
gpu_ctx_gl: misc cleanups

mbouron-gpsw:


libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -567,68 +543,19 @@ static int update_capture_cvpixelbuffer(struct gpu_ctx *s, CVPixelBufferRef capt
...
-    rt_params.colors[0].store_op = NGLI_STORE_OP_STORE;
-    rt_params.depth_stencil.load_op = NGLI_LOAD_OP_LOAD;
-    rt_params.depth_stencil.store_op = NGLI_STORE_OP_STORE;
-
-    ret = ngli_rendertarget_init(rt_load, &rt_params);
-    if (ret < 0)
+    if ((ret = create_rendertarget(s, color, resolve_color, depth_stencil,
+                                   NGLI_LOAD_OP_CLEAR, &rt)) < 0 ||
+        (ret = create_rendertarget(s, color, resolve_color, depth_stencil,
+                                   NGLI_LOAD_OP_LOAD, &rt_load)) < 0)
```
cboesch-gpsw, line 558: `create_rendertarget()` calls `ngli_rendertarget_freep()` in case of failure, but you do have them in this codepath as well (under the `fail:` label).

The code looks a bit fragile too, in particular around this:
```c
s_priv->default_rt = rt;
s_priv->default_rt_load = rt_load;
```

If it was possible to fail after these assignments (it can't right now fortunately), then `ngli_rendertarget_freep(&rt)` and `ngli_rendertarget_freep(&rt_load)` happening in the `fail` codepath would leave dangling pointers in `s_priv->default_rt` and `s_priv->default_rt_load` (because only the local pointers would have been reset to `NULL`. It's probably the case already, but I think it would be safer to reset `rt` and `rt_load` to `NULL` after they are transfered to the private context.

But now that I think about it, you may have a similar issue with the texture, where it could actually fail (for instance during the rendertarget creation) and leave a dangling free'ed texture in `s_priv->color`.

BTW, I'm assuming you're not working with `s_priv->default_rt` and `s_priv->default_rt_load` directly because you want this code to be somehow resilient in case of error? If not then maybe it's better to work directly with those.

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -161,6 +161,94 @@ static void gl_get_scissor(struct gpu_ctx *s, int *scissor)
...
+                               struct rendertarget **rendertargetp)
+{
+    int ret;
+    const struct ngl_config *config = &s->config;
+
+    struct rendertarget *rendertarget = ngli_rendertarget_create(s);
+    if (!rendertarget) {
+        ret = NGL_ERROR_MEMORY;
+        goto fail;
+    }
```
cboesch-gpsw, line 210: you could return directly `NGL_ERROR_MEMORY` here.

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -161,6 +161,94 @@ static void gl_get_scissor(struct gpu_ctx *s, int *scissor)
...
+            .clear_value[0] = config->clear_color[0],
+            .clear_value[1] = config->clear_color[1],
+            .clear_value[2] = config->clear_color[2],
+            .clear_value[3] = config->clear_color[3],
+            .store_op       = NGLI_STORE_OP_STORE,
+        },
+        .depth_stencil = {
+            .attachment = depth_stencil,
+            .load_op    = load_op,
+            .store_op   = NGLI_STORE_OP_STORE,
```
cboesch-gpsw, line 229: is it OK to set load/store op even if the depth stencil is `NULL`?

cboesch-gpsw (changes requested):

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: It is. With ngli_rendertarget_init() if the depth attachment is NULL, the load_op/store_op are ignored. With ngli_default_rendertarget_init() (wrapping the default framebuffer), the load_op/store_op are taken into account and works with NULL attachments.

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: Done.

mbouron-gpsw: I also removed the `goto` and `fail` label.

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: > `create_rendertarget()` calls `ngli_rendertarget_freep()` in case of failure, but you do have them in this codepath as well (under the `fail:` label).

Is that an issue ? the code can fail at any point and I need to cleanup all the allocated resources (cv_texture, texture, rt, rt_load).

> 
> The code looks a bit fragile too, in particular around this:
> 
> ```c
> s_priv->default_rt = rt;
> s_priv->default_rt_load = rt_load;
> ```
> 
> If it was possible to fail after these assignments (it can't right now fortunately), then `ngli_rendertarget_freep(&rt)` and `ngli_rendertarget_freep(&rt_load)` happening in the `fail` codepath would leave dangling pointers in `s_priv->default_rt` and `s_priv->default_rt_load` (because only the local pointers would have been reset to `NULL`. It's probably the case already, but I think it would be safer to reset `rt` and `rt_load` to `NULL` after they are transfered to the private context.
> 
> But now that I think about it, you may have a similar issue with the texture, where it could actually fail (for instance during the rendertarget creation) and leave a dangling free'ed texture in `s_priv->color`.
> 
> BTW, I'm assuming you're not working with `s_priv->default_rt` and `s_priv->default_rt_load` directly because you want this code to be somehow resilient in case of error? If not then maybe it's better to work directly with those.

This is the idea. I don't want to destroy the previous rendertargets before we are sure we can use the new ones. Destroying the previous resources first would require to mark the context as "failed" if an error happens when changing the rendertargets and making the gpu context acts accordingly. Maybe it is a saner behavior after all, I'm not sure being error resilient here adds much value. If the user can't render to the desired target maybe we should stop rendering (until a new target is provided) instead of rendering to the old target. What do you think ?

mbouron-gpsw: If that helps with the current code, I can remove the current `goto/label` flow, I just need to free-up all the resources in case of error while creating the two rendertargets.

mbouron-gpsw: Something like
``` patch
diff --git a/libnodegl/backends/gl/gpu_ctx_gl.c b/libnodegl/backends/gl/gpu_ctx_gl.c
index 7fb72a74..f7cee804 100644
--- a/libnodegl/backends/gl/gpu_ctx_gl.c
+++ b/libnodegl/backends/gl/gpu_ctx_gl.c
@@ -528,15 +528,14 @@ static int update_capture_cvpixelbuffer(struct gpu_ctx *s, CVPixelBufferRef capt
     struct rendertarget *rt = NULL;
     struct rendertarget *rt_load = NULL;

-    int ret = 0;
     if (capture_buffer) {
-        ret = wrap_capture_cvpixelbuffer(s, capture_buffer, &texture, &cv_texture);
+        int ret = wrap_capture_cvpixelbuffer(s, capture_buffer, &texture, &cv_texture);
         if (ret < 0)
-            goto fail;
+            return ret;
     } else {
         ret = create_texture(s, NGLI_FORMAT_R8G8B8A8_UNORM, 0, &texture);
         if (ret < 0)
-            goto fail;
+            return ret;
     }

     struct texture *color         = s_priv->ms_color ? s_priv->ms_color : texture;
@@ -546,8 +545,14 @@ static int update_capture_cvpixelbuffer(struct gpu_ctx *s, CVPixelBufferRef capt
     if ((ret = create_rendertarget(s, color, resolve_color, depth_stencil,
                                    NGLI_LOAD_OP_CLEAR, &rt)) < 0 ||
         (ret = create_rendertarget(s, color, resolve_color, depth_stencil,
-                                   NGLI_LOAD_OP_LOAD, &rt_load)) < 0)
-        goto fail;
+                                   NGLI_LOAD_OP_LOAD, &rt_load)) < 0) {
+        ngli_rendertarget_freep(&rt);
+        ngli_rendertarget_freep(&rt_load);
+        ngli_texture_freep(&texture);
+        if (cv_texture)
+            CFRelease(cv_texture);
+        return ret;
+    }

     ngli_rendertarget_freep(&s_priv->default_rt);
     ngli_rendertarget_freep(&s_priv->default_rt_load);
@@ -561,15 +566,6 @@ static int update_capture_cvpixelbuffer(struct gpu_ctx *s, CVPixelBufferRef capt
     s_priv->capture_cvtexture = cv_texture;

     return 0;
-
-fail:
-    ngli_rendertarget_freep(&rt);
-    ngli_rendertarget_freep(&rt_load);
-    ngli_texture_freep(&texture);
-    if (cv_texture)
-        CFRelease(cv_texture);
-
-    return ret;
 }
 #endif
```

cboesch-gpsw: You may be leaking `cv_texture` if create texture fails in this diff.

> > BTW, I'm assuming you're not working with `s_priv->default_rt` and `s_priv->default_rt_load` directly because you want this code to be somehow resilient in case of error? If not then maybe it's better to work directly with those.
> 
> This is the idea. I don't want to destroy the previous rendertargets before we are sure we can use the new ones. Destroying the previous resources first would require to mark the context as "failed" if an error happens when changing the rendertargets and making the gpu context acts accordingly. Maybe it is a saner behavior after all, I'm not sure being error resilient here adds much value. If the user can't render to the desired target maybe we should stop rendering (until a new target is provided) instead of rendering to the old target. What do you think ?

This logic seems specific to `update_capture_cvpixelbuffer()` (it's not done in `offscreen_rendertarget_init()` typically). I don't mind having resilience built-in but, does it actually work? What happens if the new user expectations are not actually backed by the proper data; aren't we risking overwriting or something? Isn't it going to generate garbage anyway? You're still returning an error, so right now you're making it resilient because you're expecting the user to continue the execution even in case of error.

To summarize, I'm not opposed to resilience, but it needs guarantee that it works, not because of faith in over-engineering. I understand that testing this is stuff is pretty complex, but are we able to trigger that even manually somehow? Maybe even a comment describing the exact scenario we're dealing with here would help.

mbouron-gpsw: Thanks for your input. I just had a quick look at `api.c` and it makes me wonder why I'm questioning myself about what would be the correct behavior.
``` c
static int cmd_set_capture_buffer(struct ngl_ctx *s, void *capture_buffer)
{
    struct ngl_config *config = &s->config;

    int ret = ngli_gpu_ctx_set_capture_buffer(s->gpu_ctx, capture_buffer);
    if (ret < 0) {
        if (s->scene) {
            ngli_node_detach_ctx(s->scene, s);
            ngl_node_unrefp(&s->scene);
        }
        config->capture_buffer = NULL;
        return ret;
    }

    config->capture_buffer = capture_buffer;

    return 0;
}
```

So, we should not try to be error resilient in `ngli_gpu_ctx_set_capture_buffer()`. We already detach the scene if an error happens and we should also destroy the gpu context and mark the ngl context has "*not* configured" preventing later calls to the rest of the API. Do you agree ?

cboesch-gpsw: Sounds reasonable. Do we have that kind of behavior (context reset on error) in other scenarios?

A bit unrelated but I don't see any logging happening in these scenarios. Maybe a top level log stating "thing got wrong because <error-string>, resetting context" would be welcome.

mbouron-gpsw: Added the relevant commits to simplify `update_capture_cvpixelbuffer()`.

libnodegl/api.c
```diff
@@ -234,6 +234,8 @@ static int cmd_set_capture_buffer(struct ngl_ctx *s, void *capture_buffer)
 {
     struct ngl_config *config = &s->config;
 
+    ngli_gpu_ctx_wait_idle(s->gpu_ctx);
```
cboesch-gpsw, line 237: This is gpu ctx call, so why isn't this done in the backend specific implementation of `ngli_gpu_ctx_set_capture_buffer()`? Having the user responsible for calling this sync requirement feels like exposing gpu ctx implementation specific quirks to the user.

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -161,6 +161,85 @@ static void gl_get_scissor(struct gpu_ctx *s, int *scissor)
...
+        .height = config->height,
+        .nb_colors = 1,
+        .colors[0] = {
+            .attachment     = color,
+            .resolve_target = resolve_color,
+            .load_op        = load_op,
+            .clear_value[0] = config->clear_color[0],
+            .clear_value[1] = config->clear_color[1],
+            .clear_value[2] = config->clear_color[2],
+            .clear_value[3] = config->clear_color[3],
```
cboesch-gpsw, line 215: not a blocking comment since it's already present, but I think you can do `.clear_value = {NGLI_ARG_VEC4(config->clear_color)},`

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -161,6 +161,85 @@ static void gl_get_scissor(struct gpu_ctx *s, int *scissor)
...
+        .readable = 1,
+    };
+
+    int ret;
+    if (color) {
+        ret = ngli_rendertarget_init(rendertarget, &params);
+    } else {
+        ngli_assert(!resolve_color);
+        ngli_assert(!depth_stencil);
+        ret = ngli_default_rendertarget_gl_init(rendertarget, &params);
```
cboesch-gpsw, line 232: Would it be even safer to have the asserts in `ngli_default_rendertarget_gl_init()` or can that function work with non-null attachments?

cboesch-gpsw (commented):
Despite my nitpicking I like this simplification :)

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: Done in https://github.com/gopro/gopro-lib-node.gl/pull/368, so I removed the above asserts.

libnodegl/api.c
mbouron-gpsw: Sounds good, i removed the commit entirely (no synchronization is needed on the GL side).

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -161,6 +161,83 @@ static void gl_get_scissor(struct gpu_ctx *s, int *scissor)
...
+        },
+        .readable = 1,
+    };
+
+    int ret;
+    if (color) {
+        ret = ngli_rendertarget_init(rendertarget, &params);
+    } else {
+        ret = ngli_default_rendertarget_gl_init(rendertarget, &params);
+    }
```
cboesch-gpsw, line 231: Nit:
```c
    int ret = color ? ngli_rendertarget_init(rendertarget, &params);
                    : ngli_default_rendertarget_gl_init(rendertarget, &params);
```
?

cboesch-gpsw (approved):

libnodegl/backends/gl/gpu_ctx_gl.c
mbouron-gpsw: I add code later on to the else block, so it's not possible to simplify it this way.

PR #369, opened 2022-01-11
Meson cleanups

mbouron-gpsw:


libnodegl/meson.build
```diff
@@ -90,7 +90,7 @@ trim_prefix = run_command([
   find_program('python'), '-c', 'import sys,os;print(os.path.relpath(*sys.argv[1:3]))',
   meson.current_source_dir(),
   meson.project_build_root(),
-]).stdout().strip()
+], check: true).stdout().strip()
```
cboesch-gpsw, line 93: This is needed in `tests/meson.build` as well

cboesch-gpsw (changes requested):

tests/meson.build
```diff
@@ -30,15 +30,15 @@ foreach backend : backends
...
+  has_block              = run_command(cap_cmd + ['block'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  has_compute            = run_command(cap_cmd + ['compute'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  has_instanced_draw     = run_command(cap_cmd + ['instanced_draw'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  has_shader_texture_lod = run_command(cap_cmd + ['shader_texture_lod'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  has_texture_3d         = run_command(cap_cmd + ['texture_3d'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  has_texture_cube       = run_command(cap_cmd + ['texture_cube'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  has_uint_uniforms      = run_command(cap_cmd + ['uint_uniforms'], check: false).stdout().split()[-1].to_int() == 1 ? true : false
+  max_color_attachments  = run_command(cap_cmd + ['max_color_attachments'], check: false).stdout().split()[-1].to_int()
+  max_samples            = run_command(cap_cmd + ['max_samples'], check: false).stdout().split()[-1].to_int()
```
cboesch-gpsw, line 41: Mmh, why false? Can they really fail?

mbouron-gpsw: My bad, I read the code too fast and was thinking we were checking error codes.

mbouron-gpsw: Fixed.

libnodegl/meson.build
```diff
@@ -89,8 +89,8 @@ endif
 trim_prefix = run_command([
   find_program('python'), '-c', 'import sys,os;print(os.path.relpath(*sys.argv[1:3]))',
   meson.current_source_dir(),
-  meson.build_root(),
-]).stdout().strip()
+  meson.project_build_root(),
```
eli-schwartz, line 92: If for some reason you want to avoid bumping your minimum version from 0.54 to 0.56, it is possible to do so by assigning current_build_dir() in the parent meson.build and then reusing that variable here.

tests/meson.build
eli-schwartz: False is the behavior preserving option...

The idea behind the meson change is that we kind of want to see people using/being safe with True, because if the unthinkable happens and the command does fail, the output may not actually be sane to use. Any EXIT_FAILURE case probably means things are broken enough you don't want to continue further.

libnodegl/meson.build
mbouron-gpsw: Thanks for the insight. We already depend on 0.56 in `tests/meson.build` so it is fine to bump here.

tests/meson.build
mbouron-gpsw: All these commands are not meant to fail. I adjusted the commit description to mention why we now set `check` to `true`. Thanks.

libnodegl/meson.build
cboesch-gpsw: Yeah, we pull meson from pip in our `configure.py` so even latest version could be considered fine. It would have been a problem otherwise, typically because Ubuntu 20.04 (latest LTS, Github CI) is packaging an old one (0.53.2 if I'm not mistaken). Even today, 0.54 is problematic anyway.

cboesch-gpsw (approved):

PR #376, opened 2022-01-30
pgcraft: fix buffers and textures bindings

mbouron-gpsw:


tests/data.py
```diff
@@ -483,3 +483,48 @@ def data_eval(cfg):
...
+
+@test_cuepoints(points={'c': (0, 0)}, nb_keyframes=1, tolerance=1)
+@scene()
+def data_vertex_and_fragment_blocks(cfg):
+    cfg.aspect_ratio = (1, 1)
+
+    src = ngl.Block(
+        fields=[
+            ngl.UniformVec3(value=COLORS.red, label='color'),
+            ngl.UniformFloat(value=0.5, label='opacity')
```
cboesch-gpsw, line 496: missing trailing comma

tests/data.py
```diff
@@ -483,3 +483,48 @@ def data_eval(cfg):
...
+        ],
+        layout='std140',
+    )
+    dst = ngl.Block(
+        fields=[
+            ngl.UniformVec3(value=COLORS.white, label='color')
```
cboesch-gpsw, line 502: ditto

tests/data.py
```diff
@@ -483,3 +483,48 @@ def data_eval(cfg):
     render.update_frag_resources(color=color)
 
     return render
+
+
+@test_cuepoints(points={'c': (0, 0)}, nb_keyframes=1, tolerance=1)
+@scene()
+def data_vertex_and_fragment_blocks(cfg):
```
cboesch-gpsw, line 490: can you move the commit description as a docstring into this function?

tests/data.py
```diff
@@ -483,3 +483,48 @@ def data_eval(cfg):
...
+            ngl.UniformVec3(value=COLORS.red, label='color'),
+            ngl.UniformFloat(value=0.5, label='opacity')
+        ],
+        layout='std140',
+    )
+    dst = ngl.Block(
+        fields=[
+            ngl.UniformVec3(value=COLORS.white, label='color')
+        ],
+        layout='std140',
```
cboesch-gpsw, line 504: Fine as is, but wouldn't it make sense to parametrize the layouts so that multiple combination are triggered? (ubo-ubo, ubo-ssbo, ssbo-ssbo). Not a blocking comment since we may end up dropping the layout anyway at some point, and since the specs seems pretty clear about it.

libnodegl/pgcraft.c
```diff
@@ -1150,9 +1150,14 @@ static void setup_glsl_info_gl(struct pgcraft *s)
...
-    /* Bindings are unique across stages and types */
-    for (int i = 0; i < NB_BINDINGS; i++)
-        s->next_bindings[i] = &s->bindings[i];
+
+    for (int i = 0; i < NGLI_PROGRAM_SHADER_NB; i++) {
+        /* Buffers (UBOs and SSBOs) bindings are shared across all stages */
+        s->next_bindings[BIND_ID(i, NGLI_BINDING_TYPE_UBO)] = &s->bindings[0];
+        s->next_bindings[BIND_ID(i, NGLI_BINDING_TYPE_SSBO)] = &s->bindings[0];
+        /* Texture bindings are shared across all stages */
+        s->next_bindings[BIND_ID(i, NGLI_BINDING_TYPE_TEXTURE)] = &s->bindings[1];
```
cboesch-gpsw, line 1159: Sounds like an important fix for the users. Can you mention that in the Changelog?

cboesch-gpsw (changes requested):
Only cosmetics, LGTM otherwise, thanks

tests/data.py
mbouron-gpsw: Fixed.

tests/data.py
mbouron-gpsw: Fixed.

tests/data.py
mbouron-gpsw: Done (I've kept the commit description though).

tests/data.py
mbouron-gpsw: It makes sense, the test could be declined into multiple tests triggering multiple scenarios. I've chosen explicitly std140 so this test can run on OpenGLES 3.0 hardware. But if that's ok with you, i'll do that later.

cboesch-gpsw: Sure, again not a blocker.

libnodegl/pgcraft.c
mbouron-gpsw: Done.

cboesch-gpsw (approved):

PR #377, opened 2022-01-31
rendertarget: remove ngli_rendertarget_read_pixels()

mbouron-gpsw:
Currently ngli_rendertarget_read_pixels() is only used in the CPU
capture callback of gpu_ctx_gl and can be inlined there. The new logic
is also simpler compared to the previous one since the code cannot be
called in the middle of a render pass anymore.

Moreover, ngli_rendertarget_read_pixels() is not great API-wise and a
similar functionality should be implemented as ngli_texture_download() or
something similar if one needs to download the data of a texture in
various places.


cboesch-gpsw (approved):
"ngli_rendertarget_read_pixels() is not great API-wise": you may want to elaborate here that (by basically mentioning the modern backend designs).

LGTM anyway

mbouron-gpsw:
> "ngli_rendertarget_read_pixels() is not great API-wise": you may want to elaborate here that (by basically mentioning the modern backend designs).
> 
> LGTM anyway

I adjusted the commit description, let me know if it's ok for you.

PR #384, opened 2022-02-03
api: always recreate the hud on scene change

mbouron-gpsw:
Before this commit, the hud was recreated in cmd_set_scene() only if the
new scene was non-NULL. This is an issue because the hud tracks
resources from the scene associated with the context and calling
ngl_set_scene() with a NULL scene would leave an existing hud tracking
already de-allocated resources alive leading to a memory crash on the
next call to ngl_draw().

This commit fixes the issue by:
- destroying the hud unconditionally before destroying the scene
- recreating the hud whether the new scene is NULL or not

The memory issue can be reproduced by the following C code:

    struct ngl_ctx *ctx = ngl_create();
    struct ngl_config cfg = {.offscreen = 1, .width = 1, .height = 1, .hud = 1};
    ngl_configure(ctx, &cfg);
    struct ngl_node *texture = ngl_node_create(NGL_NODE_TEXTURE2D);
    ngl_node_param_set_i32(texture, "width", 16);
    ngl_node_param_set_i32(texture, "height", 16);
    ngl_set_scene(ctx, texture);
    ngl_node_unrefp(&texture);
    ngl_set_scene(ctx, NULL);
    ngl_draw(ctx, 0.0);
    ngl_freep(&ctx);


libnodegl/api.c
```diff
@@ -249,6 +249,7 @@ static int cmd_set_scene(struct ngl_ctx *s, void *arg)
 {
     ngli_gpu_ctx_wait_idle(s->gpu_ctx);
 
+    ngli_hud_freep(&s->hud);
```
cboesch-gpsw, line 252: Don't you want this close to the `ngli_rnode_reset()` below?

Which brings me to the question: shouldn't we have it in `cmd_reset()` as well?

mbouron-gpsw: > Don't you want this close to the `ngli_rnode_reset()` below?

I prefer to have it before detaching the scene as the hud depends on the scene so it makes to deallocate it first.

> 
> Which brings me to the question: shouldn't we have it in `cmd_reset()` as well?

It's already there in cmd_reset(), I added a separate commit to move the call to ngli_hud_freep() before detaching the scene.

mbouron-gpsw: I also added a third commit factoring out the scene reset code into a dedicated function scene_reset().

cboesch-gpsw (approved):
Thanks for introducing `scene_reset()`, I like this PR much better now.

PR #385, opened 2022-02-04
pgcraft: use hwmap to scope specific samplers code paths

mbouron-gpsw:


libnodegl/pgcraft.c
```diff
@@ -268,9 +268,10 @@ static int prepare_texture_info_fields(struct pgcraft *s, const struct pgcraft_p
     for (int i = 0; i < NGLI_INFO_FIELD_NB; i++) {
         struct pgcraft_texture_info_field *field = &info->fields[i];
 
-        field->type = types_map[i];
-        if (field->type == NGLI_TYPE_NONE)
+        const int type = types_map[i];
+        if (type == NGLI_TYPE_NONE)
             continue;
+        field->type = type;
```
cboesch-gpsw, line 291: So previously `field->type` was set, even when it was `NGLI_TYPE_NONE`, which is not the case anymore. I'm assuming this is not an issue?

cboesch-gpsw (commented):
Not a huge fan of having the support check and the generated code so distant (due to the risk of desync between the 2 / mistake), but let's say it's fine for now, until that code is somehow moved to their respective backends.

libnodegl/pgcraft.c
mbouron-gpsw: It should be ok as the info structure is 0-initialized according the current code workflow:
```c
        struct pgcraft_texture_info info = {
            .stage     = texture->stage,
            .precision = texture->precision,
            .texture   = texture->texture,
            .image     = texture->image,
            .format    = texture->format,
            .writable  = texture->writable,
        };

        int ret = prepare_texture_info_fields(s, params, graphics, texture, &info);
        if (ret < 0)
            return ret;

```

cboesch-gpsw: Thanks

cboesch-gpsw (approved):

cboesch-gpsw:
See https://github.com/gopro/gopro-lib-node.gl/pull/388 as an alternative.

mbouron-gpsw:
> See #388 as an alternative.

Rebased on top of your PR.

PR #386, opened 2022-02-04
math_utils: add NGLI_CLAMP()

mbouron-gpsw:


libnodegl/math_utils.h
```diff
@@ -28,6 +28,8 @@
 #define M_PI 3.14159265358979323846
 #endif
 
+#define NGLI_CLAMP(x, min, max) (((x) > (max)) ? (max) : (((x) < (min)) ? (min) : (x)))
```
cboesch-gpsw, line 31: Can we have this in `libnodegl/utils.h` instead, along with `NGLI_MIN` and `NGLI_MAX`?

And then you could define it as `NGLI_MAX(NGLI_MIN(x, max), min)`

cboesch-gpsw (changes requested):

libnodegl/math_utils.h
mbouron-gpsw: Done.

cboesch-gpsw (approved):

PR #390, opened 2022-02-09
Pgcraft misc 

mbouron-gpsw:


libnodegl/pgcraft.h
```diff
@@ -59,7 +59,22 @@ struct pgcraft_texture {
...
+     * Optional texture associated with a corresponding pgcraft_texture_info
+     * structure generated by pgcraft. It is also associated with a
+     * corresponding entry in the generated pipeline_resource_params structure
+     * and thus used as a default resource by the pipeline created with said
+     * pipeline_resource_params structure.
+     * This field *must* not be set if pgcraft_shader_tex_type is
+     * NGLI_PGCRAFT_SHADER_TEX_TYPE_VIDEO as providing a single texture as
+     * default resource for multiple binding points with potentially
+     * incompatible types is not possible.
+     */
```
cboesch-gpsw, line 72: Here is another proposition:
> Just like the other types (uniforms, blocks, attributes), this field exists in order to be transmitted to the pipeline (through the `pipeline_resource_params` destination). That way, the data of all these specified resources can be uploaded straight after the pipeline initialization (using `ngli_pipeline_set_resources()`).
>
> In the case of the texture though, there is one exception: if the specified `type` is `NGLI_PGCRAFT_SHADER_TEX_TYPE_VIDEO`, then the `texture` field **must** be `NULL`. Indeed, this type implies special video accelerated samplings, which means the user texture may actually be backed by several samplers at different binding points unknown at pipeline init, but known later at runtime (after decoding a video frame). The actual binded textures are found in the image field.

libnodegl/pgcraft.h
```diff
@@ -59,7 +59,22 @@ struct pgcraft_texture {
...
     struct texture *texture;
+    /*
+     * Optional image associated with a corresponding pgcraft_texture_info
+     * structure generated by pgcraft.
+     */
```
cboesch-gpsw, line 77: > Just like the `texture` field, the `image` field is a bit special: if `texture` is specified (because known at init), then this field is not necessary for `ngli_pipeline_set_resources()` and thus can be ignored. Since it is transmitted back to the user through the `pgcraft_texture_info` structure, the user may still optionally set it if they plan to get some information back and honor them. In the case of `NGLI_PGCRAFT_SHADER_TEX_TYPE_VIDEO`, this field becomes pretty much mandatory if the user wants to know how to safely bind that video texture.

libnodegl/pgcraft.h
mbouron-gpsw: I reworded some part of your proposal, is that ok with you ?
```    /*
     * Just like the other types (uniforms, blocks, attributes), this field
     * exists in order to be transmitted to the pipeline (through the
     * pipeline_resource_params destination). That way, these resources can be
     * associated with the pipeline straight after the pipeline initialization
     * (using ngli_pipeline_set_resources()). In the case of the texture
     * though, there is one exception: if the specified type is
     * NGLI_PGCRAFT_SHADER_TEX_TYPE_VIDEO, then the texture field must be NULL.
     * Indeed, this type implies the potential use of multiple samplers (which
     * can be hardware accelerated and platform/backend specific) depending on
     * the image layout. This means that a single texture cannot be used as a
     * default resource for all these samplers. Moreover, the image layout
     * (which determines which samplers are used) is generally unknown at
     * pipeline initialization and it is only known once a frame has been
     * decoded/mapped. The image structure describes which layout to use and
     * which textures to bind and the pgcraft_texture_info.fields describes
     * where to bind the textures and their associated data.
     */
```

cboesch-gpsw: yep, thanks

libnodegl/pgcraft.h
cboesch-gpsw: OK

mbouron-gpsw: Commit updated with both section reworded.

cboesch-gpsw (approved):

PR #393, opened 2022-02-14
Pgcraft misc v2

mbouron-gpsw:


tests/data.py
```diff
@@ -532,3 +530,15 @@ def data_vertex_and_fragment_blocks(cfg):
...
+@test_cuepoints(points={'c': (0, 0)}, nb_keyframes=1, tolerance=1)
+@scene()
+def data_vertex_and_fragment_blocks(cfg):
+    return _data_vertex_and_fragment_blocks(cfg)
+
+
+@test_cuepoints(points={'c': (0, 0)}, nb_keyframes=1, tolerance=1)
+@scene()
+def data_vertex_and_fragment_blocks_std430(cfg):
+    return _data_vertex_and_fragment_blocks(cfg, True)
```
cboesch-gpsw, line 544: Can you explicitly send the layout instead, and do it in both calls?
```python
def _data_vertex_and_fragment_blocks(cfg, layout):
    ...


def data_vertex_and_fragment_blocks(cfg):
    return _data_vertex_and_fragment_blocks(cfg, layout='std140')

...

def data_vertex_and_fragment_blocks_std430(cfg):
    return _data_vertex_and_fragment_blocks(cfg, layout='std430')
```

libnodegl/pgcraft.c
```diff
@@ -168,6 +168,19 @@ static const char *get_glsl_type(int type)
     return ret;
 }
 
+static int get_next_available_binding(struct pgcraft *s, int stage, int type)
```
cboesch-gpsw, line 171: Not a fan of using the `get_*` semantic for functions altering the state. Maybe `query_next_binding()` or `request_next_binding()`

libnodegl/pipeline.h
```diff
@@ -101,6 +97,13 @@ struct pipeline_params {
     int nb_attributes;
 };
 
+struct pipeline_params {
+    int type;
+    const struct pipeline_graphics graphics;
+    const struct program *program;
+    struct pipeline_layout layout;
```
cboesch-gpsw: In the commit message ("pipeline: add the layout fields to a dedicated pipeline_layout structure"), can you replace the "add" with "move"?

cboesch-gpsw (changes requested):
LGTM aside from the cosmetics

libnodegl/pipeline.h
mbouron-gpsw: Done.

tests/data.py
mbouron-gpsw: Done.

libnodegl/pgcraft.c
mbouron-gpsw: Renamed to `request_next_binding()`

cboesch-gpsw (approved):

PR #400, opened 2022-02-25
resourceprops: set opts_size instead of priv_size

mbouron-gpsw:
The node options structure size must be be set to opts_size not
priv_size. This prevents the node options from being reset to 0 at
uninit.


cboesch-gpsw (approved):
Oups, LGTM thanks: but please remove the double "be be" in the commit description.

PR #401, opened 2022-03-04
Add a program label to shader compile errors

mbouron-gpsw:


libnodegl/pass.h
```diff
@@ -31,6 +31,7 @@ struct ngl_ctx;
 
 struct pass_params {
     const char *label;
+    const char* program_label;
```
cboesch-gpsw, line 34: misplaced `*`

libnodegl/backends/gl/program_gl.c
```diff
@@ -289,7 +289,8 @@ int ngli_program_gl_init(struct program *s, const struct program_params *params)
         if (ret < 0) {
             char *s_with_numbers = ngli_numbered_lines(shaders[i].src);
             if (s_with_numbers) {
-                LOG(ERROR, "failed to compile:\n%s", s_with_numbers);
+                LOG(ERROR, "failed to compile shader (label=%s):\n%s",
+                    params->label ? params->label : "undefined", s_with_numbers);
```
cboesch-gpsw, line 294: nitpick, I'd prefer:
```c
                LOG(ERROR, "failed to compile shader \"%s\":\n%s",
                    params->label ? params->label : "", s_with_numbers);
```

Can we also add a commit to assert that it is always set after every code unit has been updated so we can remove the conditional printing?

libnodegl/backends/gl/program_gl.c
```diff
@@ -271,7 +271,8 @@ int ngli_program_gl_init(struct program *s, const struct program_params *params)
     struct gpu_ctx_gl *gpu_ctx_gl = (struct gpu_ctx_gl *)s->gpu_ctx;
     struct glcontext *gl = gpu_ctx_gl->glcontext;
 
-    if (params->compute && (gl->features & NGLI_FEATURE_GL_COMPUTE_SHADER_ALL) != NGLI_FEATURE_GL_COMPUTE_SHADER_ALL) {
+    const uint64_t features = NGLI_FEATURE_GL_COMPUTE_SHADER_ALL;
+    if (params->compute && (gl->features & features) != features) {
```
cboesch-gpsw, line 275: Alternatively you could expose `libnodegl/api.c:#define ALL_FEATURES(features, mask) ((features & (mask)) == mask)` internally and use that.

Not a blocking comment.

libnodegl/hud.c
```diff
@@ -1263,6 +1263,7 @@ int ngli_hud_init(struct hud *s)
     graphicstate.blend_dst_factor_a = NGLI_BLEND_FACTOR_ONE;
 
     const struct pgcraft_params crafter_params = {
+        .program_label    = "hud",
```
cboesch-gpsw, line 1266: Shouldn't we make a distinction between user label and internal labels? Maybe something like "internal/hud"", "nodegl/hud", "ngl-hud", ...

cboesch-gpsw (changes requested):
Small nitpicks, LGTM otherwise

libnodegl/hud.c
mbouron-gpsw: Picked `nodegl/`.

libnodegl/pass.h
mbouron-gpsw: Fixed.

libnodegl/backends/gl/program_gl.c
mbouron-gpsw: Will do something like that later on for the whole code base since we use this pattern in other places.

cboesch-gpsw (approved):

libnodegl/backends/gl/program_gl.c
mbouron-gpsw: I prefer to keep the label optional for now.

PR #410, opened 2022-03-15
Adding 2 API tests

cboesch-gpsw:


tests/api.py
```diff
@@ -291,3 +291,14 @@ def api_reset_scene(width=320, height=240):
...
+
+def api_shader_init_fail(width=320, height=240):
+    viewer = ngl.Context()
+    assert viewer.configure(offscreen=1, width=width, height=height, backend=_backend) == 0
+
+    render = ngl.Render(ngl.Quad(), ngl.Program(vertex='<bug>', fragment='<bug>'))
+
+    assert viewer.set_scene(render) != 0
+    assert viewer.set_scene(render) != 0  # another try to make sure the state stays consistent
+    assert viewer.draw(0) == 0  # XXX: drawing with no scene is allowed?
```
mbouron-gpsw, line 304: drawing with no scene is allowed, you exercise it in the previous commit, you can remove the `XXX:`

mbouron-gpsw:
Other than my remark, LGTM.

tests/api.py
cboesch-gpsw: The comment is copied from another instance in the same file. Note that the previous commit doesn't check the return code, this one does.

Should we define the behavior of drawing with no scene? Is it defined clear the FBO or leave it unchanged?

mbouron-gpsw: It clears the default framebuffer using ngl_config.clear_color.

cboesch-gpsw: Thanks. Removed the `XXX` and added a commit to remove the other one.

mbouron-gpsw (approved):

PR #427, opened 2022-04-08
Maths anchor

cboesch-gpsw:


mbouron-gpsw:
Don't we want to avoid branching in such low level functions ? At some point we will want to pack data and operations and having a branch in the hot loop doesn't seem good.

cboesch-gpsw:
> Don't we want to avoid branching in such low level functions ? At some point we will want to pack data and operations and having a branch in the hot loop doesn't seem good.

We could make multiple versions of these functions, but note that the condition is already executed in the hot path currently (with `use_anchor`), so it feels like it could be done later.

mbouron-gpsw (approved):

PR #429, opened 2022-04-14
nodes: reduce activity check nodes array length

cboesch-gpsw:
activitycheck_nodes is only used to honor release/prefetch, so we only
need to register these nodes.

Before:
  Rendered 1800 frames in 3.1482 (FPS=571.755)
  Rendered 1800 frames in 3.21107 (FPS=560.562)
  Rendered 1800 frames in 3.31502 (FPS=542.984)

After:
  Rendered 1800 frames in 3.02839 (FPS=594.376)
  Rendered 1800 frames in 2.99905 (FPS=600.19)
  Rendered 1800 frames in 3.00258 (FPS=599.485)


mbouron-gpsw:
Can you describe the scene you used for the benchmark ?

cboesch-gpsw:
> Can you describe the scene you used for the benchmark in the commit description ?

Added the following mention:

    Using a large scene (~30 seconds long, 228 time range filters, a total
    of about 47000 nodes):

mbouron-gpsw (approved):

PR #432, opened 2022-04-16
API misc

mbouron-gpsw:


CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Versioning](https://semver.org/spec/v2.0.0.html) for `libnodegl`.
   backends were enabled.
 - A deadlock in `ngl-control` when specifying invalid encoding arguments
 
+### Removed
+- handle field from the ngl_config structure
```
cboesch-gpsw, line 23: can you add backticks to `handle` and `ngl_config`?

libnodegl/api.c
```diff
@@ -665,6 +665,11 @@ struct ngl_ctx *ngl_create(void)
 
 int ngl_configure(struct ngl_ctx *s, struct ngl_config *config)
 {
+    if (s->configured) {
+        dispatch_cmd(s, cmd_reset, &(int[]){KEEP_SCENE});
+        s->configured = 0;
+    }
```
cboesch-gpsw, line 659: Why not calling `cmd_reset()` at the beginning of `cmd_configure()` as well as moving all the config checks below?

cboesch-gpsw (changes requested):

libnodegl/api.c
mbouron-gpsw: Because cmd_reset() might not be executed on the same thread as cmd_configure() which depends on the platform (and in the future if the context is external).

CHANGELOG.md
mbouron-gpsw: Done.

libnodegl/api.c
cboesch-gpsw: But `cmd_reset()` is already called from within `cmd_configure()`?

mbouron-gpsw: Yes, in case of failure and in the thread where the opengl context being configured is current.

cboesch-gpsw (approved):

PR #433, opened 2022-04-18
GL leaks detection

cboesch-gpsw:


scripts/leaks.sh
```diff
@@ -0,0 +1,17 @@
...
+trap "rm -f $tmpfile" EXIT
+apitrace trace -a egl -o $tmpfile $@
+
+# executable doesn't exit in error so we look for "errors" in stderr
+leaks=$(apitrace leaks $tmpfile 2>&1 > /dev/null)
+if echo "$leaks" | grep -q "error:"; then
+    echo "GPU leaks detected: YES"
+    echo "$leaks"
+    exit 1
+else
```
mbouron-gpsw, line 15: Will this script stay dedicated to OpenGL (apitrace only support OpenGL and DX up to 11) ? If so, can we rename it to something more explicit ? (`detect-gl-leaks.sh`, `gl-leaks.sh`)
Another question, is renderdoc also able to detect these issues ?

cboesch-gpsw: > Will this script stay dedicated to OpenGL (apitrace only support OpenGL and DX up to 11) ? If so, can we rename it to something more explicit ? (`detect-gl-leaks.sh`, `gl-leaks.sh`)

Hard to say; I don't see `apitrace leaks` being extended to support Vulkan typically. I'm expecting this kind of stuff to be replaced by a debug layer typically. Renamed to `gl-leaks.sh`.

> Another question, is renderdoc also able to detect these issues ?

I couldn't find anything in the documentation associated with memory leaks detection.

PR #444, opened 2022-04-23
Misc fixes/improvements

cboesch-gpsw:


libnodegl/backends/vk/glslang_utils.c
```diff
@@ -144,7 +144,7 @@ int ngli_glslang_compile(int stage, const char *src, void **datap, size_t *sizep
 void ngli_glslang_uninit(void)
 {
     pthread_mutex_lock(&lock);
-    if (refcount && (refcount-- == 0))
+    if (refcount && (--refcount == 0))
```
mbouron-gpsw, line 147: Can we use `refcount-- == 1` instead to be consistent with the rest of the codebase ?

cboesch-gpsw: Sure, done

libnodegl/utils.c
```diff
@@ -52,7 +52,7 @@ char *ngli_strdup(const char *s)
         r = ngli_malloc(len + 1);
         if (!r)
             return NULL;
-        strcpy(r, s);
+        memcpy(r, s, len);
```
mbouron-gpsw: `len` does not take into account the trailing `\0` and r is not calloc-ed.

mbouron-gpsw (approved):

PR #445, opened 2022-04-25
Stencil split toolbox helper

cboesch-gpsw:


pynodegl-utils/pynodegl_utils/toolbox/scenes.py
```diff
@@ -0,0 +1,60 @@
...
+        scene_mask,
+        color_write_mask="",  # no need to write the color themselves, only the stencil
+        stencil_test=True,
+        stencil_ref=1,
+        stencil_read_mask=0,
+        stencil_write_mask=0xFF,
+        stencil_fail="replace",
+        stencil_depth_fail="replace",
+        stencil_depth_pass="replace",
+    )
```
mbouron-gpsw, line 41: You may want to clear the whole stencil buffer with 0 so if the underlying scene use it, it does not interfere with the compare method.

mbouron-gpsw:
`because we are doing 2 draws instead of one)`: it is actually 3 draws (4 if you clear the whole stencil buffer with 0)

pynodegl-utils/pynodegl_utils/toolbox/scenes.py
mbouron-gpsw: Or state the limitation regarding the stencil usage in the documentation of the function.

cboesch-gpsw: Done

cboesch-gpsw:
> `because we are doing 2 draws instead of one)`: it is actually 3 draws (4 if you clear the whole stencil buffer with 0)

Message adjusted

mbouron-gpsw (approved):

PR #450, opened 2022-04-30
project: add clangd compilation database references

cboesch-gpsw:
This helps editors relying on Language Server Protocol to find the
sources out of the box.


mbouron-gpsw (approved):
Tested successfully with neovim+coc+clangd and code.

PR #452, opened 2022-05-02
UniformColor + AnimatedColor (with HSV/HSL colorspaces)

cboesch-gpsw:


libnodegl/colorconv.c
```diff
@@ -150,3 +153,70 @@ int ngli_colorconv_get_ycbcr_to_rgb_color_matrix(float *dst, const struct color_
...
+    const float h = hsv[0], s = hsv[1], v = hsv[2];
+    const float h6 = h * 6.f;
+    const float c = v * s;
+    dst[0] = (sat(fabsf(h6 - 3.f) - 1.f) - 1.f) * c + v;
+    dst[1] = (sat(2.f - fabsf(h6 - 2.f)) - 1.f) * c + v;
+    dst[2] = (sat(2.f - fabsf(h6 - 4.f)) - 1.f) * c + v;
+}
+
+/* See HSL-to-linear comment */
+void ngli_colorconv_hsv2linear(float *dst, const float *hsv)
```
mbouron-gpsw, line 227: You started to use restrict in a bunch of math functions, would it make sense here ?

cboesch-gpsw: I wondered about it. But I think it makes sense to be able to use it in-place. In practice `restrict` won't make a difference here because I'm reading hsv fully before making any transform.

mbouron-gpsw: Ok, thanks.

mbouron-gpsw (approved):

PR #453, opened 2022-05-02
pynodegl & pyright

cboesch-gpsw:


mbouron-gpsw:
`pynodegl: expose symbols through __all__ ` can you give an example of linters and how to reproduce the warnings ?

cboesch-gpsw:
> `pynodegl: expose symbols through __all__ ` can you give an example of linters and how to reproduce the warnings ?

Sure, here is an example with pylsp (flake8 plugin):
2022-05-03-123824-ea7Loe3o

And here is one with pyright:
2022-05-03-124209-Ti5LooZu

I believe you can reproduce these warnings using `flake8` and `pyright` command line interfaces.

mbouron-gpsw:
Reproduced the warnings with flake8.

mbouron-gpsw (approved):

PR #456, opened 2022-05-05
ci/linux: update to Ubuntu 22.04

mbouron-gpsw:


mbouron-gpsw:
> Should be fine, just a few things:
> 
>     1. can we merge the stress-test PR that includes the apitrace drop first?

Sure, it was my plan to wait for the stress test PR to be merged before this one.

> 
>     2. I'm assuming the bump of Ubuntu in `.github/workflows/ci_android.yml` is problematic?

I didn't try yet. I planned to do that later on as I also plan to bump the ndk version.

PR #458, opened 2022-05-05
Hooks fix

cboesch-gpsw:


mbouron-gpsw:
`One case of this can be reproduced with the following:` +scenario

mbouron-gpsw:
`This commit addresses this issue by doing 2 things:` + extra line break (to make it consistent with the previous list)

mbouron-gpsw (approved):

PR #461, opened 2022-05-09
hmap: fix bucket use after free

cboesch-gpsw:
If a bucket ends up being empty after a deletion, the pointer must be
reset to NULL or it will cause issue for the following realloc (when
re-adding an entry in that same bucket) or free (when deleting the
hmap).


mbouron-gpsw:
Did you trigger the issue with a specific scenario using the libnodegl API ?

cboesch-gpsw:
> Did you trigger the issue with a specific scenario using the libnodegl API ?

Yes, by coincidence in the dict test here: https://github.com/gopro/gopro-lib-node.gl/pull/462/

PR #467, opened 2022-05-11
Log trim

cboesch-gpsw:


libnodegl/backends/vk/vkcontext.c
```diff
@@ -70,7 +70,10 @@ static VKAPI_ATTR VkBool32 VKAPI_CALL debug_callback(VkDebugUtilsMessageSeverity
     if (type & VK_DEBUG_UTILS_MESSAGE_TYPE_VALIDATION_BIT_EXT)       msg_type = "VALIDATION";
     else if (type & VK_DEBUG_UTILS_MESSAGE_TYPE_PERFORMANCE_BIT_EXT) msg_type = "PERFORMANCE";
 
-    ngli_log_print(level, __FILE__, __LINE__, "debug_callback", "%s: %s", msg_type, cb_data->pMessage);
+    int msg_len = strlen(cb_data->pMessage);
+    while (msg_len > 0 && strchr(" \r\n", cb_data->pMessage[msg_len - 1]))
```
mbouron-gpsw, line 74: Would it be possible to use isspace() ? (and would it make sense ?)

mbouron-gpsw: Nevermind, isspace is locale depend, it looks like it's not a good idea.

mbouron-gpsw (approved):

PR #470, opened 2022-05-14
Render* + Geometry fix

cboesch-gpsw:


tests/shape.py
```diff
@@ -186,6 +186,34 @@ def shape_geometry_normals_indices(cfg):
...
 
 
+@test_fingerprint()
+@scene()
+def shape_geometry_with_renderother(cfg):
+    cfg.aspect_ratio = (1, 1)
+
+    vertices_data = array.array("f")
+    uvcoords_data = array.array("f")
+    boxes = (
```
mbouron-gpsw, line 196: `boxes` is a bit confusing, would `boundaries` ok ?

mbouron-gpsw:
Other my comment, LGTM.

tests/shape.py
cboesch-gpsw: Ah yeah sure, it was boxes in a past iteration (a list of x,y pairs), and I forgot to rename. Done.

mbouron-gpsw (approved):

PR #471, opened 2022-05-15
Blocks & buffers

cboesch-gpsw:


mbouron-gpsw:
` buffer: make buffer and layout offset consistent view block view ` -> ` buffer: make buffer and layout offset consistent with block view` ?

libnodegl/internal.h
```diff
@@ -200,6 +200,8 @@ int ngli_node_buffer_ref(struct ngl_node *node);
 int ngli_node_buffer_init(struct ngl_node *node);
 void ngli_node_buffer_unref(struct ngl_node *node);
 int ngli_node_buffer_upload(struct ngl_node *node);
+int ngli_node_buffer_cpu_size(struct ngl_node *node);
+int ngli_node_buffer_gpu_size(struct ngl_node *node);
```
mbouron-gpsw, line 204: Can you name them ngli_node_buffer_get_{cpu,gpu}_size() ?

cboesch-gpsw: sure, done

libnodegl/pass.c
```diff
@@ -345,15 +345,14 @@ static int register_attribute(struct pass *s, const char *name, struct ngl_node
...
-        ngli_node_block_extend_usage(block_node, NGLI_BUFFER_USAGE_VERTEX_BUFFER_BIT);
     } else {
         stride = attribute_priv->layout.stride;
         offset = 0;
         buffer = attribute_priv->buffer;
-        ngli_node_buffer_extend_usage(attribute, NGLI_BUFFER_USAGE_VERTEX_BUFFER_BIT);
     }
 
     const int attr_type = attribute_priv->layout.type;
+    ngli_node_buffer_extend_usage(attribute, NGLI_BUFFER_USAGE_VERTEX_BUFFER_BIT);
```
mbouron-gpsw, line 335: Can you move this call after the `else {` block, as `attr_type` is directly used by `crafter_attribute`.

libnodegl/node_block.c
```diff
@@ -106,6 +106,11 @@ static const struct param_choices layout_choices = {
                                        NGL_NODE_TIME,                   \
                                        -1}
 
+struct block_priv {
+    struct block_info blk;
```
mbouron-gpsw, line 110: Can you name it `block_info`, `info` or `bi` as `blk` is a bit confusing.

libnodegl/node_buffer.c
```diff
@@ -256,6 +257,17 @@ static int buffer_init_from_count(struct ngl_node *node)
     return 0;
 }
 
+static const struct block_field *get_field(const struct darray *fields_array, const char *name)
```
mbouron-gpsw, line 260: `get_block_field()`

mbouron-gpsw:
In the commit `buffer: work with the block structure instead of the node`, can you mention that it will allow to make block_opts (and maybe another struct ?) private in a later commit ?

libnodegl/node_block.c
cboesch-gpsw: This is consistent with `buf` in the `buffer_priv` for the `buffer_info`; how do you want to rename `buf`?

libnodegl/pass.c
cboesch-gpsw: done

libnodegl/node_buffer.c
cboesch-gpsw: done

cboesch-gpsw:
> In the commit `buffer: work with the block structure instead of the node`, can you mention that it will allow to make block_opts (and maybe another struct ?) private in a later commit ?

Mmh technically this commit doesn't allow this but only makes the code consistent with the code that goes away in geometry and pass. It does allow accessing the field offset in a later commit though, which is the information we don't are lacking when only poking at the node level.

I've updated the description.

mbouron-gpsw (approved):

PR #472, opened 2022-05-16
Ensure that color is not negative for sRGB <-> linear conversions

mbouron-gpsw:


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Versioning](https://semver.org/spec/v2.0.0.html) for `libnodegl`.
   the Wayland platform ID)
 - Undefined behaviour when using geometries with `Render*` nodes (`Render` node
   is not affected)
+- Undefined behaviour in GLSL sRGB conversions with negative colors.
```
cboesch-gpsw, line 23: Can you:
1. remove the trailing point
2. remove "GLSL"
3. replace "sRGB" with "sRGB/linear"?

For the commit description, I suggest:
```
From the pow() documentation:

    Results are undefined if x<0 or if x=0 and y=0.

Fixes undefined behaviour on multiple GPUs/backends.

If the color is negative the branches with the pow() are not supposed to be
taken, but they are calculated anyway and end up being used in the mix code:
mix(v, NaN, 0.0). Even if 0% of the NaN is taken, it is part of the formula so
the undefined behaviour is propagated.

We decide to clamp only in the pow() and not the input color in order to
preserve the formula for negative colors: indeed, it might make sense to
temporarily have negative color components (preserving the proportions between
them) and have them corrected up later down the pipeline. Clamping inside the
pow() has no impact over the expected result because the branch is entirely
discarded.

The CPU functions ngli_colorconv_srgb2linear() and ngli_colorconv_linear2srgb()
are not affected because the pow() branches are never taken with negative values.
```

cboesch-gpsw (changes requested):

tests/color.py
```diff
@@ -59,3 +59,16 @@ def scene_func(_):
 color_static_srgb = _get_static_color_scene_func((1.0, 0.5, 0.0), "srgb")
 color_static_hsl = _get_static_color_scene_func((0.6, 0.9, 0.4), "hsl")
 color_static_hsv = _get_static_color_scene_func((0.3, 0.7, 0.6), "hsv")
+
+
+@test_cuepoints(points={"c": (0, 0)}, nb_keyframes=10, tolerance=0)
+@scene()
+def color_negative_values_srgb_gpu(cfg):
+    cfg.duration = 5
+    kfs = (
```
cboesch-gpsw, line 68: Can you add a comment saying that "elastic_in has the special property of undershooting below 0"?

CHANGELOG.md
mbouron-gpsw: Done. Thanks.

cboesch-gpsw (approved):

PR #473, opened 2022-05-16
Misc cleanups

cboesch-gpsw:


libnodegl/backends/gl/texture_gl.h
```diff
@@ -36,6 +36,10 @@ struct texture_gl {
     GLint format;
     GLint internal_format;
     GLenum format_type;
+
```
mbouron-gpsw, line 39: Can you remove this extra line ?

mbouron-gpsw (approved):
Except from my comment, LGTM.

libnodegl/backends/gl/texture_gl.h
cboesch-gpsw: Yup done

mbouron-gpsw:
Do you mind scoping only the wrapped field to texture_gl in your last commit ? I will provide a commit to make use of external_storage / bytes_per_pixel in texture_vk if that's ok with you ?

cboesch-gpsw:
> Do you mind scoping only the wrapped field to texture_gl in your last commit ? I will provide a commit to make use of external_storage / bytes_per_pixel in texture_vk if that's ok with you ?

Is this going to be set in the common path or every backend is going to have the responsibility to keep these fields consistent?

mbouron-gpsw:
> Is this going to be set in the common path or every backend is going to have the responsibility to keep these fields consistent?

I guess yes but after a third thought, you can keep your commit as is. Sorry for the noise.

PR #474, opened 2022-05-17
buffer_vk: allow fallback on uncached memory

mbouron-gpsw:
Host cached memory might not be available depending on the GPU device.
The only downside of this fallback is that host memory accesses to
uncached memory are slower than to cached memory.

This allows Android devices to use the node.gl `capture_buffer` API.


cboesch-gpsw (approved):
Sounds like something you may want to document on the changelog

mbouron-gpsw:
> Sounds like something you may want to document on the changelog

Changelog updated, let me know if you are ok with the message.

PR #475, opened 2022-05-18
Texture misc

mbouron-gpsw:


libnodegl/backends/gl/hwmap_vaapi_gl.c
```diff
@@ -82,7 +83,21 @@ static int vaapi_init(struct hwmap *hwmap, struct sxplayer_frame *frame)
         return -1;
     }
 
+    ngli_glGenTextures(gl, 1, vaapi->gl_planes);
```
cboesch-gpsw, line 86: shouldn't this be 2 planes?

libnodegl/backends/gl/hwmap_videotoolbox_darwin_gl.c
```diff
@@ -111,7 +111,7 @@ static int vt_darwin_map_plane(struct hwmap *hwmap, IOSurfaceRef surface, int in
         return -1;
     }
 
-    ngli_glBindTexture(gl, GL_TEXTURE_RECTANGLE, 0);
+    ngli_glBindTexture(gl, plane_gl->target, 0);
```
cboesch-gpsw, line 114: Why abstract it here but not in `vt_darwin_init()`?

cboesch-gpsw (changes requested):
I like this PR

libnodegl/backends/gl/hwmap_vaapi_gl.c
mbouron-gpsw: Good catch, it is fixed. For the record, playing back a single media was working fine, the special texture handle 0 was used for the chroma plane without any issue :')

libnodegl/backends/gl/hwmap_videotoolbox_darwin_gl.c
mbouron-gpsw: I changed the commit and went the other way around (ie: using GL_TEXTURE_RECTANGLE where relevant).

libnodegl/backends/gl/hwmap_vaapi_gl.c
cboesch-gpsw: Hopefully https://github.com/gopro/gopro-lib-node.gl/pull/479 will help here

cboesch-gpsw (approved):

PR #478, opened 2022-05-19
gpu_ctx_vk: wait any pending semaphores before executing the update c…

ghost:
…ommand buffer

Fixes the following validation error when the graph is only updated and
not drawn (ie: when visualizing the graph at a given time in
ngl-control):

vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] is signaling VkQueue
0x55cd607f3cc0[] (VkSemaphore 0xf56c9b0000000004[]) that was previously
signaled by VkQueue 0x55cd607f3cc0[] but has not since been waited on by
any queue.


libnodegl/backends/vk/gpu_ctx_vk.c
```diff
@@ -1023,7 +1023,24 @@ static int vk_begin_update(struct gpu_ctx *s, double t)
...
+    if (res != VK_SUCCESS)
+        return res;
+
+    VkSemaphore *wait_sems = ngli_darray_data(&s_priv->pending_wait_sems);
+    for (int i = 0; i < ngli_darray_count(&s_priv->pending_wait_sems); i++) {
+        res = ngli_cmd_vk_add_wait_sem(s_priv->cur_cmd,
+                                       &wait_sems[i],
+                                       VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT |
+                                       VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT |
+                                       VK_PIPELINE_STAGE_TRANSFER_BIT);
```
cboesch-gpsw, line 1036: I'm not familiar with VK but what's the risk of missing a flag here?

Also, can't we have a test to cover this? If it's reproducible with ngl-control, it should be reproducible with `tests/api.py`?

hsong-gpsw (approved):
good catch,

rlavarec-gpsw (approved):

PR #480, opened 2022-05-20
utils/assets: replace H264 mire with a cat video

cboesch-gpsw:
The H264 mire is in YUV444P. This typically causes the VideoToolbox
decoder to fallback on its own software decoder, which then also fails.
Unfortunately, it fails too late for a fallback on a working software
decoder (because the Apple decoder accepts the video but fails to return
frames). In this commit we replace that video with a 169K video of a cat
(of about 10 seconds), in H264 and YUV420P.

The original mire is kept for future experiments, but it is disabled
from the available samples in all scenes.


mbouron-gpsw (approved):
Except my offline remark regarding the commit desc. LGTM.

PR #481, opened 2022-05-21
Buffers & blocks - Part II

cboesch-gpsw:


mbouron-gpsw:
`buffer: move buffer init from type in a dedicated function` -> `buffer: move buffer initialization in a dedicated function`

mbouron-gpsw:
In `block: remove ref/unref mechanism` you use references to `ngli_node_block_{cpu,gpu}_size()` (that should be `ngli_node_block_get_{cpu,gpu}_size()`

libnodegl/node_buffer.c
```diff
@@ -351,6 +315,16 @@ static int buffer_init(struct ngl_node *node)
     if (ret < 0)
         return ret;
 
+    if (s->buf.block) {
+        const struct block_info *block_info = o->block->priv_data;
```
mbouron-gpsw, line 279: Can you use `s->buf.block->priv_data ` so it is consistent with the check above and the one in the uninit function ?

libnodegl/node_animatedbuffer.c
```diff
@@ -132,6 +132,20 @@ static int animatedbuffer_init(struct ngl_node *node)
     return 0;
 }
 
+static int animatedbuffer_prepare(struct ngl_node *node)
+{
+    struct animatedbuffer_priv *s = node->priv_data;
+
+    if (s->buf.buffer->size)
```
mbouron-gpsw, line 142: Can you use an intermediate variable `buffer_info` or `info` ?

libnodegl/node_buffer.c
```diff
@@ -78,27 +78,27 @@ void ngli_node_buffer_extend_usage(struct ngl_node *node, int usage)
...
+    if (s->buf.buffer->size)
         return 0;
 
-    int ret = ngli_buffer_init(s->buffer, s->data_size, s->usage);
+    int ret = ngli_buffer_init(s->buf.buffer, s->buf.data_size, s->buf.usage);
     if (ret < 0)
         return ret;
 
-    ret = ngli_buffer_upload(s->buffer, s->data, s->data_size, 0);
+    ret = ngli_buffer_upload(s->buf.buffer, s->buf.data, s->buf.data_size, 0);
```
mbouron-gpsw: Performing the buffer initialization in the prepare callback leads to:
- allocating an unused GPU buffer in case the buffer is used as data source for a texture
- allocating an unused GPU buffer in case the buffer is used as a field of a block
Moreover, moving the GPU logic directly in the node callbacks will prevent us from doing any kind of packing or interleaving (which can be useful for the geometry for example).

libnodegl/node_buffer.c
cboesch-gpsw: Done

libnodegl/node_animatedbuffer.c
cboesch-gpsw: I added the following commits on top:
```
streamedbuffer: use an intermediate info variable for s->buf
animatedbuffer: use an intermediate info variable for s->buf
buffer: use an intermediate info variable for s->buf
block: use an intermediate info variable for s->blk
```

Tell me if that rename really is what you want. Note that for consistency we should do that for the `s->var.xxx` as well in the variables.

libnodegl/node_buffer.c
cboesch-gpsw: Oh wow, good point. That was a terrible mistake to upload buffers indiscriminately, I totally missed that.

Here is my proposal: I added a `NGLI_BUFFER_FLAG_GPU_UPLOAD` flag that is being requested explicitly by parents who want the buffer to be uploaded for them. This is the case for geometry and pass attributes, so I've used it here. It made me realize that the GPU size for buffers was actually wrong, so I've that flag there.

See the following commits:
```
buffer: honor GPU upload flag in ngli_node_buffer_get_gpu_size()
buffer: add a GPU upload flag
```

As a consequence of this addition, I took the liberty to add the following commit as well: `buffer: integrate dynamic flag into flags` to avoid the proliferation of booleans in that structure.

Thanks for pointing this out.

libnodegl/internal.h
```diff
@@ -179,6 +179,8 @@ struct buffer_layout {
     int offset;     // offset where the data starts in the buffer, in bytes
 };
 
+#define NGLI_BUFFER_FLAG_GPU_UPLOAD (1 << 0) /* The buffer is responsible for uploading its data to the GPU */
```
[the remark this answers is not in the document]
cboesch-gpsw: Done:
```c
#define NGLI_BUFFER_INFO_FLAG_GPU_UPLOAD (1 << 0) /* The buffer is responsible for uploading its data to the GPU */
#define NGLI_BUFFER_INFO_FLAG_DYNAMIC    (1 << 1) /* The buffer CPU data may change at every update */
```

libnodegl/node_render.c
```diff
@@ -182,7 +182,7 @@ static int render_init(struct ngl_node *node)
     struct pass_params params = {
         .label = node->label,
         .program_label = o->program->label,
-        .geometry = o->geometry,
+        .geometry = *(struct geometry **)o->geometry->priv_data,
```
mbouron-gpsw, line 181: Can you use an intermediate variable ?

cboesch-gpsw: Done. Also added a commit to make it `const`.

libnodegl/node_buffer.c
mbouron-gpsw: Thanks.

mbouron-gpsw (approved):
Tested locally, I haven't dedected any scenario where there is a superfluous GPU allocation and no validation error is reported nor a warning. Thanks and LGTM.

PR #483, opened 2022-05-24
build: handle OpenGLES as a separate backend

mbouron-gpsw:
Some platforms only support OpenGL or OpenGLES (for example, macOS only
supports OpenGL).

This prevents our probing mechanism from trying to create an OpenGLES
context with NSGL and outputting the following error message:

    nsgl_init: unsupported backend: 2, only OpenGL is supported by NSGL


libnodegl/meson.build
```diff
@@ -540,7 +572,7 @@ foreach gbackend_name, gbackend_cfg : gbackends_cfg
   message('backend: @0@ enabled: @1@'.format(gbackend_name, all_dep_found.to_string()))
   if all_dep_found
 
-    if gbackend_name == 'gl' and vaapi_enabled
+    if (gbackend_name == 'gl' or gbackend_name == 'gles') and vaapi_enabled
```
cboesch-gpsw, line 575: `gbackend_name in ['gl', 'gles']`

configure.py
```diff
@@ -614,7 +614,7 @@ def _run():
         "--debug-opts",
         nargs="+",
         default=[],
-        choices=("gl", "vk", "mem", "scene", "gpu_capture"),
+        choices=("gl", "gles", "vk", "mem", "scene", "gpu_capture"),
```
cboesch-gpsw, line 617: `libnodegl/meson_options.txt` doesn't have a `gles` option for `--debug-opts`, and I think that's fine: there is currently no way to have a build where gl and gles backends have a different debug system (this would imply the addition of runtime checks for `check_error_code` / `ngli_glcontext_check_gl_error`. I think you should leave this out and keep a `DEBUG_GL` shared for both until then.

cboesch-gpsw (changes requested):

libnodegl/meson.build
mbouron-gpsw: Done.

configure.py
mbouron-gpsw: Removed.

cboesch-gpsw (approved):
LGTM thanks

You may want to add some ifdefery in `gpu_ctx_gl.c` to prevent the systematic declarations of both `ngli_gpu_ctx_{gl,gles}`

cboesch-gpsw (approved):

PR #488, opened 2022-06-03
nodes: ensure nodes are updated after their activation

mbouron-gpsw:


libnodegl/nodes.c
```diff
@@ -627,6 +627,12 @@ int ngli_node_visit(struct ngl_node *node, int is_active, double t)
     const int queue_node = node->visit_time != t;
 
     if (queue_node) {
+        /*
+         * If a node is going to be activated but has already been updated for
+         * that time previously, we need to force its update.
+         */
```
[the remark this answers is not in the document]
mbouron-gpsw: Done.

tests/api.py
```diff
@@ -345,3 +345,42 @@ def api_shader_init_fail(width=320, height=240):
...
+    trf = _create_trf(group, start, start + 1)
+
+    trf_start = _create_trf(trf, start, start + 1)
+    group = ngl.Group(children=(trf,))
+    trf_end = _create_trf(group, end - 1.0, end + 1.0)
+
+    return ngl.Group(children=(trf_start, trf_end))
+
+
+def api_trf_seek(width=320, height=240):
```
cboesch-gpsw, line 373:
```python
"""
Run a special time sequence on a particularly crafted time filtered diamond-tree
graph to detect potential release/prefetch issues.
"""
```

mbouron-gpsw:
> > * first draw call (at 10.0) disables the left branch and enables the
> >   right one. The right branch is then updated/drawn (Group 1,
> >   TimeRangeFilters 2, Group 2 and TimeRangeFilters 3)
> > * second draw call (at 0.0) enables the left branch and disables the
> >   right one. The left branch is then updated/drawn  (Group 1,
> >   TimeRangeFilters 1, TimeRangeFilters 3, Group 3, RTT, RenderTexture)
> > * at this point, Group 2 is disabled but it has its last_update_time
> >   still set to 10.0. TimeRangeFilters 3 is allowing the descent into the
> >   graph.
> > * third draw call (at 10.0), disables the left branch (and de-allocated
> >   its resources which include the RTT and RenderTexture) and enables the
> >   right one. However, during the update operation, the descent is
> >   blocked by Group 2 which has its last_update_time equal to the current
> >   update time. This, in turn, prevents TimeRangeFilters 3 from being
> >   updated leading to the node keeping its previous state, allowing the
> >   descent into the graph while it should not for the current draw time.
> >   This ultimately leads to a crash because the draw operation is called
> >   on a branch that has been disabled/de-allocated.
> 
> Can you add numbers to each point? I'd like to refer to one point in particular later in the description (see suggestion below).

Done.

> 
> > This commit ensures the nodes are updated after their activation. This
> > ensures that, in the above scenario, Group 2 gets updated when activated
> 
> The dup of "ensures" sounds funny

Already changed to sounds less funny :D

> 
> > which, in turn, allows TimeRangeFilters 3 to be updated as it should be,
> > fixing the crash.
> 
> Could you add the following at the end?

Done.

> 
> > The reason why Group 2 has its last_update_time still set to 10.0 at step 2 is
> > because node_release() has not been called on it (which does reset the
> > last_update_time). This is due to a regression introduced in
> > d83b9e49f5ffcdb2d3f89cef0e1aa38a470c2dab¹ reducing the number of nodes onto
> > which we call node_prefetch()/node_release().
> > In order to avoid re-inserting a large number of nodes in the activity check
> > array, we instead add a special condition detecting that particular situation.
> > There were multiple other alternative ways of fixing the issue, but this one is
> > probably the simplest and most optimal. Typically the reason we reset the time
> > at prefetch instead of at release like before is because detecting a release is
> > much harder as it can be done only when everything has been visited: only one
> > use is enough imply a prefetch, but a release is conditioned by the lack of any
> > use (it requires full knowledge).
> > [1]: this commit is present in release v2022.4, v2022.5 and v2022.6

tests/api.py
```diff
@@ -345,3 +345,42 @@ def api_shader_init_fail(width=320, height=240):
...
+
+def _create_trf_scene(start, end):
+    texture = ngl.Texture2D(width=64, height=64)
+    rtt = ngl.RenderToTexture(ngl.Identity(), clear_color=(1.0, 0.0, 0.0, 1.0), color_textures=(texture,))
+    render = ngl.RenderTexture(texture=texture)
+    group = ngl.Group(children=(rtt, render))
+    trf = _create_trf(group, start, start + 1)
+
+    trf_start = _create_trf(trf, start, start + 1)
+    group = ngl.Group(children=(trf,))
```
cboesch-gpsw, line 367:
```python
# This group could be any node as long as it has no prefetch/release callback
```

tests/api.py
```diff
@@ -345,3 +345,42 @@ def api_shader_init_fail(width=320, height=240):
...
+
+    start = 0.0
+    end = 10.0
+    scene = _create_trf_scene(start, end)
+    ret = ctx.set_scene(scene)
+    assert ret == 0
+
+    assert ctx.draw(end) == 0
+    assert ctx.draw(start) == 0
+    assert ctx.draw(end) == 0
```
cboesch-gpsw, line 386:
```python
# The following time sequence is designed to create an inconsistent time state
```

cboesch-gpsw (changes requested):

tests/api.py
mbouron-gpsw: Done.

tests/api.py
mbouron-gpsw: Done.

tests/api.py
mbouron-gpsw: Done.

cboesch-gpsw (approved):

qdespinoy-gpsw (approved):

PR #494, opened 2022-09-05
sxplayer: use version 9.13.0

mrobertseidowsky-gpsw:


cboesch-gpsw:
Can you bump to `9.13.0`?

cboesch-gpsw (changes requested):
Can you bump to `9.13.0`?

PR #508, opened 2023-02-14
GPU capture support for macOS

steventang-gpsw:
This PR added OpenGL GPU capture for macOS.


libnodegl/backends/gl/gpu_ctx_gl.h
```diff
@@ -60,7 +60,11 @@ struct gpu_ctx_gl {
     struct texture *depth_stencil;
     /* Offscreen capture callback and resources */
     capture_func_type capture_func;
-#if defined(TARGET_IPHONE)
+#if  defined(TARGET_DARWIN)
```
kgrant-gpsw, line 63:
```suggestion
#if defined(TARGET_DARWIN)
```

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -131,6 +132,66 @@ static int wrap_capture_cvpixelbuffer(struct gpu_ctx *s,
...
+    };
+
+    const struct texture_gl_wrap_params wrap_params = {
+        .params  = &attachment_params,
+        .texture = fbName,
+    };
+
+    int ret = ngli_texture_gl_wrap(texture, &wrap_params);
+    if (ret < 0) {
+        CFRelease(cv_texture);
```
kgrant-gpsw, line 184: It looks like `CFRelease` is equivalent to `CVBufferRelease` except that the latter will implicitly check for `NULL` (`CFRelease` crashes if the texture is still `NULL`).

It is probably safe to assume that `CVOpenGLTextureCacheCreateTextureFromImage` will provide a non-`NULL` texture though, since the error is checked above.

libnodegl/backends/gl/glcontext_nsgl.m
```diff
@@ -114,7 +115,17 @@ static int nsgl_init(struct glcontext *ctx, uintptr_t display, uintptr_t window,
...
         [nsgl->handle setView:nsgl->view];
     }
-
+    
+    CVReturn err = CVOpenGLTextureCacheCreate(kCFAllocatorDefault,
+                                                NULL,
+                                                nsgl->handle.CGLContextObj, 
+                                                nsgl->pixel_format.CGLPixelFormatObj, 
+                                                NULL,
+                                                &nsgl->texture_cache);
```
kgrant-gpsw, line 124: Do we need a unique cache, or can `ngli_glcontext_get_texture_cache` be used?

libnodegl/backends/gl/glcontext_nsgl.m
```diff
@@ -237,6 +265,10 @@ static void nsgl_uninit_external(struct glcontext *ctx)
 
     if (nsgl->framework)
```
kgrant-gpsw, line 266: What is the difference between these two `uninit` cases (and how do we know that the cache must be freed in both)?

kgrant-gpsw (commented):
I think the changes mostly look fine, just wondering about the extra cache.
Also it seems a linter check failed, maybe it does not like formatting somewhere.

libnodegl/backends/gl/gpu_ctx_gl.c
```diff
@@ -131,6 +132,66 @@ static int wrap_capture_cvpixelbuffer(struct gpu_ctx *s,
...
+    
+    const size_t width = CVPixelBufferGetWidth(buffer);
+    const size_t height = CVPixelBufferGetHeight(buffer);
+    
+    CVOpenGLTextureCacheRef cache = ngli_glcontext_get_texture_cache(gl);
+    CVReturn cv_ret = CVOpenGLTextureCacheCreateTextureFromImage(kCFAllocatorDefault,
+                                                                   cache,
+                                                                   buffer,
+                                                                   NULL,
+                                                                   &cv_texture);
```
cboesch-gpsw, line 153: vertical indent is off

cboesch-gpsw (changes requested):
1. It looks like your PR includes several changes; in particular there is a mix between the NSGL and the backend itself. Can't those be split?
2. Why do you mention "GPU capture"?
3. Your branch name doesn't follow the convention `<name>/<short-desc>` and is also leaking internal information about the GoPro company into a public repository

VERSION
```diff
@@ -1 +1 @@
-2022.8
+2023.2
```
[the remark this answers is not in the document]
steventang-gpsw: Sorry, didn't know the rules, will look into it and do some modifications
1. About the version number, I'll revert it back
2. About your question 1, if it is mandatory to split, I'll split, otherwise, it is better to keep it as one since partial might be blocked by test.
3. About question 2, is "Support OS X Core Video capture with OpenGL" better? or your suggestions
4. That branch name was generated by the JIRA ticket, any suggested branch name?

steventang-gpsw:
> I think the changes mostly look fine, just wondering about the extra cache. Also it seems a linter check failed, maybe it does not like formatting somewhere.

Will fix the format, BTW, do we have any rules which we can apply to clang-format for automatic formatting?

libnodegl/backends/gl/glcontext_nsgl.m
steventang-gpsw: when calling OpenGL context creation, we can ask the backend to create either external OpenGL context or Internal OpenGL context, the definition of external is 
```
struct ngl_config_gl {
    /*
     * Whether the OpenGL context is external or not. If the OpenGL context is
     * external, it is the user responsibility to manage the OpenGL context and
     * make sure it is current before calling any of the ngl_* functions.
     */
    int external;
    /*
     * External OpenGL framebuffer used for rendering. The framebuffer must
     * have a color attachment composed of 4 color components (R, G, B, A) and
     * a combined depth and stencil buffer attached to it.
     */
    uint32_t external_framebuffer;
};
```
One of two deinitialization  (nsgl_uninit and nsgl_uninit_external) will be called when doing clean up

libnodegl/backends/gl/glcontext_nsgl.m
steventang-gpsw: The texture cache was created here [OpenGL context initialization] only once, later, we use ngli_glcontext_get_texture_cache to get the reference of this one for using until it was destroyed.

libnodegl/backends/gl/gpu_ctx_gl.c
steventang-gpsw: Will update to use CVOpenGLTextureRelease, this function is equivalent to CFRelease, but is NULL safe.

libnodegl/backends/gl/gpu_ctx_gl.h
steventang-gpsw: Adopted

steventang-gpsw:
Close this one, use PR # 513 instead.

PR #509, opened 2023-02-15
build: move lib sources into a dedicated src sub-directory

cboesch-gpsw:
This prevents a mix of project and source files that can confuse toolchains. In particular, the VERSION file happens to sometimes be included on case insensitive filesystem (a "version" file is looked up by cstddef somehow when trying to compile C++ code).

Aside from this particular issue, the benefit of this change is to have a better isolation of the sources from other files (build specs, documentation, scripts, …).

Note that the documentation and the logging now display the "src" sub-directory. The string is short enough that we can consider this harmless.

The generation rules (for the doc, specs and OpenGL wrappers) were tested and working as expected. Also, the extensive use of files() in the meson.build makes sure that there is no forgotten source file reference on any platforms.


hsong-gpsw:
This is looking good, no more conflicts with c++ compiler

hsong-gpsw (approved):

PR #511, opened 2023-02-15
Add option to generate pdb for windows in release mode

rlavarec-gpsw:


libnodegl/meson.build
```diff
@@ -653,15 +653,31 @@ elif host_system == 'iphone'
...
 
+lib_link_args = []
+
+# Add options to generate pdb under Windows for release
+if cc.get_id() == 'msvc' and not (( get_option('buildtype') == 'debug') or (get_option('buildtype') == 'debugoptimized'))
+  lib_c_args += '/Zi /FS'
+  lib_link_args += '/PDB:'+get_option('prefix') / get_option('bindir')  / 'nodegl-'+version_array[0]+'.pdb'
+  lib_link_args += '/DEBUG'
+  lib_link_args += '/OPT:REF'
+  lib_link_args += '/OPT:ICF'
```
cboesch-gpsw, line 664: So I'm not convinced by this since meson seems to have support for this: `mesonbuild/linkers/linkers.py:        return self._apply_prefix(['/DEBUG', '/PDB:' + self.get_debugfile_name(targetfile)])`

I believe a `debug` should trigger that, but there is definitely a way to have them builtin without a maintenance cost on our side.

cboesch-gpsw (changes requested):

libnodegl/meson.build
rlavarec-gpsw: True PDB are generated with debug flag
The problem exposed: https://github.com/mesonbuild/meson/issues/3596
In this thread we can understand that "debugoptimized" is not "RelWithDebInfo"

But after checking the code

mesonbuild/compilers/mixins/visualstudio.py
```
        assert crt_val in {'from_buildtype', 'static_from_buildtype'}
        dbg = 'mdd'
        rel = 'md'
        if crt_val == 'static_from_buildtype':
            dbg = 'mtd'
            rel = 'mt'
```

```
        elif buildtype == 'debugoptimized':
            return self.crt_args[rel]
        elif buildtype == 'release':
            return self.crt_args[rel]
```
"debugoptimized" seems to be "RelWithDebInfo"

But I don't know on the other platform if "debugoptimized" is like a "RelWithDebInfo", because Microsoft is a bit different.

That's why I limited the impact with:

`if cc.get_id() == 'msvc' and not (( get_option('buildtype') == 'debug') or (get_option('buildtype') == 'debugoptimized'))`

PR #512, opened 2023-02-15
ci:  use stable versions for python linters

mrobertseidowsky-gpsw:


.github/workflows/linters.yml
```diff
@@ -12,15 +12,33 @@ jobs:
...
+          sortPaths: ${{steps.changed-python-files.outputs.python_files}}
+          configuration: "--check --diff "
 
   black:
     name: black
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: psf/black@stable
+      - uses: dorny/paths-filter@v2
```
cboesch-gpsw, line 33: psf is Python Software Foundation, and dorny is a single individual. Same thing with isort. Not exactly a fan of that…

mrobertseidowsky-gpsw:
@cboesch-gpsw let's close this PR since you did one with black version freeze?

cboesch-gpsw:
> @cboesch-gpsw let's close this PR since you did one with black version freeze?

Well, I did update the files but I didn't freeze the black version: can't we configure the CI to force a version and bump it manually or something?

.github/workflows/linters.yml
mrobertseidowsky-gpsw: - For black, I can do something like:

```
- uses: psf/black@stable
  with:
    options: "--check --diff"
    version: "~= 23.0"
```
to get a stable version for year 2023. See [Stability Police](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy)

- For isort, I can do something like:

`- uses: isort/isort-action@v1.1.0`
to use a freezed version.

- About the question of processing linters on all or subset of files, let me know what you decide together.

cboesch-gpsw: The subset won't be needed anymore if the linters versions don't change :)

rlavarec-gpsw (approved):

PR #515, opened 2023-02-16
utils: add shared read for ngli_get_filesize for Windows

rlavarec-gpsw:
utils: This option under windows enable multi process to read the same file at the same time.
If we do not set this option, 2 or more ngl-desktop process running in parallels could failed to get size of asset.
No side effects.

### DIFF
--- a/libnodegl/src/utils.c
+++ b/libnodegl/src/utils.c
@@ -181,7 +181,7 @@ void ngli_thread_set_name(const char *name)
 int ngli_get_filesize(const char *filename, int64_t *size)
 {
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


cboesch-gpsw (changes requested):
Can you add in the commit description an explanation on why we need this change, and what are the potential side effects / restrictions of setting that?

Please also use the "utils:" prefix in the commit message.

cboesch-gpsw (changes requested):
1. You updated the PR title but not the commit itself.
2. The PR need a rebase
3. Nitpick: while at it, please lowercase the `add`.

cboesch-gpsw (approved):

PR #522, opened 2023-03-03
Add Support For Layered Image Resources

wmacdonald-gpsw:
Support for treating layered resources (cube maps, 3d textures, and texture arrays) as 2D image arrays when they are bound as images. This is used for compute shaders, where the use of imageLoad() and especially imageStore() is critical.


hsong-gpsw (approved):

wmacdonald-gpsw:
@cboesch-gpsw in addition to code feedback, please advise on preferences for breaking up the commit. Is one commit ok since it is all related to one feature add? Should I separate cube texture feature from 3D texture feature? Should I separate changes by each file, or by sections (add core enums, add pgcraft support, add gl support, add vk support, etc.) Should adding tests be a separate commit?

tests/compute.py
```diff
@@ -359,3 +359,163 @@ def compute_image_load_store(cfg: SceneCfg, show_dbg_points=False):
...
+
+    if show_dbg_points:
+        cuepoints = _get_compute_histogram_cuepoints()
+        group.add_children(get_debug_points(cfg, cuepoints))
+
+    return group
+
+
+@test_fingerprint(nb_keyframes=1, tolerance=1)
+@scene(show_dbg_points=scene.Bool())
```
cboesch-gpsw, line 478: You have to choose between the debug points and the fingerprint signature. In this case I believe you actually want the cue points (6 points spread on each colored area). The fingerprint kinda works but here the cue points would be more reliable I believe.

Try in ngl-control to check the `show_dbg_points` option (bottom-left of the UI) to see where the points are actually tested. You have to position them appropriately (and reduce their count). You can either hardcode the 6 positions, or use a function to generate them (below you used the histogram function to get them, you don't want that)

cboesch-gpsw (changes requested):
Please use a prefix in the commit message to be a bit more specific, something like: `compute,render: support layered image texture formats`?

BTW, I'm assuming this also works with a `Render` pipeline, not just `Compute`?

With regards to your concern about code split, as is is probably fine. When the tests are large we tend to split them out but that's OK here IMO.

tests/compute.py
wmacdonald-gpsw: I was wondering about this, thank you for the clarification.

tests/compute.py
```diff
@@ -359,3 +359,170 @@ def compute_image_load_store(cfg: SceneCfg, show_dbg_points=False):
...
+    color = imageLoad(tex_in, pos_load);
+    imageStore(tex_out, pos_store, color);
+}
+"""
+
+_B_X = 6;
+_B_Y = 4;
+
+def _get_compute_colorbox_cuepoints():
+    c = lambda i,f: (i / f + (1 / (2 * f))) * 2.0 - 1.0
```
cboesch-gpsw, line 433: Nitpick, not blocking: you don't need to explicit floats (`2.0` → `2`)

cboesch-gpsw:
> @cboesch-gpsw I have made all the updates except for adding additional tests for layered images in render pipelines. Please review again, and let me know your thoughts on adding those tests.

I'm fine without these tests if you're sure you're not breaking anything

wmacdonald-gpsw:
I have addressed the linter issue and also your other feedback.

cboesch-gpsw (approved):

PR #523, opened 2023-03-14
nodegl macos corevideo

steventang-gpsw:
This PR allows macOS to turn pixel-buffer into OpenGL Rectangle texture, when doing exporting, OpenGL renders the target video frames directly to the pixel buffer surface, and the encoder grabs from there directly for high performance.


kgrant-gpsw (approved):

libnodegl/src/backends/gl/gpu_ctx_gl.h
```diff
@@ -41,6 +41,13 @@ struct rendertarget;
 
 typedef void (*capture_func_type)(struct gpu_ctx *s);
 
+#define NGLI_CFRELEASE(ref) do { \
+    if (ref) {                   \
+        CFRelease(ref);          \
+        ref = NULL;              \
+    }                            \
+} while (0);
```
cboesch-gpsw, line 49: no `;` at the end of a `do { ... } while (0)` macro

libnodegl/src/backends/gl/glcontext_nsgl.m
```diff
@@ -114,7 +115,17 @@ static int nsgl_init(struct glcontext *ctx, uintptr_t display, uintptr_t window,
 
         [nsgl->handle setView:nsgl->view];
     }
-
+
```
cboesch-gpsw, line 118: seems you're adding trailing whitespaces here

libnodegl/src/backends/gl/gpu_ctx_gl.h
```diff
@@ -70,6 +70,9 @@ struct gpu_ctx_gl {
 #if defined(TARGET_IPHONE)
     CVPixelBufferRef capture_cvbuffer;
     CVOpenGLESTextureRef capture_cvtexture;
+#elif defined(TARGET_DARWIN)
+	CVPixelBufferRef capture_cvbuffer;
+    CVOpenGLTextureRef capture_cvtexture;
```
cboesch-gpsw, line 75: You didn't address the comment I made originally here

libnodegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -131,6 +132,68 @@ static int wrap_capture_cvpixelbuffer(struct gpu_ctx *s,
...
+        .width  = width,
+        .height = height,
+        .usage  = NGLI_TEXTURE_USAGE_COLOR_ATTACHMENT_BIT,
+    };
+
+    const struct texture_gl_wrap_params wrap_params = {
+        .params  = &attachment_params,
+        .texture = fbName,
+    };
+    NGLI_CFRELEASE(cv_texture);
```
cboesch-gpsw, line 183: Is it really OK to bind a texture you just released?

cboesch-gpsw (changes requested):
Please do not re-open PRs it's a nightmare to have a followup on previous comments. Just re-use the PR by push-forcing into your current working branch.

libnodegl/src/backends/gl/gpu_ctx_gl.h
steventang-gpsw: oh, it was a tab and I didn't notice, replaced it with 4 spaces

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: Yes, I didn't many video exporting tests from 720P to 5.3K video, all of them are smooth, memory usage were reasonable and no crash happened.

steventang-gpsw:
I was using fixup, but it only merged the update to the previous one, hope it will be ok

wmacdonald-gpsw (approved):

libnodegl/src/backends/gl/gpu_ctx_gl.c
cboesch-gpsw: What I'm afraid of is that it's going to read released GPU memory; so you may end up reading the wrong data, typically with a pipeline with multiple video texture bound, or eventually multiple pipelines. But maybe I'm misunderstanding something.

cboesch-gpsw:
> I was using fixup, but it only merged the update to the previous one, hope it will be ok

No it's not; please adjust the commit themselves

libnodegl/src/backends/gl/gpu_ctx_gl.c
wmacdonald-gpsw: Looking at it closer, I also share this concern. VRAM is generally implemented using scheduled garbage collection, so actual crashes when binding a just-released resource would be rare, but there is risk of invalid data in some cases. If it were actually invalid to release this, I would expect it to be flagged in debug builds, or at least under the GPU instruments. I do wonder if it is made safe by the capture on the CoreVideo side still holding a handle internally. I'll take a closer look at this flow and see if I can understand it enough to offer some help.

libnodegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -131,6 +132,67 @@ static int wrap_capture_cvpixelbuffer(struct gpu_ctx *s,
 
     return 0;
 }
+#else
+static int wrap_capture_cvpixelbuffer(struct gpu_ctx *s,
```
nsilvy-gpsw, line 136: I wonder if defining only one function and using platform specific code each time something differs between iOS and macOS would not be easier to maintain? In order to have as much common code as possible.
```c
#if defined(TARGET_IPHONE)
CVOpenGLESTextureCacheCreateTextureFromImage();
#else
CVOpenGLTextureCacheCreateTextureFromImage();
#endif
```

libnodegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -653,6 +653,7 @@ static int update_capture_cvpixelbuffer(struct gpu_ctx *s, CVPixelBufferRef capt
                                    NGLI_LOAD_OP_LOAD, &s_priv->default_rt_load)) < 0)
         return ret;
 
+    ngli_texture_freep(&s_priv->color);
```
nsilvy-gpsw, line 718: How is this related to this commit?

libnodegl/src/backends/gl/gpu_ctx_gl.c
nsilvy-gpsw, line 1: A couple of remarks on the commit message:
1. You could explain why you need to introduce this macro
2. `CFRelease` is actually part of `Core Foundation` and not `Core Video` like you mentioned in the title

Together with my other remarks, and looking at the commit history, I would rename the commit to something like:
```
gl/gpu_ctx_gl: add NGLI_SAFE_CFRELEASE macro

A safe-release mechanism is needed when releasing Core Foundation
objects because we don't want dangling pointers... blablabla
```

I'm not sure about which commit prefix to use though.

FYI: https://github.com/gopro/gopro-lib-node.gl/blob/main/doc/dev/ref/developers.md#integration

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: I updated and did a test, looks like the memory seems to be normal, I'll keep the new updated for safety.
image

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: NGLI_CFRELEASE is using in iOS, keep it consistent, @cboesch-gpsw what do you think?

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: I have a different opinion, common code is difficult to maintain, when you update the common code, you need to build both iOS and macOS to check if both are working, yes, as much common code as possible will save space, but these days the space is not a problem for the code,  the problem is people need spend double time to verify common code, and more #if #else will cause code very difficult to read.

libnodegl/src/backends/gl/gpu_ctx_gl.c
cboesch-gpsw: GPU memory leaks will not be observed with a traditional memory profiler.

libnodegl/src/backends/gl/gpu_ctx_gl.c
nsilvy-gpsw: Oh, I didn't know the macro already existed on iOS. So why creating a new one for macOS?

libnodegl/src/backends/gl/gpu_ctx_gl.h
```diff
@@ -41,6 +41,13 @@ struct rendertarget;
 
 typedef void (*capture_func_type)(struct gpu_ctx *s);
 
+#define NGLI_CFRELEASE(ref) do { \
```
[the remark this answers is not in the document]
cboesch-gpsw: The macro is fine; it will allow us to share it in a common utils without additional ifdefery + lib detection within the build system

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: Updated in the new edited commits

libnodegl/src/backends/gl/glcontext_nsgl.m
cboesch-gpsw: seems not afaict

libnodegl/src/backends/gl/texture_gl.c
```diff
@@ -284,6 +284,8 @@ static int texture_init_fields(struct texture *s)
         s_priv->target = GL_TEXTURE_3D;
     else if (params->type == NGLI_TEXTURE_TYPE_CUBE)
         s_priv->target = GL_TEXTURE_CUBE_MAP;
+    else if (params->type == NGLI_TEXTURE_TYPE_RECTANGLE)
+        s_priv->target = GL_TEXTURE_RECTANGLE;
```
cboesch-gpsw, line 288: I don't think you can have this commit before the one handling that particular target. Why did you split out this? It felt more meaningful before. Or did I miss something?

libnodegl/src/backends/gl/gpu_ctx_gl.c
```diff
@@ -121,7 +122,7 @@ static int wrap_capture_cvpixelbuffer(struct gpu_ctx *s,
 
     int ret = ngli_texture_gl_wrap(texture, &wrap_params);
     if (ret < 0) {
-        CFRelease(cv_texture);
+        NGLI_CFRELEASE(cv_texture);
```
cboesch-gpsw, line 125: Isn't this related to another commit? Didn't you squashed back some changes together?

cboesch-gpsw (changes requested):

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: The last commit is a bit big, I'll revise it.

libnodegl/src/backends/gl/gpu_ctx_gl.c
steventang-gpsw: agree, code revised.

libnodegl/src/backends/gl/glcontext_nsgl.m
wmacdonald-gpsw: I found two trailing whitespaces remaining on lines 121 and 122.

libnodegl/src/backends/gl/gpu_ctx_gl.c
wmacdonald-gpsw: @steventang-gpsw I'm not entirely sure I agree with that, but my counterargument probably depends on some standardization and testing support that I can't guarantee we have nailed down. I have no issue with the current way you have this. It is likely to get cleaned up in an audit if we make improvements to how portability is handled.

libnodegl/src/backends/gl/glcontext_nsgl.m
steventang-gpsw: Fixed, thanks

PR #525, opened 2023-03-22
make-release: macos date command does not support a default value for -I

wmacdonald-gpsw:
Long term, we should evaluate moving scripts like these to something portable like Python. In the interim, this patch with at least allow making releases from either linux or macos.


cboesch-gpsw (approved):
Small note, not blocking: please describe the change in the commit message, not a "true statement" (the git history reflects changes over time). So something like "scripts/make-release: make the date command work on macOS"

PR #527, opened 2023-04-05
ci: update vcpkg to fix missing msys2-runtime-3.2.0-8-x86_64.pkg.tar.zst

rlavarec-gpsw:


.github/workflows/ci_msvc.yml
```diff
@@ -57,6 +57,15 @@ jobs:
           .\VulkanRT-Installer.exe /S
 
           cd C:\vcpkg
+
+          # update vcpkg
```
mrobertseidowsky-gpsw, line 61: *`in case vcpkg cache is no more valid`
?

rlavarec-gpsw: If the cache is invalid, it will fail to retrieve it and not use it.

PR #529, opened 2023-05-08
Set rtt tests texture mag_filter to linear

jstys-gpsw:
* Needed to pass tests on Mac ARM architecture
* Not required on Intel Macs, and it's unclear on what the root issue with ARM devices is in this case


wmacdonald-gpsw:
Yes I agree, we should make sure this doesn't affect crafter tests. Pending that, it seems ok.

jstys-gpsw:
Crafter should be unaffected since this is just a change to the unit tests here. But I did run with and without the change and had the same results.

wmacdonald-gpsw (approved):

PR #531, opened 2023-05-09
metal backend for node.gl

hsong-gpsw:


libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+                                        std::vector<uint32_t> &spv_data ) {
+    ShaderTools::initialize();
+    const struct {
+        int stage;
+        EShLanguage lang;
+    } stages[] = {
+        { /*NGLI_PROGRAM_SHADER_VERT*/0, EShLangVertex },
+        { /*NGLI_PROGRAM_SHADER_FRAG*/1, EShLangFragment },
+        { /*NGLI_PROGRAM_SHADER_COMP*/2, EShLangCompute },
+    };
```
rlavarec-gpsw, line 198: Why not use directly  ?
`{ NGLI_PROGRAM_SHADER_VERT, EShLangVertex },`

libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+        LOG(ERROR, "unable to parse glsl: \n%s", shader->getInfoLog());
+        return false;
+    }
+
+    program->addShader(shader.get());
+    if (!program->link(EShMsgDefault)) {
+        LOG(ERROR, "unable to link shader: \n%s", program->getInfoLog());
+        return false;
+    }
+
```
rlavarec-gpsw, line 220: Maybe we could clear
` spv_data.clear();` in all case.

libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+        LOG(ERROR, "unable to get program intermediate");
+        return false;
+    }
+
+    return true;
+}
+
+bool ShaderTools::convert_spirv_to_msl(const std::vector<uint32_t> &spv_data, std::string &msl_data) {
+    auto compiler = std::make_unique<spirv_cross::CompilerMSL>(spv_data);
+    msl_data.clear();
```
rlavarec-gpsw, line 240: compile will affect it after
`msl_data = compiler->compile();`

libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+}
+
+void ShaderTools::finalize() {
+    ShaderTools::glslang_mutex.lock();
+    if (ShaderTools::glsl_initialized) {
+        glslang::FinalizeProcess();
+        ShaderTools::glsl_initialized = false;
+    }
+    ShaderTools::glslang_mutex.unlock();
+}
```
rlavarec-gpsw, line 186: RAII std::scoped_lock?

libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+void ShaderTools::write_to_file(const std::string &filename, std::string &data) {
+    ShaderTools::file_mutex.lock();
+    std::ofstream out(filename);
+    assert(out);
+    out.write(data.data(), data.size());
+    out.close();
+    ShaderTools::file_mutex.unlock();
+}
```
rlavarec-gpsw, line 252: Should it be locked?
Jeff used a hash of the file to generate the filename, so there is no collision. In fact There can be a collision if 2 processes running at the same time want to generate the same file. I have set the exclusive write and forbid the read during the write. And for reading the file, I'm trying several times and wait between each reading because of the multiprocess problem. It's mostly because of the unit tests.

libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+    assert(file);
+    auto output = std::ostringstream{};
+    output << file.rdbuf();
+    file.close();
+    ShaderTools::file_mutex.unlock();
+    return output.str();
+}
+
+bool ShaderTools::read_byte_code(const std::string &filename, char **byte_code, size_t &size) {
+    if (!byte_code)
```
rlavarec-gpsw, line 278: we could use a `std::vector<char> byte_code` for the byte code.
Because the interface is cpp with std::string, it is easier to manipulate and sure to clean memory allocation.

libnodegl/src/backends/common/shader_tools.cpp
```diff
@@ -0,0 +1,350 @@
...
+            map[m[2]] = std::stoi(m[3]);
+        }
+        input = m.suffix().str();
+    }
+}
+
+bool ShaderTools::shader_exists(const std::string &id, int stage) {
+    std::filesystem::path cache_folder{cache_dir};
+    if (!std::filesystem::exists(cache_folder)) {
+        std::filesystem::create_directories(cache_folder);
```
rlavarec-gpsw, line 330: I think you need to try catch for error. std::filesystem is throwing like crazy.

PR #532, opened 2023-05-18
Make node refcount changes atomic

jstys-gpsw:
This should ensure that in a multithreaded environment, only the last decrement will enter the delete block


hsong-gpsw:
both clang and gcc support __sync_fetch_and_add()

jstys-gpsw:
> both clang and gcc support __sync_fetch_and_add()

It's really just the Windows specificity I needed there - how does the below sound?
```
if defined (_WIN32)
#include <winnt.h>
#define ngli_atomic_add(ptr, num) InterlockedExchangeAdd ((ptr), (num))
#else
#include <stdatomic.h>
#define ngli_atomic_add(ptr, num) __sync_fetch_and_add ((ptr), (num))
#endif
```

Ideally I wanted to use `_Atomic` here, but MSVC support is still [experimental](https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/)

hsong-gpsw (approved):

jstys-gpsw:
Sounds good, I added that comment

jstys-gpsw:
I agree that this doesn't protect against all forms of race conditions (anything with data access certainly has to be protected externally), but I still think there is value in at least protecting reference count updates akin to how `std::shared_ptr` does it.
I'm definitely open to the more general discussion of what warranties node.gl should or shouldn't provide though.

rlavarec-gpsw:
We could use #include <stdatomic.h>
https://en.cppreference.com/w/c/language/atomic

rlavarec-gpsw:
> We could use #include <stdatomic.h> https://en.cppreference.com/w/c/language/atomic It is standard now and more easy to manage.
Because of a late support of Microsoft for it
https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/
This is not even supported by ISO C17 (2018) Standard (/std:c17)
Only a special flag enable to support it '/experimental:c11atomics'

We can stick to the old api

libnodegl/src/nodes.c
```diff
@@ -540,7 +549,7 @@ static int node_set_ctx(struct ngl_node *node, struct ngl_ctx *ctx, struct ngl_c
         if (node->state > STATE_UNINITIALIZED) {
             if (node->ctx != pctx)
                 return 0;
-            if (node->ctx_refcount-- == 1) {
+            if (ngli_atomic_add(&node->ctx_refcount, -1) == 1) {
```
rlavarec-gpsw, line 552: Maybe we could use `ngli_atomic_sub`, because if the change ctx_refcount to size_t, it could not work.

jstys-gpsw: Good point, I can switch to that

jstys-gpsw: Oh, Windows strikes again. The [subtract variant of the API](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-interlockedexchangesubtract) actually only operates on unsigned. So if we do switch to that, we could use it, but until then we should be using the add variant with signed `int`.

rlavarec-gpsw: Yes we are unlucky 🙅 

Their is InterlockedIncrement and InterlockedDecrement working on LONG and limited to 1.

jstys-gpsw: Those give the post-action result rather than pre-action as is currently though. We would need to change the delete comparison to 0, which would differ from some [other refcount comparisons](https://github.com/gopro/gopro-lib-node.gl/blob/main/libnodegl/src/backends/vk/glslang_utils.c#L147) that are made. (Side note: that first comparison is also under a lock, though [this other one](https://github.com/gopro/gopro-lib-node.gl/blob/main/libnodegl/src/backends/vk/ycbcr_sampler_vk.c#L140) I found is not. Might be worth securing the latter as part of this PR as well)

I do think that comparing to 0 might be a little clearer that all references are removed, but unless we change that everywhere, I would err on the side of consistency and keep the 1 comparison. @wmacdonald-gpsw, any preference?

rlavarec-gpsw: Personally, I prefer also to test 0. But it is true == 1 seems to be "the norm" in node.gl. I validate your PR Jack.

rlavarec-gpsw (approved):

jstys-gpsw:
I added another commit moving the function to utils, and renaming to clarify order of operations and operand type. Now also using this in `ycbcr_sampler_vk` so all the `refcount` variables I found will be updated atomically.

jstys-gpsw:
Updated commit text to better reflect guidelines, and added to the changelog (including the previous rtt unit test PR update)

PR #535, opened 2023-06-06
Improve hud for multi backend

fallera-gpsw:
Display node.gl version and backend in hud


libnodegl/src/hud.c
```diff
@@ -325,6 +327,18 @@ struct widget_spec {
...
 
+static const char* get_api_id(int backend)
+{
+    const char* api_id_map[] =
+    {
+        [NGL_BACKEND_OPENGL] = "opengl",
+        [NGL_BACKEND_OPENGLES] = "opengles",
+        [NGL_BACKEND_VULKAN] = "vulkan"
+    };
+
```
rlavarec-gpsw, line 338: I think we could reuse 
`api_map` struct from api.c file.

libnodegl/src/hud.c
```diff
@@ -855,6 +904,19 @@ static const struct widget_spec widget_specs[] = {
...
+        .text_rows     = NB_GLOBALINFOS,
+        .graph_w       = 320,
+        .nb_data_graph = NB_GLOBALINFOS,
+        .priv_size     = 0,
+        .init          = widget_globalinfos_init,
+        .make_stats    = widget_globalinfos_make_stats,
+        .draw          = widget_globalinfos_draw,
+        .csv_header    = widget_globalinfos_csv_header,
+        .csv_report    = widget_globalinfos_csv_report,
+        .uninit        = widget_globalinfos_uninit,
```
rlavarec-gpsw, line 936: Maybe we could put it optional.
`.uninit        = 0`
And test uninit before to be called.

libnodegl/src/hud.c
rlavarec-gpsw: There will be a cleanup phase on duplicate names after this PR
backend_map, api_map and api_id_map will be merged.

libnodegl/src/hud.c
```diff
@@ -447,11 +447,6 @@ static int widget_drawcall_init(struct hud *s, struct widget *widget)
     return make_nodes_set(scene, &priv->nodes, node_types);
 }
 
-static int widget_globalinfos_init(struct hud* s, struct widget* widget)
-{
-    return 0;
-}
-
```
jbnoel-gpsw: Why removing code that has just been added ?
Try to create a simpler git history.

jbnoel-gpsw (commented):
Thanks for this addition ! I'll test it.
See my small comment.
I'll let @wmacdonald-gpsw and the team do the approval

libnodegl/src/hud.c
jbnoel-gpsw: Why don't we start with the cleanup to avoid more duplication ?

fallera-gpsw: I made a new branch (fallera/Cleaning) and a new PR for this branch to do the cleanup first. And I rebased to the new Cleaning branch and simplified the commits

libnodegl/src/hud.c
```diff
@@ -421,9 +403,8 @@ static int widget_drawcall_init(struct hud *s, struct widget *widget)
 
 static void register_time(struct hud *s, struct latency_measure *m, int64_t t)
 {
-    m->total_times = m->total_times - m->times[m->pos] + t;
-    m->times[m->pos] = t;
-    m->pos = (m->pos + 1) % s->measure_window;
+    m->total_times = m->total_times - m->time + t;
+    m->time = t;
     m->count = NGLI_MIN(m->count + 1, s->measure_window);
 }
```
jbnoel-gpsw, line 427: I don't have enough context to validate this change.

fallera-gpsw: I'll remove this part

libnodegl/src/api.c
```diff
@@ -58,29 +58,35 @@ extern const struct api_impl api_gl;
...
         .api_impl = &api_gl,
 #endif
     },
     [NGL_BACKEND_VULKAN] = {
-        .string_id = "vulkan",
 #ifdef BACKEND_VK
         .api_impl = &api_vk,
 #endif
     },
 };
```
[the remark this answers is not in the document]
fallera-gpsw: This is a local struct contained only in the api.c file so it cannot be accessed in other files. It is not part of the public api. 
The public api have these struct that I did not modify: 
- gpu_ctx, contained in gpu_ctx.h
- ngl_backend, contained in nodegl.h
I created a new function in internal.h to be able to access it from different files (api.c, gpu_ctx.c and hud.c)

libnodegl/src/hud.c
jbnoel-gpsw: Maybe the rest of the node.gl team can validate the change.

fallera-gpsw: It is not a valid change, the buffer of times is needed. I should not have pushed it in my PR.

libnodegl/src/hud.c
```diff
@@ -247,7 +248,8 @@ static const struct globalinfos_spec {
     const char* label;
     const uint32_t color;
 } globalinfos_specs[] = {
-    [GLOBALINFOS_NGL_VERSION]   = { "NGLVersion", 0xF43DF4FF }
+    [GLOBALINFOS_NGL_VERSION]   = { "NGLVersion", 0xF43DF4FF },
+    [GLOBALINFOS_BACKEND]       = { "Backend", 0x3DF4F4FF }
 };
```
rlavarec-gpsw, line 253: Maybe add the color as a comment next to it.

fallera-gpsw: I did this in my Cleaning branch and rebased this one on the Cleaning branch

libnodegl/src/hud.c
```diff
@@ -249,16 +255,26 @@ static const struct drawcall_spec {
...
+NGLI_STATIC_ASSERT(hud_nb_drawcall,     NGLI_ARRAY_NB(drawcall_specs)    == NB_DRAWCALL);
+NGLI_STATIC_ASSERT(hud_nb_globalinfos,  NGLI_ARRAY_NB(globalinfos_specs) == NB_GLOBALINFOS);
 
 enum widget_type {
     WIDGET_LATENCY,
     WIDGET_MEMORY,
     WIDGET_ACTIVITY,
     WIDGET_DRAWCALL,
+    WIDGET_GLOBALINFOS,
+    NB_WIDGET
```
jbnoel-gpsw, line 279: Keep trailing coma when possible.

libnodegl/src/hud.c
```diff
@@ -259,7 +260,8 @@ static const struct globalinfos_spec {
     const char* label;
     const uint32_t color;
 } globalinfos_specs[] = {
-    [GLOBALINFOS_NGL_VERSION]   = { "NGLVersion", BRIGHT_MAGENTA }
+    [GLOBALINFOS_NGL_VERSION]   = { "NGLVersion",   BRIGHT_MAGENTA },
+    [GLOBALINFOS_BACKEND]       = { "Backend",      BRIGHT_CYAN }
```
jbnoel-gpsw, line 264: keep trailing coma when possible.

libnodegl/src/internal.h
```diff
@@ -76,6 +76,8 @@ struct api_impl {
     int (*gl_wrap_framebuffer)(struct ngl_ctx *s, uint32_t framebuffer);
 };
 
+const char* ngl_get_backend_string_id(int backend);
```
jbnoel-gpsw, line 79: Shouldn't it be `ngli_` prefix (internal.h) ?

libnodegl/src/api.c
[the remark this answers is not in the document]
jbnoel-gpsw: I still think having everything in one place in `api_map` is easier, fragmented information is harder to maintain.
Why can't `api_map` be declared in `internal.h` and  implemented in `api.c` ?

libnodegl/src/hud.c
fallera-gpsw: I kept the same logic as the other enums where the "NB" enum values have no trailing coma like NB_LATENCY or NB_MEMORY for exemple

libnodegl/src/hud.c
fallera-gpsw: I updated the code with it

libnodegl/src/api.c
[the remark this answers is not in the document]
fallera-gpsw: Well, I think separating it is better. Information contained in api_map are specific to api.c file and only used here so other files shouldn't know its existence. It's usually better to keep stuff in the smallest scope possible to make reading, debugging and maintenance of the code easier.
The new functions is used in multiple places and is not specific to api.c hence the declaration of the function in internal.h.
I implemented the function in api.c since there is no internal.c file and api.c seems to be the most suited place. 
Also I used a function to check if the parameter is a defined backend and assert if not while this is not possible with an array and we would need to check in every places where we would use it (gpu_context.c, api.c and hud.c for now).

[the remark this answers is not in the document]
rlavarec-gpsw: It is true the string id for the backend name was repeated several time (and it is also repeated in the tools). It is better to have only one function to call to get the names, easier to change.

libnodegl/src/api.c
```diff
@@ -646,7 +652,7 @@ int ngl_configure(struct ngl_ctx *s, struct ngl_config *config)
     s->api_impl = api_map[config->backend].api_impl;
     if (!s->api_impl) {
         LOG(ERROR, "backend \"%s\" not available with this build",
-            api_map[config->backend].string_id);
+            ngli_get_backend_string_id(config->backend));
```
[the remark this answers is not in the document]
wmacdonald-gpsw: Yes, I see what you mean. I think there is some lack of clarity about the layers, such as ngli, within node.gl. There was some discussion to establish ngli as a separate public layer eventually, that could be used as a portable API for making low level GPU calls (for prototyping, unit testing of shaders, etc.). This might be something we just need to work to more clearly define and establish within the code structure. For now, it's not that critical.

Good point about string not being part of the abstraction because you don't have an instance. I suppose my C++ background is leading me to want class-like abstraction, where each backend can provide static methods that don't require an instance. This seems like a useful concept anyway, as we introduce more backends and want to be able to enumerate and describe the available backends in a given build/platform.  Probably something that isn't critical to worry about for now, but worth a future design review.

wmacdonald-gpsw (approved):

libnodegl/src/internal.h
fallera-gpsw: I updated the name with `ngli_` prefix

PR #536, opened 2023-06-09
Fallera/cleaning

fallera-gpsw:
Do some cleaning to prepare the PR on hud.


libnodegl/src/hud.c
```diff
@@ -421,9 +403,8 @@ static int widget_drawcall_init(struct hud *s, struct widget *widget)
 
 static void register_time(struct hud *s, struct latency_measure *m, int64_t t)
 {
-    m->total_times = m->total_times - m->times[m->pos] + t;
-    m->times[m->pos] = t;
-    m->pos = (m->pos + 1) % s->measure_window;
+    m->total_times = m->total_times - m->time + t;
+    m->time = t;
     m->count = NGLI_MIN(m->count + 1, s->measure_window);
 }
```
rlavarec-gpsw, line 409: I think the previous code was good.
But it need some comments.
Each time register_time is called, it remove current index -> m->times[m->pos] to replace by the new one t.
At first all is 0, so it remove "nothing", and the algo kick in after the first loop, where it remove from "total_times" the previous value before to add a new one.

rlavarec-gpsw: `m->count = NGLI_MIN(m->count + 1, s->measure_window);`
Could be replace by 
```
if(m->count<s->measure_window)
     m->count += 1;
```
No need to affect all the time the value if not needed.

fallera-gpsw: I'll remove this part for now. I'll come back on this in another cleaning pass for the count computation

wmacdonald-gpsw (approved):

wmacdonald-gpsw:
Looks good, but we need to investigate the MinGW failure, which looks like a machine config issue.

rlavarec-gpsw (approved):

PR #542, opened 2023-07-03
configure.py: update pkgconfig source

rlavarec-gpsw:


mrobertseidowsky-gpsw (approved):
OK with this new functionnal link!
Can you just reword the commit message please to mention `pkgconf` (and not `pkgconfig` that can be confusing), thanks.

PR #544, opened 2023-07-05
Fix cross-compilation on macOS

nsilvy-gpsw:
Context
-------

The macOS platform is now able to cross-compile between Intel and Arm architectures, but the build script doesn't allow it. This PR simply adds support for this use case.

Commit message
--------------

When cross-compiling on macOS, we got the following error:
> Tried to install a target for the build machine in a cross build.

The problem was that the flag `is_native` wasn't correctly set.

The proposed solution is to rely on `meson.is_cross_build()` to set the `is_native` flag, instead of manually setting it depending on the `host_system`.

### DIFF
--- a/libnodegl/meson.build
+++ b/libnodegl/meson.build
@@ -85,10 +85,7 @@ if host_system == 'windows'
 endif
 
 # Need a flag to declare if using cross compilation or not
-is_native = true
-if host_system == 'android' or host_system == 'iphone'
-  is_native = false
-endif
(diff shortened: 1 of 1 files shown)


hsong-gpsw (approved):

wmacdonald-gpsw (approved):

nsilvy-gpsw:
Closing in favor of https://github.com/gopro/gopro-lib-node.gl/pull/545

PR #546, opened 2023-07-20
hud: add draw time debug logging when enabled

jstys-gpsw:
In addition to displaying in the hud, this will now also print the draw times with DEBUG level. This should make it easier to share performance numbers when needed.


hsong-gpsw (approved):

fallera-gpsw (approved):

wmacdonald-gpsw (approved):

gabbe-gpsw:
Are you aware of the csv file export in nodegl ?
This seems redundant but I don't know what your use case is...

jstys-gpsw:
That is news to me 😅 But I was using the HUD with app rendering, where app actions were relevant to different render timing. I think there is still a use in having draw logging intermingled with other logging contexts

PR #548, opened 2023-08-03
node_animkeyframe: rename AnimKeyFrameQuat node value

jstys-gpsw:
Most usage of this node feeds in the values at construction, so I didn't actually find any existing references to the old `quat` field here.


libnodegl/src/node_animkeyframe.c
```diff
@@ -92,7 +92,7 @@ ANIMKEYFRAME_PARAMS(float, value, NGLI_PARAM_TYPE_F64, scalar);
 ANIMKEYFRAME_PARAMS(vec2,  value, NGLI_PARAM_TYPE_VEC2, value);
 ANIMKEYFRAME_PARAMS(vec3,  value, NGLI_PARAM_TYPE_VEC3, value);
 ANIMKEYFRAME_PARAMS(vec4,  value, NGLI_PARAM_TYPE_VEC4, value);
-ANIMKEYFRAME_PARAMS(quat,  quat,  NGLI_PARAM_TYPE_VEC4, value);
+ANIMKEYFRAME_PARAMS(quat,  value,  NGLI_PARAM_TYPE_VEC4, value);
 ANIMKEYFRAME_PARAMS(color, color, NGLI_PARAM_TYPE_VEC3, value);
```
jbnoel-gpsw, line 96: @sfranz-gpsw @jstys-gpsw What about `color` ?

jbnoel-gpsw (commented):
@sfranz-gpsw Please test from the Python side.
See my suggestion about `color`

libnodegl/src/node_animkeyframe.c
jstys-gpsw: `UniformQuat` and `UniformColor` both use `value` - so making the change would match those. As a counterpoint, we do use `color` in `RenderColor`, and `RenderGradient` has multiple `color_` fields, but I'd say those are more specific contexts where that naming makes sense when considering the additional fields there

jstys-gpsw:
I added a commit with the color param change

wmacdonald-gpsw (changes requested):

libnodegl/src/node_animkeyframe.c
```diff
@@ -92,7 +92,7 @@ ANIMKEYFRAME_PARAMS(float, value, NGLI_PARAM_TYPE_F64, scalar);
 ANIMKEYFRAME_PARAMS(vec2,  value, NGLI_PARAM_TYPE_VEC2, value);
 ANIMKEYFRAME_PARAMS(vec3,  value, NGLI_PARAM_TYPE_VEC3, value);
 ANIMKEYFRAME_PARAMS(vec4,  value, NGLI_PARAM_TYPE_VEC4, value);
-ANIMKEYFRAME_PARAMS(quat,  quat,  NGLI_PARAM_TYPE_VEC4, value);
+ANIMKEYFRAME_PARAMS(quat,  value,  NGLI_PARAM_TYPE_VEC4, value);
```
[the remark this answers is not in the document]
jstys-gpsw: Whoops, fixed it

wmacdonald-gpsw (approved):

fallera-gpsw (approved):

PR #555, opened 2023-08-18
mingw dll load failed while importing pynodegl

rlavarec-gpsw:


ubitux:
Please preserve the original authorship when cherry-picking commits made by others, and ideally the other metadata as well. `git cherry-pick --signoff` is your friend if you're making amends to the commit (preserves the metadata and signs it off with your name to record that you made changes). Thanks.

PR #558, opened 2023-08-24
backend: remove OpenGLES from being used inside Windows

rlavarec-gpsw:


jbnoel-gpsw (approved):
OK, if we are not able to only exclude OpenGLES 2 (and support OpenGLES 3)

wmacdonald-gpsw (approved):

PR #559, opened 2023-08-25
metal: Release memory in node api calls

hsong-gpsw:
Wrap node API with NSAutoreleasePool to ensure memory is released.


libnodegl/src/backends/metal/hwmap_videotoolbox_mtl.mm
```diff
@@ -101,10 +101,12 @@ static int vt_mtl_map_frame(struct hwmap *hwmap, struct sxplayer_frame *frame)
     sxplayer_release_frame(vt->frame);
     vt->frame = frame;
 
+    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
```
jstys-gpsw, line 104: Can we use `@autoreleasepool {}` blocks instead? [Apple docs](https://developer.apple.com/documentation/foundation/nsautoreleasepool) mention them as more efficient in general, but with the numerous returns that would also make it easier to not miss a release somewhere

hsong-gpsw: Yes,  That is great idea.  Thanks @jstys-gpsw

jstys-gpsw (approved):

PR #566, opened 2023-09-19
Fallera/d3d12

fallera-gpsw:
d3d12: merge msaa support and performances improvement


libnodegl/src/backends/d3d12/pipeline_d3d12.cpp
```diff
@@ -339,6 +339,7 @@ int d3d12_pipeline_set_resources(pipeline* s,
     }
     
 	// Add support for gl_NumWorkGroups
+    if(false)
```
rlavarec-gpsw, line 326: Peut être mieux tester
`if(ENABLE_GPU_VALIDATION)
`

PR #586, opened 2023-10-30
opengl:  return 0 when make current success in iOS EAGL

hsong-gpsw:
This eliminated false error message of "Failed resize video engine" in video engine


libnodegl/src/backends/gl/glcontext_eagl.m
```diff
@@ -293,7 +293,7 @@ static int eagl_make_current(struct glcontext *ctx, int current)
         ret = [EAGLContext setCurrentContext:nil];
     }
 
-    return ret;
+    return ret == 1? 0: -1;
```
rlavarec-gpsw, line 296: https://developer.apple.com/documentation/opengles/eaglcontext/1624882-setcurrentcontext
Seems to return a bool.
We could do != FALSE to rule out TRUE not been 1.
` ret != FALSE? 0: -1;`

hsong-gpsw: Thanks @rlavarec-gpsw   I have change according to your suggestion.

rlavarec-gpsw (approved):

libnodegl/src/backends/gl/glcontext_eagl.m
```diff
@@ -293,7 +293,7 @@ static int eagl_make_current(struct glcontext *ctx, int current)
         ret = [EAGLContext setCurrentContext:nil];
     }
 
-    return ret;
+    return ret != FALSE ? 0: -1;
```
rlavarec-gpsw, line 296: It seems it is more `NO` use in pure Objective C.

https://developer.apple.com/documentation/objectivec/objective-c_runtime/boolean_values

hsong-gpsw: right, corrected.  Thanks

libnodegl/src/backends/gl/glcontext_eagl.m
```diff
@@ -293,7 +293,7 @@ static int eagl_make_current(struct glcontext *ctx, int current)
         ret = [EAGLContext setCurrentContext:nil];
     }
 
-    return ret;
+    return ret != NO ? 0: -1;
```
rlavarec-gpsw, line 296: Sorry I think about it overnight and we could write the function the other way, to do not bother avec FALSE or NO or false.

This is working in every case and simpler to read.
`ret` is not returning, so we do not need to use it as a name.

My new final proposition
```
    BOOL isOK;
    struct eagl_priv *eagl = ctx->priv_data;

    if (current) {
        isOK = [EAGLContext setCurrentContext:eagl->handle];
    } else {
        isOK = [EAGLContext setCurrentContext:nil];
    }

    return isOK ? 0: -1;
```

fallera-gpsw (approved):

